PR #2, opened 2019-03-21
Add RHEL7 dockerfile from OSE

dcbw:
Also revert the travis change, it breaks linting and we don't actually need to change any paths from openvswitch to openshift.

@pecameron

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ go:
 before_install:
     - export GOPATH=$HOME/go
     - export PATH=$HOME/usr/local/go/bin:$GOPATH/bin:$PATH
-    - mkdir -p $GOPATH/src/github.com/openshift
-    - mv $TRAVIS_BUILD_DIR $GOPATH/src/github.com/openshift/ovn-kubernetes
-    - cd $GOPATH/src/github.com/openshift/ovn-kubernetes
+    - mkdir -p $GOPATH/src/github.com/openvswitch
+    - mv $TRAVIS_BUILD_DIR $GOPATH/src/github.com/openvswitch/ovn-kubernetes
+    - cd $GOPATH/src/github.com/openvswitch/ovn-kubernetes
 install:
     - eval "$(gimme 1.10.4)"
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: required
 language: go
 go:
-      - 1.8.1
+      - 1.10.4
 before_install:
     - export GOPATH=$HOME/go
     - export PATH=$HOME/usr/local/go/bin:$GOPATH/bin:$PATH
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,15 +1,15 @@
...
 go:
-      - 1.8.1
+      - 1.10.4
 before_install:
     - export GOPATH=$HOME/go
     - export PATH=$HOME/usr/local/go/bin:$GOPATH/bin:$PATH
-    - mkdir -p $GOPATH/src/github.com/openshift
-    - mv $TRAVIS_BUILD_DIR $GOPATH/src/github.com/openshift/ovn-kubernetes
-    - cd $GOPATH/src/github.com/openshift/ovn-kubernetes
+    - mkdir -p $GOPATH/src/github.com/openvswitch
```
pecameron: openvswitch is for the upstream openvswitch/ovn-kubernetes community repo,
The new repo is openshift/ovn-kubernetes. The ocp repo is ose-ovn-kubernetes.

Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
+
+WORKDIR /go-controller
+COPY go-controller/ .
+
+# build the binaries
+RUN make
+
+
+FROM rhel7
```
pecameron: I think this is a problem for the OKD builds. I have not found out how we are building OKD images at present. Subscription manager is involved when rhel is used.

Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+	yum clean all
+
+RUN yum install -y  \
+	PyYAML bind-utils \
+	openssl \
+	numactl-libs \
+	firewalld-filesystem \
+	libpcap \
+	hostname \
+	"openvswitch-2.9*" \
```
pecameron: Again, depending on how this is built the rpms will be different. openvswith-2.9 is from the centos build, although I think we are now using 2.9.2. With rhel, fast datapath should be used with 2.11. Guru didn't like 2.9.0, he wanted 2.9.2 or 2.10 which is why openvswitch/ovn-kubernetes builds with 2.9.2. There are additional rpms needed for 2.9.2 and 2.11

.travis.yml
dcbw: Yes, I know the repos changed here. But all the import paths are still openvswitch, and that's fine for Travis. We don't need to change the travis config here at all, even though we're in a different repo.

Dockerfile
dcbw: Actually, I think we should be using:

`FROM openshift/origin-base`

here; that's what other components appear to be doing.

Dockerfile
dcbw: You're right, we should do openvswitch2.11-*

dcbw:
@pecameron PTAL thanks!

pecameron:
/lgtm
Shall we just merge?

PR #8, opened 2019-05-15
WIP Fix image build

danwinship:


danwinship:
> Is this work in progress?

Yes, hence the "WIP" in the title. (There is unfortunately no way, AFAIK, to get openshift-ci-bot to *not* tag people for review.)

> A note on kubectl vs oc

In OpenShift they are they same binary. But the image we're grabbing it from only has `/usr/bin/oc` on disk, so we copy that over, and then symlink it to `kubectl` as well. So this isn't introducing any oc dependency, it's just getting an OpenShift-built kubectl binary into the OpenShift-built ovn-kubernetes image.

pecameron:
@danwinship originally oc/kubernetes was not needed in the image and then we rearranged the install and got included. It would be great if didn't need it.

danwinship:
Currently ovnkube.sh requires kubectl. In fact, it requires kubectl in more places than it did 6 months ago. We can try to get rid of that dependency upstream, but for now, that dependency is there

PR #9, opened 2019-06-04
Merge upstream from 2019-06-04

dcbw:
@danwinship


danwinship:
/lgtm

and looks like the daemonsets/deployments haven't changed so we don't need any CNO updates

PR #11, opened 2019-06-22
WIP: Make the image slightly smaller

danwinship:
- Use origin-cli as the base
- Don't install containernetworking-plugins
- Don't copy in ovn-kube-util

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,9 +17,7 @@ COPY go-controller/ .
 # build the binaries
 RUN make
 
-FROM openshift/origin-cli AS cli
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -17,9 +17,7 @@ COPY go-controller/ .
 # build the binaries
 RUN make
 
-FROM openshift/origin-cli AS cli
-
-FROM openshift/origin-base
+FROM openshift/origin-cli
```
squeed: FYI, this is ignored by OSBS, so you'll have to update this in `release` too.

PR #13, opened 2019-07-22
Add iptables wrappers to Fedora image

danwinship:
It is "difficult" to build the RHEL-based openshift/ovn-kubernetes image by hand. ("Difficult" meaning I have no idea how to do it; if you try to build from the toplevel Dockerfile it ends up trying to pull RPMs from some repo inside the CI cluster.)

However, the upstream image also no longer works with our CNO DaemonSet, which assumes you will have iptables wrappers installed.

So this fixes `Dockerfile.fedora` in our fork of ovn-kubernetes to use the iptables wrappers, so we can build images based off that for local testing.

(@dcbw I believe the upstream `Dockerfile.fedora` currently creates an image that has *no* iptables binaries in it, which makes the node image crash at startup at least with local gateway mode.)

### DIFF
--- a/dist/images/Dockerfile.fedora
+++ b/dist/images/Dockerfile.fedora
@@ -50,6 +50,14 @@ COPY .git/refs/heads/ /root/.git/refs/heads/
 COPY ovnkube.sh /root/
 COPY ovn-debug.sh /root/
 
+# iptables wrappers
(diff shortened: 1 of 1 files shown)


danwinship:
It's not an upstream fix. It's a hack to make it possible to build local testing images of this repo.

PR #18, opened 2019-08-16
[4.3]OVN ovnkube.sh remove version 1,2 support

pecameron:
Back port (most of) https://github.com/ovn-org/ovn-kubernetes/pull/787
Remaining pieces are for future 4.3 features.


danwinship:
If we don't want any of this until 4.3 then you can probably just close the PR because we'll definitely want to do a full rebase to upstream for 4.3 anyway, some time after 4.2 branches off

pecameron:
We are no longer using ovnkube.sh

PR #19, opened 2019-08-23
Bug 1749131: Block access to MCS and metadata

danwinship:
Since we didn't figure out a higher-level approach to this for 4.2, here is a port of https://github.com/openshift/origin/pull/22821 to our build of ovn-kubernetes. This should go away in 4.3.

(This lack of this currently causes an e2e-aws-ovn-kubernetes test failure, `node client cert requests armoring: deny pod's access to /config/master API endpoint`.)

### DIFF
--- a/go-controller/pkg/cni/helper_linux.go
+++ b/go-controller/pkg/cni/helper_linux.go
@@ -213,6 +213,20 @@ func setupSriovInterface(netns ns.NetNS, containerID, ifName, macAddress, ipAddr
 	return hostIface, contIface, nil
 }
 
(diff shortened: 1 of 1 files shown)


danwinship:
/test e2e-aws-ovn-kubernetes

/hold
until we see that it actually passes that test

danwinship:
/test e2e-aws-ovn-kubernetes
(the "MCS is unreachable" test passed, but some others failed that shouldn't have... want to see if they're flakes or if maybe we somehow broke something...)

danwinship:
ok, looks like just flakes. ready to go

PR #20, opened 2019-08-29
Do not install arch-specific dependencies

yselkowitz:
numactl-libs is not built on s390x in RHEL 7.  It will be pulled in as a requirement of openvswitch2.11 on those architectures on which it is required.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN yum install -y  \
 	yum clean all
 
 RUN INSTALL_PKGS=" \
-	PyYAML bind-utils openssl numactl-libs firewalld-filesystem \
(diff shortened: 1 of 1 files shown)


pecameron:
@yselkowitz how far back does this dependency resolution go? Is it in 2.9, and 2.10 as well as 2.11? Upstream is at ovn-org/ovn-kubernetes and there are different versions of ovs including 2.9, 2.10, 2.11 depending on the image being built. This Dockerfile is not used upstream, however there are similar Dockerfiles that we can change for this.

yselkowitz:
The 2.9, 2.10, and 2.11 builds all show a dependency on libnuma.so.1()(64bit) etc. on those arches on which dpdk is enabled (x86_64, and ppc64le since 2.10).

PR #23, opened 2019-09-10
Bug 1749403: add iptables rules for Azure, etc, loadbalancers

danwinship:
Certain CloudProviders (eg, Azure) assume that they can redirect loadbalancer traffic directly to a node with no rewriting, and the node will accept the packets. This works with kube-proxy because of a rule that was intended just as a shortcut to avoid sending local traffic to the load-balancer unnecessarily. Make it work with ovn-kubernetes too.

This was easier to fix for local gateway only than it would be to fix for everything, so I only fixed it for local gateway here. We need to fix this correctly upstream still.

This PR also pulls in the fixes to make HostPorts work and a random iptables-rules cleanup fix because it simplifies my new code. Only the last commit is new.


dcbw:
@pcameron no it won't work for other gateway types, at least for now.

/lgtm

PR #26, opened 2019-09-17
bug 1748162: - [GCP]Failed to install cluster with OVN network type

pecameron:
Comment #11 redirects this to a vsphere bug
OVN as network type to install cluster on vsphere.

cannot load glue library: libibverbs.so.1

https://bugzilla.redhat.com/show_bug.cgi?id=1748162


pecameron:
@danwinship @JacobTanenbaum PTAL
The bug was originally for gcp but it is now vsphere. There is a missing library.

danwinship:
We don't actually need that library. It's only for OVS+DPDK and the error is harmless. (You'll see it in the OVS logs on successful installs too.)

pecameron:
@danwinship closing this is just plain wrong at so many levels.

pecameron:
@dcbw Created https://bugzilla.redhat.com/show_bug.cgi?id=1755607 to track this issue in ovs.

PR #27, opened 2019-09-19
Add hybrid cluster extensions framework

dcbw:
Doesn't pass traffic between Linux/Windows quite yet, but the master should be working.

On the master side, enable with:

`--enable-extensions --extensions-cluster-subnets=<windows cluster subnets>`

On the node, just:

`--enable-extensions`

@JacobTanenbaum @danwinship @squeed


dcbw:
@pecameron it should soonish go upstream, yes, but I need to start that discussion and unfortunately deadlines are somewhat tight on this one.

go-controller/extensions/pkg/controller/node_linux.go
```diff
@@ -0,0 +1,234 @@
...
+	// ovn-bridge-mappings maps a physical network name to a local ovs bridge
+	// that provides connectivity to that network.
+	_, stderr, err = util.RunOVSVsctl("set", "Open_vSwitch", ".",
+		fmt.Sprintf("external_ids:ovn-bridge-mappings=%s:%s", hybridNetworkName, hybridBridgeName))
+	if err != nil {
+		return fmt.Errorf("Failed to set ovn-bridge-mappings for ovs bridge %s"+
+			", stderr:%s: %v", hybridBridgeName, stderr, err)
+	}
+
+	_, stderr, err = util.RunOVSVsctl("--may-exist", "add-port", hybridBridgeName, "vxlan0",
```
danwinship: ovn-kubernetes generally gives its interfaces more descriptive names than `vxlan0` doesn't it?

go-controller/extensions/pkg/node/node_windows.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (n *nodeController) Delete(node *kapi.Node) {
+	return
+}
+
+// InitSelf initializes the node it is currently running on.
+// On Windows, this means:
+//  1. Setting up this node and its VxLAN extension for talking to other nodes
+//  2. Setting back annotations about its VTEP and gateway MAC address to its own node object
+func (n *nodeController) InitSelf() {
```
danwinship: (this whole file gets deleted in the next commit...)

go-controller/vendor/k8s.io/client-go/testing/fixture.go
```diff
@@ -129,6 +132,7 @@ func ObjectReaction(tracker ObjectTracker) ReactionFunc {
 			return true, nil, nil
 
 		case PatchActionImpl:
+			GinkgoT().Logf("#####################################")
```
danwinship: did you mean for this to be in the PR?

go-controller/pkg/cluster/master.go
```diff
@@ -378,12 +385,12 @@ func (cluster *OvnClusterController) syncNodes(nodes []interface{}) {
...
 }
 
-func (cluster *OvnClusterController) watchNodes() error {
+func (cluster *OvnClusterController) watchNodes(nodeFilterFunc NodeFilterFunc) error {
 	_, err := cluster.watchFactory.AddNodeHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			node := obj.(*kapi.Node)
 			logrus.Debugf("Added event for Node %q", node.Name)
-			err := cluster.addNode(node)
+			err := cluster.addNode(node, nodeFilterFunc)
```
danwinship: it would be more idiomatic to add `watchFactory.AddFilteredNodeHandler` a la `AddFilteredPodHandler`.

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -96,6 +97,12 @@ func main() {
 	c.Flags = append(c.Flags, config.OvnNBFlags...)
 	c.Flags = append(c.Flags, config.OvnSBFlags...)
 	c.Flags = append(c.Flags, config.OVNGatewayFlags...)
+	c.Flags = append(c.Flags, extcontroller.GetExtensionCLIFlags([]cli.Flag{
+		cli.BoolFlag{
+			Name:  "enable-extensions",
+			Usage: "Enables SDN extensions for hybrid clusters",
```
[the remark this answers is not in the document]
dcbw: @danwinship it's also for stuff like F5 or other stuff like that, where we give a direct tunnel into the overlay of the cluster.

go-controller/extensions/cmd/hybrid-sdn/hybrid-sdn.go
```diff
@@ -0,0 +1,82 @@
...
+
+	factory, err := factory.NewWatchFactory(clientset, stopChan)
+	if err != nil {
+		return err
+	}
+
+	master := ctx.Bool("master")
+	nodeName := ctx.String("node")
+
+	if err := controller.StartExtensionSDN(ctx, master, nodeName, clientset,
```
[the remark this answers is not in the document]
dcbw: @danwinship hybrid-sdn is the binary that Windows nodes run.

go-controller/vendor/k8s.io/client-go/testing/fixture.go
dcbw: Sure didn't.

go-controller/extensions/pkg/node/node_windows.go
dcbw: @danwinship fixed

go-controller/extensions/pkg/controller/master.go
```diff
@@ -0,0 +1,196 @@
...
+	if !util.IsWindowsNode(node) {
+		if ovnHostsubnet == nil {
+			if extHostsubnet != nil {
+				// remove any ExtensionHostSubnet
+				logrus.Infof("Removed node %s extensions HostSubnet", node.Name)
+				err = m.kube.DeleteAnnotationOnNode(node, types.ExtensionHostSubnet)
+			}
+		} else if !sameCIDR(ovnHostsubnet, extHostsubnet) {
+			// sync the ExtensionsHostSubnet with the OVN-assigned one
+			logrus.Infof("Sycned node %s extensions HostSubnet %s", node.Name, ovnHostsubnet.String())
```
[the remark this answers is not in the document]
dcbw: @danwinship Windows nodes will be looking for the extensions hostsubnet value so they know what subnets the Linux nodes are using. I suppose they could theoretically look at the OVN hostsubnet value, but that annotation is kinda internal ovn-kubernetes implementation details. It may well be the case that the Windows side of all this isn't part of ovnkube eventually, so we thought to keep things better abstracted.

go-controller/extensions/pkg/controller/master.go
```diff
@@ -0,0 +1,196 @@
...
+		extHostsubnet, err = subnetAllocator.GetNetwork()
+		if err == netutils.ErrSubnetAllocatorFull {
+			// Current subnet exhausted, check next possible subnet
+			continue
+		} else if err != nil {
+			logrus.Errorf("Error allocating network for node %s: %v", node.Name, err)
+			return
+		}
+		logrus.Infof("Allocated node %s extensions HostSubnet %s", node.Name, extHostsubnet.String())
+	}
```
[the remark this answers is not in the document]
dcbw: @danwinship I agree, this should all be handled better in the allocator. Ported the fixes over from 840 and 841 in any case.

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -235,6 +242,17 @@ func runOvnKube(ctx *cli.Context) error {
...
+
+	enableExtensions := ctx.Bool("enable-extensions")
+	if enableExtensions {
+		if err := extcontroller.StartExtensionSDN(ctx, master != "", node, clientset, factory, stopChan); err != nil {
+			logrus.Errorf(err.Error())
+			panic(err.Error())
+		}
+	}
+
+	if master != "" || node != "" || enableExtensions {
```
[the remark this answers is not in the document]
dcbw: Done.

go-controller/cmd/ovnkube/ovnkube.go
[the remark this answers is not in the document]
danwinship: The name still sounds like it means "enable arbitrary ovn-kubernetes plugins", not "enable VXLAN tunnels to other networks"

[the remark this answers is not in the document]
dcbw: @danwinship have any other suggestions?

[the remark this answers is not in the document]
squeed: vxlan-portal?

go-controller/extensions/pkg/controller/controller.go
```diff
@@ -0,0 +1,59 @@
+package controller
```
squeed: I'm a bit confused - shouldn't this be in `go-controller/pkg/controller/hybrid` or something like that? Why the completely separate codebase?

go-controller/extensions/pkg/controller/controller.go
```diff
@@ -0,0 +1,59 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	extensionClusterSubnetsFlag string = "extension-cluster-subnets"
+)
+
+// StartExtensionSDN starts one or both of the master and node controllers for
+// extension SDN
+func StartExtensionSDN(ctx *cli.Context, master bool, nodeName string,
```
squeed: This function seems kind of unnecessary. How about just `StartMaster` and `StartNode` or something like that?

go-controller/extensions/pkg/controller/master.go
```diff
@@ -0,0 +1,202 @@
...
+		kube: &kube.Kube{KClient: clientset},
+	}
+
+	subnets, err := config.ParseClusterSubnetEntries(clusterSubnets)
+	if err != nil {
+		return nil, err
+	}
+
+	alreadyAllocated := make([]string, 0)
+	existingNodes, err := m.kube.GetNodes()
```
squeed: This seems.. racy. Why can't we just use the shared-informer semantics where we get an Add(Node) every time we start?

go-controller/extensions/pkg/controller/node_linux.go
```diff
@@ -0,0 +1,237 @@
...
+// Delete handles node deletions
+func (n *NodeController) Delete(node *kapi.Node) {
+	cookie := nodeNameToCookie(node.Name)
+	_, _, err := util.RunOVSOfctl("del-flows", hybridBridgeName, fmt.Sprintf("cookie=0x%08x/0xffffffff", cookie))
+	if err != nil {
+		logrus.Debugf("failed to delete flows for node %s: %v", node.Name, err)
+	}
+}
+
+func (n *NodeController) getNodeSubnet(nodeName string) (*net.IPNet, error) {
```
squeed: This doesn't seem to be used?

squeed:
Architecture-wise, running this in a separate process is a bit awkward. It means that you have to hard-code the logic that defines which nodes should be vxlan-attached vs. ovn-attached.

I would rather see that defined as a proper label selector and the processes integrated.

go-controller/extensions/pkg/controller/node_windows.go
```diff
@@ -0,0 +1,318 @@
...
+
+	return n, nil
+}
+
+// Start is the top level function to run hybrid-sdn in node mode
+func (n *NodeController) Start(wf *factory.WatchFactory, stopChan chan struct{}) error {
+	return extutil.StartNodeWatch(n, wf, stopChan)
+}
+
+func validNode(node *kapi.Node) (*net.IPNet, net.IP, net.HardwareAddr) {
```
squeed: This function doesn't seem like it should be build-tag differentiated. It doesn't have any os-level implementation. Also, it has a different signature from the _linux one.

Please rename one of them (and document this).

go-controller/extensions/pkg/controller/node_linux.go
```diff
@@ -0,0 +1,237 @@
...
+
+	kapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	hybridBridgeName  string = "br-hybrid"
+	hybridNetworkName string = "hybridnet"
+	hybridCIDR        string = "100.65.0.1/16"
```
squeed: Hard-coding CIDRs is not the root of all evil, but it is close :-). Can you make this configurable?

Separately, what does this do? Why do we need another IP range, and why does it have to be so big?

go-controller/extensions/pkg/controller/node_linux.go
```diff
@@ -0,0 +1,237 @@
...
+
+	logrus.Infof("found node %s subnet %s", nodeName, subnet.String())
+	return subnet, nil
+}
+
+func (n *NodeController) initGateway(nodeName string) error {
+	hybridName := "hyb-" + nodeName
+	gatewayRouter := "GR_" + nodeName
+
+	hybridLRPMac, stderr, err := util.RunOVNNbctl("--if-exist", "get", "logical_router_port", "rtos-"+hybridName, "mac")
```
squeed: Can this fail if ovn-kube master is behind? Do we need retry logic somewhere?

go-controller/extensions/pkg/controller/controller.go
danwinship: Yeah, I thought that too but then I realized that ovn-kubernetes lets you run the master and node modes in the same process. We should probably get rid of that, but for now, this is consistent with everything else.

go-controller/extensions/pkg/controller/controller.go
dcbw: @squeed it could be, but Rajat originally had things as a completely separate process.

go-controller/extensions/pkg/controller/master.go
dcbw: @squeed we do exactly the same thing in StartClusterMaster()

go-controller/extensions/pkg/controller/node_linux.go
dcbw: fixed.

dcbw:
> Architecture-wise, running this in a separate process is a bit awkward. It means that you have to hard-code the logic that defines which nodes should be vxlan-attached vs. ovn-attached.
> 
> I would rather see that defined as a proper label selector and the processes integrated.

@squeed it's only a separate process on Windows since that doesn't run full ovn-kubernetes. And based on Winships comments, I do have full label selector handling done locally.

go-controller/extensions/pkg/controller/controller.go
```diff
@@ -0,0 +1,68 @@
...
+	return config.ParseClusterSubnetEntries(ctx.String(extensionClusterSubnetsFlag))
+}
+
+// GetExtensionCLIFlags returns standard CLI flags for extensions functionality,
+// and appends any passed-in flags to the list for convenience
+func GetExtensionCLIFlags(flags []cli.Flag) []cli.Flag {
+	return append(flags, []cli.Flag{
+		cli.StringFlag{
+			Name:  extensionClusterSubnetsFlag,
+			Value: "11.128.0.0/16/24",
```
squeed: This is reachable IP space. There shouldn't be a default.

squeed:
Looks like ovn failed to roll out. Something's not working.

go-controller/hybrid-overlay/pkg/controller/master.go
```diff
@@ -0,0 +1,203 @@
...
+
+	_, subnet, err := net.ParseCIDR(sub)
+	if err != nil {
+		return nil, fmt.Errorf("Error in parsing %q hostsubnet: %v", annotation, err)
+	}
+
+	return subnet, nil
+}
+
+func sameCIDR(a, b *net.IPNet) bool {
```
squeed: can you not to `reflect.DeepEqual`?

go-controller/hybrid-overlay/pkg/controller/master.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// NewMaster a new master controller that listens for node events
+func NewMaster(clientset kubernetes.Interface, subnets []config.CIDRNetworkEntry) (*MasterController, error) {
+	m := &MasterController{
+		kube: &kube.Kube{KClient: clientset},
+	}
+
+	alreadyAllocated := make([]string, 0)
+	existingNodes, err := m.kube.GetNodes()
```
squeed: Correct me if I'm wrong, but will GetNodes have its cache sunk at this time? Do we need to worry about partial results here?

squeed: (We could move a decent amount of this logic to the `Sync()` function if not.

squeed:
Question: does this handle host network -> extension pod traffic? And how?

go-controller/hybrid-overlay/pkg/controller/master.go
dcbw: @squeed theoretically, but Sync() cannot return an error and kill the process unless we want to panic(). The reason it's still in the startup part is to ensure we can return an error when setting up our allocator.

But also, yes the cache is sunk. The Factory ensures that caches are synced before it returns from NewWatchFactor() in ovnkube.go.

dcbw:
> Question: does this handle host network -> extension pod traffic? And how?

@squeed no; wasn't clear that was required. You can talk to a Service that is backed by a Windows pod. But you cannot access a Windows pod directly from the node itself. We can add that to the br-ext bridge in a fairly contained manner though if required.

dcbw:
Already merged in a different form from another PR.

PR #25, opened 2019-09-13
bug 1748162 : OVN: make db ports configurable

pecameron:
Added environment variables OVN_NB_PORT, OVN_SB_PORT to support custom
ports. The defaults are 6641, 6642 respectively.

Jira: SDN-353 - OVN: move ports to 9000-9999
https://jira.coreos.com/browse/SDN-353

cherry-pick ovn-org/pr/832

Signed-off-by: Phil Cameron <pcameron@redhat.com>

PR #28, opened 2019-10-02
Rebase to latest ovn-kubernetes upstream

dcbw:
@squeed


danwinship:
So this makes it really hard to see what local changes we have... I think we need to start doing this in a more-controlled fashion. I'm not sure what that more-controlled fashion is though, given that we want to rebase fairly regularly.

Anyway, diffing https://github.com/openshift/ovn-kubernetes/pull/28/commits/8a86cb0d45fd6da6ee8ffae7059f98dff17aabd8 against https://github.com/openshift/ovn-kubernetes/pull/28/commits/940cef0ea9b73a28dab9f12a5f0cccbcf8d5d94e, I see differences in:

- `Dockerfile` and `OWNERS`, which are expected
- `dist/images/Dockerfile.fedora` and `dist/images/iptables-scripts/*`, which are from #10 and are expected.
- `dist/images/ovnkube.sh` which I think is a bad merge. (#25 was modified from its upstream version because we were missing another relevant commit, but that difference should go away now)
- `go-controller/pkg/cluster/gateway_localnet.go`, which is because of #23 and looks correct
- `go-controller/pkg/cni/helper_linux.go`, which is because of #19 and looks correct

dcbw:
@danwinship repushed after inspecting ovnkube.sh differences, there is now no difference with upstream.

danwinship:
/lgtm
but we need to figure out a better way to do this...

squeed:
The failures are almost certainly due to ovnkube.sh incompatabilities with the daemonset. That should go away when https://github.com/openshift/cluster-network-operator/pull/329 merges

squeed:
We then need https://github.com/openshift/origin/pull/23665 and https://github.com/openshift/origin/pull/23896 to merge for CI to go green.

squeed:
So, the ideal solution would allow for
1 - Frequent rebases
2 - Clean history
3 - no force pushes
4 - Carry patches are obvious

Distros have solved this in a slightly ugly but reasonable way: carrying patchfiles. We could do that, and write a bit of tooling to make it pretty transparent. It would be a bit ugly, for sure, but maybe not too bad.  Maybe even have a git submodule checked out to /upstream, patches in /patches, and the rendered version in /rendered.

danwinship:
> So, the ideal solution would allow for
> 1 - Frequent rebases
> 2 - Clean history
> 3 - no force pushes
> 4 - Carry patches are obvious

I don't accept "no force pushes" as being a goal in itself, although it's probably a necessary side effect of "clean history", which I think *is* desirable.

squeed:
I wonder if we can solve this with branches & submodules. Hang on. I have an idea.

squeed:
Okay, this could work. Basically, we create a new branch cut from upstream, re-apply our carry patches, and commit. In master we have a submodule that points just to that branch and is updated any time you want to bump. Also solves the perpetual OWNERS and Dockerfile rebase issues.

squeed:
Looks like the network came up and the tests that failed are ones we expect. Once the test-skip PRs merge, this can go it.

danwinship:
oops, I meant for it to go in now. e2e-aws-ovn-kubernetes isn't blocking anyway
/hold cancel

squeed:
e2e-aws-ovn-kubernetes *is* blocking for this repo. Not for CNO. I might be able to force, hang on.

squeed:
Nope, but you can, @danwinship

danwinship:
oh, huh, the tide status previously just said that it was mergeable except for /hold
/override e2e-aws-ovn-kubernetes

PR #31, opened 2019-10-10
SDN-490: Porting clearance of NetworkUnavailable status from SDN to OVN

alexanderConstantinescu:
Hi 

This is a port of the SDN code for removing the initial `NodeNetworkUnavailable` condition that k8s sets for GCE nodes. 

I am opening this PR as I happened to pick up this user story earlier today, prior to the planning, and implemented it quite quickly. Please have a look and discuss:

* If this is a necessary PR (I understood that @danwinship was not convinced that it was) 
* If this is the right repo, or if it should target upstream instead. 

/assign @danwinship @squeed

### DIFF
--- a/go-controller/pkg/cluster/master.go
+++ b/go-controller/pkg/cluster/master.go
@@ -6,7 +6,9 @@ import (
 	"strings"
 
 	kapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
@@ -237,6 +239,8 @@ func (cluster *OvnClusterController) ensureNodeLogicalNetwork(nodeName string, h
 }
 
 func (cluster *OvnClusterController) addNode(node *kapi.Node) (err error) {
+	cluster.clearInitialNodeNetworkUnavailableCondition(node)
+
 	hostsubnet, _ := parseNodeHostSubnet(node)
 	if hostsubnet != nil {
 		// Node already has subnet assigned; ensure its logical network is set up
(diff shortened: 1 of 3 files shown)


danwinship:
> * If this is a necessary PR (I understood that @danwinship was not convinced that it was)

I was under the impression that we had successfully run ovn-kubernetes on GCP. If we have, then this patch is unnecessary.

There was some attempt to fix upstream to not erroneously set the NetworkUnavailable status, so it's possible that eventually got fixed. Alternatively, something else in OpenShift might be removing that status for us...

> * If this is the right repo, or if it should target upstream instead.

If it is necessary, it should go upstream, as it's a general "third-party network plugins on GCP" problem, not an OpenShift problem.

danwinship:
> There was some attempt to fix upstream to not erroneously set the NetworkUnavailable status

@dcbw may remember details and have pointers to this...

squeed:
That effort upstream was never merged. We still have to patch nodes to fix this.

I know it's insane. We should upstream this.

go-controller/pkg/cluster/master.go
```diff
@@ -329,10 +333,56 @@ func (cluster *OvnClusterController) deleteNode(nodeName string, nodeSubnet *net
 	return nil
 }
 
+// As for openshift-sdn, OVN uses an overlay and doesn't need GCE Routes, we need to
+// clear the NetworkUnavailable condition that kubelet adds to initial node
+// status when using GCE.
+// TODO: make upstream kubelet more flexible with overlays and GCE so this
+// condition doesn't get added for network plugins that don't want it, and then
+// we can remove this function.
```
danwinship: If we can find the upstream bug, discussion, we should link to it here

go-controller/pkg/kube/kube.go
```diff
@@ -17,6 +17,7 @@ import (
 type Interface interface {
 	SetAnnotationOnPod(pod *kapi.Pod, key, value string) error
 	SetAnnotationOnNode(node *kapi.Node, key, value string) error
+	SetUpdateStatusOnNode(*kapi.Node) error
```
danwinship: Shouldn't have "Set" in the name. Just "`UpdateNodeStatus`" or "`UpdateStatusOnNode`"

go-controller/pkg/cluster/master.go
squeed: The upstream bug (and @dcbw)'s fix got timed out (https://github.com/kubernetes/kubernetes/pull/34398). All that remains is a big ugly TODO in the upstream code:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_network.go#L46

It is possible that out-of-tree cloudproviders have fixed this.

squeed:
The upstream bug (and @dcbw)'s fix got timed out (https://github.com/kubernetes/kubernetes/pull/34398). All that remains is a big ugly TODO in the upstream code:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_network.go#L46

It is possible that out-of-tree cloudproviders have fixed this, but I've not paid attention to the space for a bit.

alexanderConstantinescu:
So, I've finally been able to confirm that we cannot create clusters on GCP using OVN (due to the NetworkUnavailable condition being set on nodes). I will take your comments into account @danwinship and create a PR upstream on ovn-org.

PR #32, opened 2019-10-14
add prometheus capability to ovn-kubernetes

JacobTanenbaum:
This PR does not add any metrics but does allow the metrics to be exposed from within ovn-kubernetes to a prometheus server.

along with cluster-network-operator PR: openshift/cluster-network-operator#318

will allow metrics to be added

cherry-picks of ovn-kubernetes commits from PR https://github.com/ovn-org/ovn-kubernetes/pull/837


pecameron:
@JacobTanenbaum is this upstream or going to be upstream?

PR #41, opened 2019-10-30
gateway/localnet: work around disappearing br-nexthop IP address

dcbw:
Something is removing the IP address from br-nexthop which of
course makes things go south. Use a giant hammer to add it back.

@squeed @danwinship @JacobTanenbaum


squeed:
Merging this so we can make some progress. We can track this in https://bugzilla.redhat.com/show_bug.cgi?id=1766356

PR #47, opened 2019-11-01
Wrap IPv6 address in [] before adding port

russellb:
This code assumed an IPv4 address.  When the IP address is IPv6, we
must wrap it with square brackets to differentiate the address from
the : that separates the address from the port number.

This is the openshift equivalent of https://github.com/ovn-org/ovn-kubernetes/pull/901


danwinship:
Upstream fix will be pulled in via #57

PR #56, opened 2019-11-08
[wip] Parallel pod node events

dcbw:


squeed:
I'm not sure that we should be having an unbounded number of worker processes; pod creation is pretty quick these days, and it seems like we might hammer ovn-northd.

What, instead, if we had a fixed number of workers (10) and distributed events to then based on a hash of the NamespacedName? That way we'd always have individual pod actions serialized, but no unbounded worker growth, and, best of all: no locks anywhere!

PR #58, opened 2019-11-10
ovn: retry pod setup in update if it fails during add

dcbw:
If a failure occurred during a pod Add the error would be logged
but setup would never be retried because the pod would remain
scheduled. Instead, use the 'ovn' annotation as a marker of
whether the pod has been set up or not. Since setting the annotation
should be the last thing we do, we can mostly use this as an
indicator of setup success.

### DIFF
--- a/go-controller/pkg/ovn/ovn.go
+++ b/go-controller/pkg/ovn/ovn.go
@@ -330,25 +330,49 @@ func (oc *Controller) ovnControllerEventChecker(stopChan chan struct{}) {
 	}
 }
 
+func podScheduledAndNetworked(pod *kapi.Pod) bool {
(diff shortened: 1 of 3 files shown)


dcbw:
Failed tests seem like flakes and not network related:
[sig-apps] StatefulSet [k8s.io] Basic StatefulSet functionality [StatefulSetBasic] should perform rolling updates and roll backs of template modifications with PVCs [Suite:openshift/conformance/parallel] [Suite:k8s]
[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: dir] [Testpattern: Pre-provisioned PV (default fs)] volumes should store data [Suite:openshift/conformance/parallel] [Suite:k8s]

PR #61, opened 2019-11-13
[wip] unidling: only keep empty LB for service when it has been idled

dcbw:
Service annotations indicate when the service has been idled, so
only keep the empty LB for those services. Otherwise services with
no endpoints time out rather than have requests rejected as expected.

Fixes failures of the OpenShift/Kubernetes e2e test for:

[sig-network] Services should be rejected when no endpoints exist


dcbw:
Went upstream instead.

PR #63, opened 2019-11-16
Upstream merge 2019-11-15 + latest hybrid overlay

dcbw:
We had most of the big items in here already, in slightly earlier forms (SDN-410/gateway, hybrid overlay, gcp annotation, retry pods on Update, unidling empty LB thing, metrics). The largest change is probably the move to go mod.

@danwinship @squeed


squeed:
I see a `go-controller/hybrid-overlay/pkg/controller/overlay_windows.go.orig` - that seems like an oopsie.

dcbw:
@squeed fixed

PR #65, opened 2019-11-22
Lock in version of ovs/ovn

pecameron:
Openshift 4.3 requires:
ovs-2.12-2.12.0
ovn2.11-2.11.1


pecameron:
@dcbw PTAL

dcbw:
@pecameron I guess we have to track what's most current in FDP beta; that would be openvswitch2.12-2.12.0-8.el7fdp

danwinship:
@pecameron obsoleted by #72?

pecameron:
No longer needed since ovn 2.12 is now GA. This maybe reopened when we have another beta version.

PR #66, opened 2019-12-03
CARRY: honor openshift-sdn legacy annotations for idling and multicast

dcbw:
@squeed @danwinship @rcarrillocruz @alexanderConstantinescu @JacobTanenbaum @pecameron @abhat


danwinship:
Did this requirement come from somewhere? We have separate openshift-sdn and ovn-kubernetes docs anyway, so it doesn't simplify the documentation (much), and we don't support migration (and if we did we could easily migrate the annotations), so... ?

dcbw:
> Did this requirement come from somewhere? We have separate openshift-sdn and ovn-kubernetes docs anyway, so it doesn't simplify the documentation (much), and we don't support migration (and if we did we could easily migrate the annotations), so... ?

@danwinship just though it would be nice-to-have. But if we don't need it, I'm happy to drop it.

danwinship:
fewer carry patches seems better to me... but we can remember this PR is here if we decide we need it later

PR #68, opened 2019-12-10
Create separate image for Windows binary artifacts

yselkowitz:
This would have to be paired with a new container request and PR to ocp-build-data (which I can also provide).  How these binaries would be extracted and distributed to users would be up to ART.

CC: @dcbw

### DIFF
--- a/Dockerfile.artifacts
+++ b/Dockerfile.artifacts
@@ -1,0 +1,22 @@
+# This Dockerfile builds an image containing the Windows ovn binaries
+# layered on top of the Linux ovn-kubernetes image.
+FROM openshift/origin-release:golang-1.12 AS builder
+
+WORKDIR /go-controller
+COPY go-controller/ .
+
+# build the binaries
+RUN CGO_ENABLED=0 make windows
+
+FROM openshift/origin-ovn-kubernetes
+
+COPY --from=builder /go-controller/_output/go/bin/windows/ovnkube.exe /usr/share/openshift/windows/
+COPY --from=builder /go-controller/_output/go/bin/windows/hybrid-overlay.exe /usr/share/openshift/windows/
+COPY --from=builder /go-controller/_output/go/bin/windows/ovn-k8s-cni-overlay.exe /usr/share/openshift/windows/
+
+LABEL io.k8s.display-name="ovn kubernetes" \
+      io.k8s.description="This is a component of OpenShift Container Platform that provides an overlay network using ovn." \
+      summary="This is a component of OpenShift Container Platform that provides an overlay network using ovn." \
+      io.openshift.tags="openshift" \
+      maintainer="Phil Cameron <pcameron@redhat.com>"
+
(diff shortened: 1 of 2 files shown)


yselkowitz:
/hold
still waiting to hear if this is the route you want to go for providing these or not.

dcbw:
@yselkowitz it was our understanding that we could do something like https://github.com/openshift/ovn-kubernetes/pull/69 and ART could just pull the EXE out of the image? At least that was my understnading of the conversation last week...

yselkowitz:
Do you really want to increase the size of ovn-kubernetes solely for the benefit of a Windows binary which is only present to be extracted and published elsewhere?  That's why I suggested a separate image for that purpose.

dcbw:
@yselkowitz sure, a separate image would not have that problem. I have no idea what impact that has on ART's process though. It was clear to me that it was trivial for ART to extract the Windows binary from the existing image. But it's much less clear to me what all needs to be done for a new image.

If ART is fine with a new image then sure.

Dockerfile.artifacts
```diff
@@ -0,0 +1,22 @@
...
+
+WORKDIR /go-controller
+COPY go-controller/ .
+
+# build the binaries
+RUN CGO_ENABLED=0 make windows
+
+FROM openshift/origin-ovn-kubernetes
+
+COPY --from=builder /go-controller/_output/go/bin/windows/ovnkube.exe /usr/share/openshift/windows/
```
dcbw: @yselkowitz we don't need ovnkube.exe or ovn-k8s-cni-overlay.exe packaged. The only thing we need in the image is the `hybrid-overlay.exe`

go-controller/hack/build-go.sh
```diff
@@ -35,11 +35,11 @@ build_windows_binaries() {
     set -x
     for bin in "$@"; do
         binbase=$(basename ${bin})
-        GOOS=windows go build -v \
+        GOOS=windows GOARCH=amd64 go build -v \
```
dcbw: @yselkowitz These changes should go upstream

yselkowitz: Filed: https://github.com/ovn-org/ovn-kubernetes/pull/995

yselkowitz: Merged upstream.

Dockerfile.artifacts
```diff
@@ -0,0 +1,22 @@
+# This Dockerfile builds an image containing the Windows ovn binaries
+# layered on top of the Linux ovn-kubernetes image.
+FROM openshift/origin-release:golang-1.12 AS builder
+
+WORKDIR /go-controller
+COPY go-controller/ .
+
+# build the binaries
+RUN CGO_ENABLED=0 make windows
```
trozet: @yselkowitz do you need to rebase and remove the windows building from #69 ?

yselkowitz:
Since #69 has since been merged, the second half of this is moot.  Closing this and filing a separate PR for just the first commit (which is now upstream).

PR #75, opened 2020-01-09
Only build Windows binary on supported platforms

trozet:
Patch changes docker building to only build windows on x86 platforms.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,15 @@ COPY go-controller/ .
 
 # build the binaries
 RUN CGO_ENABLED=0 make
-RUN CGO_ENABLED=0 make windows
+
+# only build windows on supported platforms
+# otherwise create a dummy windows binary to allow subsequent COPY to still work
+RUN if [[ $(arch) =~ ^(x86_64|i386)$ ]]; then \
+      CGO_ENABLED=0 make windows; \
+    else \
+      mkdir -p /go-controller/_output/go/bin/windows/ && \
+      touch /go-controller/_output/go/bin/windows/hybrid-overlay-noop; \
+    fi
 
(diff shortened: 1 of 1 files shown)


trozet:
@dcbw Windows is supported on ARM architectures as well...not sure if those should be supported as well here?

trozet:
I can also remove the windows directory after the noop copy, if we dont want a noop file/windows directory to even exist at all on unsupported arch. Just let me know.

danwinship:
do we have any way to test that this actually does fix the s390x build?

trozet:
@danwinship I tested it in a negative scenario by using an unsupported arch and it built just fine.

Dockerfile
```diff
@@ -16,7 +16,15 @@ COPY go-controller/ .
 
 # build the binaries
 RUN CGO_ENABLED=0 make
-RUN CGO_ENABLED=0 make windows
+
+# only build windows on supported platforms
+# otherwise create a dummy windows binary to allow subsequent COPY to still work
+RUN if [[ $(arch) =~ ^(x86_64|i386)$ ]]; then \
```
danwinship: ugh... this is really ugly and I don't have any good suggestion on how to improve it, except that maybe this means we should not be building the windows binary here (since we don't actually need it to be in this image anyway and are only doing that as a hack).

@dcbw?

danwinship:
/lgtm
/approve
/hold

if we don't come up with a better fix feel free to "`/hold cancel`" to get this to merge

Dockerfile
trozet: there is a PR to move it into its own image: #68

trozet:
closing in favor of #76

PR #78, opened 2020-01-16
Dynamic HA - `--no-leader-only` for ovn commands

alexanderConstantinescu:
This is a complementary PR to 

https://github.com/openshift/cluster-network-operator/pull/447

Please see that one for a explanation on why this change is needed


danwinship:
/hold
there's ongoing discussion in https://github.com/openshift/cluster-network-operator/pull/447

PR #79, opened 2020-01-16
[wip] asdfadsf

dcbw:


pecameron:
@dcbw Seems like aws is having problems. There are no nodes in the artifacts. I am having this problem on PR72 as well.

dcbw:
`Error: Error waiting for instance (i-0be5b8c6803122077) to become ready: Failed to reach target state. Reason: Server.InternalError: Internal error on launch`

/retest

PR #80, opened 2020-01-22
Add windows nodes to ovn's node watcher

JacobTanenbaum:
Currently the ovn controller treats pods on windows and linux nodes the same. By adding a windows node cache the logic can be implemented to treat them differently. 

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1789881
upstream PR: https://github.com/ovn-org/ovn-kubernetes/pull/1015

### DIFF
--- a/go-controller/cmd/ovnkube/ovnkube.go
+++ b/go-controller/cmd/ovnkube/ovnkube.go
@@ -24,8 +24,6 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 
-	"k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kexec "k8s.io/utils/exec"
 )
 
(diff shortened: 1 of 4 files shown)


JacobTanenbaum:
@dcbw @knobunc PTAL

abhat:
@JacobTanenbaum just for my understanding, why are we not fixing this upstream first? The PR you had open upstream is closed?

JacobTanenbaum:
@abhat the code for windows has not merged upstream so it is not applicable

dcbw:
@JacobTanenbaum what we have been doing upstream is forking my personal ovnkube repo and then doing PRs against my 'extensions' branch, which I then merge into my 'extensions' branch and then push the upstream PR. A bit complicated but it works.

In any case, it's probably OK doing the PR here for review, but when it's ready I will cherry-pick the patches and put them into my 'extensions' branch upstream.

dcbw:
@JacobTanenbaum so similar to what you're doing here, but in a more generic way, we can use "did we assign a subnet to this node or not" logic as the decision on whether to wait for the node switch when a pod is added.

We can use oc.logicalSwitchCache for this too. I have a refactor of that in https://github.com/ovn-org/ovn-kubernetes/pull/1023 that you could base your stuff off. The missing bit is that neither my PR or the original code has the info about explicitly not assigning a subnet, but that is easy to fix by adding an element to the oc.logicalSwitchCache map with a `nil` value and then using the 'ok' pattern to see if the value exists in the map or not. If it exists and is nil, then we didn't assign a subnet and there's no point in running addLogicalPort(). If it doesn't exist in the map, then a decision hasn't been made yet and we should wait.

Make sense?

PR #83, opened 2020-01-28
Bug 1796531: CARRY: fix race updating exclude IPs on node switch

dcbw:
When nodes are added to the cluster, it may be that our pod add
handler is waiting for the node switch to show up. It might win
the race and create the pod between our node switch being
created in master.go and the exclude IPs being updated immediately
after. Fix this by reverting to setting the exclude IPs during
logical switch creation.

This is essentially a revert of part of 806dff11

Fixes: 80d6dff11983a4358693eef06467b1390e96e880

### DIFF
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -474,10 +474,7 @@ func ParseNodeHostSubnet(node *kapi.Node) (*net.IPNet, error) {
 }
 
 func (oc *Controller) ensureNodeLogicalNetwork(nodeName string, hostsubnet *net.IPNet) error {
-
-	// Get firstIP for gateway.  Skip the second address of the LogicalSwitch's
-	// subnet since we set it aside for the management port on that node.
-	firstIP, _ := util.GetNodeWellKnownAddresses(hostsubnet)
+	firstIP, secondIP := util.GetNodeWellKnownAddresses(hostsubnet)
 	nodeLRPMac := util.IPAddrToHWAddr(firstIP.IP)
 	clusterRouter := util.GetK8sClusterRouter()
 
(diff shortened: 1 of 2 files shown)


dcbw:
Failures are supposedly due to https://github.com/openshift/installer/pull/3007 (per abhinav)

/test e2e-aws-ovn

go-controller/pkg/ovn/master.go
```diff
@@ -477,7 +477,7 @@ func (oc *Controller) ensureNodeLogicalNetwork(nodeName string, hostsubnet *net.
 
 	// Get firstIP for gateway.  Skip the second address of the LogicalSwitch's
 	// subnet since we set it aside for the management port on that node.
-	firstIP, _ := util.GetNodeWellKnownAddresses(hostsubnet)
+	firstIP, secondIP := util.GetNodeWellKnownAddresses(hostsubnet)
```
rcarrillocruz: you probably want to change the comment before this line since you are now grabbing secondIP

dcbw:
Now that https://github.com/openshift/origin/pull/24477 is in...

/test e2e-aws-ovn

go-controller/pkg/ovn/master.go
```diff
@@ -489,19 +486,37 @@ func (oc *Controller) ensureNodeLogicalNetwork(nodeName string, hostsubnet *net.
 		return err
 	}
 
+	// Create a logical switch and set its subnet. If all cluster subnets are
+	// big enough (/24 or greater), exclude the hybrid overlay port IP (even
+	// if hybrid overlay is not enabled) to allow enabling hybrid overlay
+	// in a running cluster without disrupting nodes.
+	var excludeIPs string
+	if !config.IPv6Mode {
+		excludeIPs = secondIP.IP.String()
```
abhat: nit: initialize `excludeIPs` to `"other-config:exclude_ips="` and then on line 495 do `excludeIPs += secondIP.IP.String()`

dcbw: Good catch, fixed.

dcbw:
KubeAPILatencyHigh alert for AWS, unrelated flakes for GCP.

/retest

dcbw:
Flake https://bugzilla.redhat.com/show_bug.cgi?id=1796739

/retest

PR #85, opened 2020-02-07
Merge upstream ovn-kubernetes 

abhat:
Cherrypick:

Commit: e7abf1283e68bb95d0897cecc2ea6caff87c162e
Commit: c372bb0e5ecf7e228fedf228fd3944421abf3f4b
Commit: 3051ccac6d62bd1dc1b7c44c27fa92e47e1e24de
Commit: 4e4da057569e3a96806a119f51ba3466c208d815

### DIFF
--- a/go-controller/pkg/config/config.go
+++ b/go-controller/pkg/config/config.go
@@ -23,6 +23,7 @@ import (
 const DefaultEncapPort = 6081
 
 const MetricNamespace = "ovnkube"
+const DefaultAPIServer = "http://localhost:8443"
 
 // The following are global config parameters that other modules may access directly
 var (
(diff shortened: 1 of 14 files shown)


danwinship:
The "parallelize Node events" PR has problems and shouldn't be cherry-picked yet. (See https://github.com/ovn-org/ovn-kubernetes/pull/1043)

dcbw:
@abhat we haven't settled on what `Fixes accidental override of file conf settings with defaults` should actually be doing yet. If we pull it in before it's actually upstream it's going to be a pain to deal with in a future merge.

abhat:
@danwinship I think somehow github is picking that commit because of the history of changes to that file. The eventual change to the ovn.go file doesn't have @dcbw's changes: https://github.com/openshift/ovn-kubernetes/blob/af6549312c8d53dc0ef426ceb1c23c76fc078a67/go-controller/pkg/ovn/ovn.go 

@dcbw, @danwinship wanted me to pick that change so we could unblock testing with the expectation that when we fix it the right way, it will be a matter of reverting this commit out and then replaying the other commit.

danwinship:
> I think somehow github is picking that commit because of the history of changes to that file.

How did you generate this PR? You should be just creating a branch off master and then doing a series of `git cherry-pick` commands to get the commits you want. That should not pull in any commits that you did not explicitly ask for.

abhat:
Right, that's exactly what I did. As such the files look like they don't have the changes from that commit.

PR #88, opened 2020-02-12
Bug 1798613: OVN: ovn2.12 x86_64 build hack

pecameron:
The s390x and ppcle64 ovn2.12 rpms are not available in fast datapath.
This allows the ocp-4.4 ovn-kubernetes image to be built.

When the needed rpms are added to fastdatapath, this change can be
backed out.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,11 +32,11 @@ RUN yum install -y  \
 	selinux-policy && \
 	yum clean all
 
-RUN INSTALL_PKGS=" \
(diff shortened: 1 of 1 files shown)


pecameron:
@dcbw @danwinship @jupierce PTAL

pecameron:
We found a better way to solve this problem. The missing s390x and ppcle64 rpms have been pushed and are now present in the FDP-next repo.

PR #92, opened 2020-02-17
Fix setting of other-config:exclude_ips

danwinship:
Fixes a bug in the merging of the hybrid overlay code with upstream ipv6 changes: We can't just pass `""` to `RunOVNNbctl()` because it actually gets passed as a 0-length argument, resulting in

    ovn-nbctl: : missing column name

/assign @dcbw


dcbw:
Obsoleted by https://github.com/openshift/ovn-kubernetes/pull/93

danwinship:
This was not obsoleted by #93; the code is still there. (This is not an upstream fix, it's a fix to the hybrid overlay integration.)

danwinship:
oops, no, I was looking at release-4.4, not master

PR #94, opened 2020-02-19
HACK: add ipv4 interfaces to pods that need them

danwinship:
This is a hack we originally did for IPv6 on AWS on 4.3, where lots of pods needed IPv4 access in order to make anything work. We were hoping to not need it for Azure, but at least for the moment DNS is broken without it. We may be able to kill it in the future. Note that it completely no-ops unless you're in a single-stack IPv6 cluster.

I left in the AWS hack list in case we need it again in the future...

Note that we do not plan to ever support single-stack IPv6 for customers on cloud platforms so the hardcoded IP range doesn't really matter.

/assign @dcbw 
/cc @russellb

### DIFF
--- a/go-controller/pkg/cni/cnishim.go
+++ b/go-controller/pkg/cni/cnishim.go
@@ -125,6 +125,11 @@ func (p *Plugin) CmdAdd(args *skel.CmdArgs) error {
 		}
 	}
 
+	err = maybeAddIPv4Hack(args, result)
+	if err != nil {
+		return fmt.Errorf("failed to add hack IPv4 interface: %v", err)
+	}
+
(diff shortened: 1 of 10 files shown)


danwinship:
Turns out, nope, this is just as necessary with Azure as it was with AWS; the Azure APIs are also only available via IPv4

danwinship:
/retest

/assign @knobunc 
@dcbw had several days to scream in horror at this PR and he failed to, except he also failed to lgtm it, and we need it to merge

knobunc:
Thanks.  I hate it.

But I understand why we need it :-)

/lgtm

russellb:
> Thanks. I hate it.

lol! <3

PR #111, opened 2020-03-03
Revert "cluster.go: Disable Southbound DB conditional monitoring."

pecameron:
This reverts commit 799df52e72d24d77d5531e0961609ae9e43ca8b6.


pecameron:
@dcbw @danwinship PTAL

danwinship:
@dcbw knows the details of what we're doing here so I'll defer to him.

If the upstream code fails with some versions of OVN though should we revert/fix it upstream?

dcbw:
@pecameron can you backport for 4.5 as well please?

PR #114, opened 2020-03-05
Bug 1810817: [4.3] CARRY: ovn: fix cloud load balancer rules for IPv6

danwinship:
Backport of #112


danwinship:
/close
This is only useful if cloud LoadBalancers work, which they don't, and they're not going to be fixed in 4.3

PR #121, opened 2020-03-19
WIP: test latest 2.13 build

alexanderConstantinescu:
Need to test 

https://bugzilla.redhat.com/show_bug.cgi?id=1805248

DO NOT MERGE


alexanderConstantinescu:
Hey @dcbw : these new 2.13 OVN binaries are compiled for RHEL8 and it will only support RHEL 8 according to @numansiddique . 

I don't understand what the heck is going with this build and why it refuses to realize that libssl.so.1.1 exists since we do `yum install openssl`. But I just wanted to raise the point as ` openshift/origin-base` is RHEL7 based

danwinship:
> I don't understand what the heck is going with this build and why it refuses to realize that libssl.so.1.1 exists since we do `yum install openssl`

iirc it's `libssl.so.1.0` on RHEL 7. Or something like that. Anyway "RHEL 7 vs RHEL 8"

alexanderConstantinescu:
> > I don't understand what the heck is going with this build and why it refuses to realize that libssl.so.1.1 exists since we do `yum install openssl`
> 
> iirc it's `libssl.so.1.0` on RHEL 7. Or something like that. Anyway "RHEL 7 vs RHEL 8"

Yeah it's libssl.so.1.1 on RHEL 8. But given that I just changed our image to `registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-base:ubi8` (which is RHEL 8) and installed openssl on that, I would expect the damn thing to work fine...gahhhh

alexanderConstantinescu:
closing this as I see: https://github.com/openshift/ovn-kubernetes/pull/122 already builds with:

```
---> Package openvswitch2.13-devel.x86_64 0:2.13.0-0.20200121git2a4f006.el7fdn will be installed
---> Package ovn2.13.x86_64 0:2.13.0-7.el7fdn will be installed
---> Package ovn2.13-central.x86_64 0:2.13.0-7.el7fdn will be installed
---> Package ovn2.13-host.x86_64 0:2.13.0-7.el7fdn will be installed
---> Package ovn2.13-vtep.x86_64 0:2.13.0-7.el7fdn will be installed
```

Which builds with a newer release and more importantly RHEL 7

PR #127, opened 2020-03-30
Bug 1819888: [4.3] bump OVS/OVN to 2.13

dcbw:


dcbw:
gcp failure is legit, but not a regression... https://github.com/ovn-org/ovn-kubernetes/issues/1202

PR #129, opened 2020-04-02
Bug 1819881: [4.4] bump OVS/OVN to 2.13

dcbw:


dcbw:
Latest test failures are not related to networking; aws-ovn failure was console-operator failure with "bad tls certificates"

PR #130, opened 2020-04-03
OWNERS: add bugzilla info

danwinship:


dcbw:
Some of the failures should be "fixed" by https://github.com/openshift/origin/pull/24833

PR #132, opened 2020-04-07
Bug 1814712: Removing node bring up debug log level

alexanderConstantinescu:
This is apparently not needed as it's been removed in release-4.5 and master, and upstreams. So let's just remove it.

### DIFF
--- a/go-controller/pkg/ovn/ovn.go
+++ b/go-controller/pkg/ovn/ovn.go
@@ -158,10 +158,6 @@ func NewOvnController(kubeClient kubernetes.Interface, wf *factory.WatchFactory,
 
 // Run starts the actual watching.
 func (oc *Controller) Run(stopChan chan struct{}) error {
-	// Setting debug log level during node bring up to expose bring up process.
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/ovn.go
```diff
@@ -158,10 +158,6 @@ func NewOvnController(kubeClient kubernetes.Interface, wf *factory.WatchFactory,
 
 // Run starts the actual watching.
 func (oc *Controller) Run(stopChan chan struct{}) error {
-	// Setting debug log level during node bring up to expose bring up process.
-	// Log level is returned to configured value when bring up is complete.
-	logrus.SetLevel(5)
```
danwinship, line 163: There is presumably code later that does the "return to configured value" part that should also be removed?

alexanderConstantinescu: Nope, that's the entire problem.

danwinship: lol oops

PR #135, opened 2020-04-11
Bug 1823557: Upstream merge 2020-04-20

dcbw:
4.4 version of #123 and https://github.com/openshift/ovn-kubernetes/pull/141

@openshift/networking

PR #138, opened 2020-04-14
WIP NetworkPolicy/Namespace locking test

danwinship:
#135 plus https://github.com/ovn-org/ovn-kubernetes/pull/1244, to see if it fixes the flaky e2e test
/cc @trozet


danwinship:
oops, misclick. anyway, all three e2e tests failed to bring up a cluster, but it looks like #135 has the same problem so this isn't the bad commit

PR #144, opened 2020-04-22
cni: update MCS-blocking code to do both IPv4 and IPv6

danwinship:
noticed by @celebdor 
/assign @dcbw 

Also fixed failure-to-do-MCS-blocking to be a fatal pod error rather than just a warning. And for the same reason, I made the IPv4 and IPv6 sections conditional on whether we are using IPv4/IPv6, since you don't want to block IPv4-only pods because we couldn't create ip6tables rules.

### DIFF
--- a/go-controller/pkg/cni/helper_linux.go
+++ b/go-controller/pkg/cni/helper_linux.go
@@ -20,6 +20,7 @@ import (
 	"github.com/vishvananda/netlink"
 
 	"k8s.io/apimachinery/pkg/util/wait"
+	utilnet "k8s.io/utils/net"
 )
 
 func renameLink(curName, newName string) error {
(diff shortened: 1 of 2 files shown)


celebdor:
I saw that openshift/sdn blocks it at the node level for hostnetworking pods. Is it something ovn-kubernetes should be covering for as well?

go-controller/pkg/cni/helper_linux.go
```diff
@@ -309,15 +309,39 @@ func (pr *PodRequest) ConfigureInterface(namespace string, podName string, ifInf
...
-			if err != nil {
-				return fmt.Errorf("could not set up pod iptables rules: %s", string(out))
+			if hasIPv4 {
+				out, err := exec.Command("iptables", args...).CombinedOutput()
+				if err != nil {
+					return fmt.Errorf("could not set up pod iptables rules: %s", string(out))
+				}
+			}
+			if hasIPv6 {
+				out, err := exec.Command("ip6tables", args...).CombinedOutput()
```
squeed, line 331: You're getting lucky here, because ipv4 addresses (line ~233) also happen to parse as ipv6 addresses. I think you'll need separate list of family-agnostic and -non-agnostic rules.

danwinship: @squeed fixed. can you re-review?

PR #149, opened 2020-04-27
[WIP] Create OVS user/group to match what is used in RHCOS

mccv1r0:
Create OVS user/group to match what is used in RHCOS


squeed:
Hey @mccv1r0, can we close this?

PR #154, opened 2020-05-04
Add a CNO kind job to CI

nerdalert:
- Runs the CNO kind deployment against the ovn-kubernetes PR on push/merge

Signed-off-by: Brent Salisbury <bsalisbu@redhat.com>

### DIFF
--- a/.github/workflows/cno-ci.yml
+++ b/.github/workflows/cno-ci.yml
@@ -1,0 +1,131 @@
+name: ovn-cno-ci
+
+on:
+  push:
+  pull_request:
+    branches: [ master ]
+
+env:
+  GO_VERSION: 1.13.4
+  K8S_VERSION: v1.17.2
+  KIND_CLUSTER_NAME: ovn
+  BUILD_CNO: true
+  BUILD_OVN: true
+
+jobs:
+  k8s:
+    name: Build k8s
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Set up environment
+        run: |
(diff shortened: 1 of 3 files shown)


danwinship:
Why? We already run tests against a full OCP cluster...

nerdalert:
@danwinship this PR was based off conversations with @trozet and @dave-tucker. Sounded like some bugs had slipped through with lack of CNO testing.

danwinship:
Maybe they meant we should add this *upstream*?

nerdalert:
@danwinship upstream was discussed, I think we thought it might be kind of weird having OCP specific testing in the upstream OSS. Maybe we can flesh it out on tomorrow's standup. Ty for taking a look.

dave-tucker:
Yeah we don't need this as I think `prow` takes care of running integration against OCP.
What we need is automation to keep this repo up-to-date with upstream so that integration is being done often enough to catch issues, and not only when we decide to manually sync the two.

trozet:
Yeah this patch needs to be upstream in ovn-org.

@danwinship the point of this is because patches get merged upstream and then we find out later CNO needed missing changes and now things are broken downstream. Then the person trying to merge downstream has to go back and search numerous commits and try to figure out what happened.

Since CNO works with KIND, we can have a simple job here to run CNO+KIND deployment and catch those errors and label the PR with "needs-CNO-change" or something so that the author or committer needs to go downstream and link a corresponding PR for the required change.

PR #164, opened 2020-05-14
Bug 1830370: lock OVS version to 2.13.0-10 to ensure signed RPMs

dcbw:
We need signed OVN RPMs which are in the FDN repos, but those repos
also include OVS 2.13.0-17 which is unsigned and which we don't yet
need. So we need to lock both OVS and OVN versions to get the right
signed RPMs for now.

@sdodson


sdodson:
/retitle Bug 1830370: lock OVS version to 2.13.0-10 to ensure signed RPMs

PR #167, opened 2020-05-15
Bug 1835985: [4.4] hybrid external gw

trozet:


eparis:
/lgtm cancel
I think 57 straight CI failures is probably at the level some effort should be put into root causing what's going on. What does the team think?

eparis:
Also #169 passe aws ovn on the first try.

trozet:
@eparis I think the failures were other issues with 4.4 CI, which maybe are now fixed which is why this finally passed aws and your commit did as well? We can run retest again on aws now and see if it passes again. Otherwise if it it is still flaky then I'll take a look at it.

/test e2e-aws-ovn

rcarrillocruz:
Been looking at failures, seeing random infra related issues from Terraform and quite a bit of storage jobs failures.
If it was a problem with this patch I would assume networking jobs failing all over.
I'm launching a 4.4 cluster with hybrid overlay enabled and will switch to a build of OVN-K based off this PR.
Will report back.

PR #170, opened 2020-05-20
Bug 1759338: hostnetwork pod can access MCS port 22623 or 22624 on master

rcarrillocruz:
fixes #1759338

### DIFF
--- a/go-controller/pkg/node/block_mcs.go
+++ b/go-controller/pkg/node/block_mcs.go
@@ -1,0 +1,26 @@
+// +build linux
+
+package node
+
+func generateBlockMCSRules(rules *[]iptRule) {
+	*rules = append(*rules, iptRule{
+		table: "filter",
+		chain: "FORWARD",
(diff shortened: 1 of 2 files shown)


danwinship:
The details look right... I assume it's tested?

It might be better to try to isolate the code a bit more to try to avoid causing conflicts with upstream merges

rcarrillocruz:
By isolating, you mean putting the new iptables rules into its own function?

danwinship:
> By isolating, you mean putting the new iptables rules into its own function?

Yes, and not just that, but putting that function and function call into a place where they're unlikely to cause merge conflicts.

eg for the IPv4-in-IPv6 hack (#94), almost all of the new code is in its own file. We do that with a lot of our carried kubernetes patches too. eg, https://github.com/openshift/kubernetes/commit/d486b072. (In that case there are no modifications to the existing files at all, because we're just adding new public methods that can be used to access existing private methods/fields.)

PR #176, opened 2020-06-03
using to launch cluster-bot with more debugging

JacobTanenbaum:
Temporarily increase addressSet debugging level to help debug stress Test

it is very hard to debug the ovn networking stress tests with the current level of debugging.
All the runs in the ovn-controller seem to have the same parsing error where it seems an
addressSet is being used before it is actually created

try to confirm this

**- How to verify it**


squeed:
Hey @JacobTanenbaum can we close this?

PR #179, opened 2020-06-09
Bug 1787434: Cleanup Conntrack when endpoints and pods are deleted 

JacobTanenbaum:

When pods and endpoints are deleted if there are ongoing networking connections there will be a stale conntrack entry. This PR ensures that those conntrack entries are cleaned up 

**- How to verify it**

See the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1787434) for reproduction information

Add cleanup of pods and endpoints conntrack entries


go-controller/pkg/util/net_linux.go
```diff
@@ -167,3 +169,25 @@ func LinkNeighExists(link netlink.Link, neighIP net.IP, neighMAC net.HardwareAdd
 	}
 	return false, nil
 }
+
+func DeleteConntrack(ip string) {
+	ipAddress := net.ParseIP(ip)
+	if ipAddress == nil {
+		klog.Errorf("Value \"%s\" passed to DeleteConntrack is not an IP address", ipAddress)
```
dcbw, line 176: @JacobTanenbaum should be a return here, right?

JacobTanenbaum: fixed

JacobTanenbaum:
@dcbw added my follow up patch

squeed:
Hang on, is this going to work? If a sandbox crashes and is recreated, there's no guarantee that we see the CNI DEL for the old sandbox before the CNI ADD for the new sandbox... That means we could break existing connections.

PR #181, opened 2020-06-09
Bug 1845674: release-4.4z Cleanup Conntrack when endpoints and pods are deleted

JacobTanenbaum:

When pods and endpoints are deleted if there are ongoing networking connections there will be a stale conntrack entry. This PR ensures that those conntrack entries are cleaned up 

**- How to verify it**

See the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1845674) for reproduction information

Add cleanup of pods and endpoints conntrack entries


JacobTanenbaum:
closing, due to the api update this is no longer acceptable to fix

PR #188, opened 2020-06-11
Fix order of pods, namespaces, services and endpoints

pecameron:
On a restart of ovnkube master due to leader election change
(normal) and perhaps loading and processing all the existing pods.

E0609 18:43:27.194389       1 ovn.go:396] timeout waiting for namespace event

Occured. We process services and endpoints before namespaces, which is
the wrong order because pods (which are first) depend on Namespaces.
So if there are a lot of services & endpoints their initial processing
will block the namespaces which will block and fail pods.

SDN-1037 - Processing services and endpoints incorrectly happens before namespaces.
https://issues.redhat.com/browse/SDN-1037

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/ovn.go
+++ b/go-controller/pkg/ovn/ovn.go
@@ -167,8 +167,8 @@ func (oc *Controller) Run(stopChan chan struct{}) error {
 		return err
 	}
 
-	for _, f := range []func() error{oc.WatchPods, oc.WatchServices, oc.WatchEndpoints,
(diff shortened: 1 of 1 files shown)


pecameron:
@dcbw PTAL 
For SDN-1037

dcbw:
/retest
/lgtm
/approve

@pecameron can you do an upstream PR too?

pecameron:
@dcbw bot says "merged commit ef654d2" but this is not "merged" What is going on?

trozet:
@dcbw why did we merge this when it is not upstream?

PR #195, opened 2020-06-23
Adding a wait for the pod annotation in the hybrid network watcher

JacobTanenbaum:
it seems like the hybrid network controller loop is trying to get pods
ovn annotations before the ovn controller loop finishes setting it

**- How to verify it**

### DIFF
--- a/go-controller/hybrid-overlay/pkg/controller/node.go
+++ b/go-controller/hybrid-overlay/pkg/controller/node.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"time"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/hybrid-overlay/pkg/types"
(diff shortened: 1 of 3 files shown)


dcbw:
@JacobTanenbaum while this will work, I think it's the wrong approach. There's probably an issue with Dave's informer PR and update notifications. I would throw some debugging klogs into podChanged() and NewDefaultEventHandler() to see why the hybrid overlay's AddPod() is not called when the pod's annotation is updated by the master.

We'd expect to see `klog.V(5).Infof("cleaning up hybrid overlay pod %s/%s because %v", pod.Namespace, pod.Name, err)` a couple times and then when the master sets the annotation we we should get another AddPod() call and the annotation will be correct so it should get beyond that part of AddPod().

JacobTanenbaum:
closed in favor of https://github.com/openshift/ovn-kubernetes/pull/199

PR #200, opened 2020-07-10
Upstream merge 2020-07-10

dcbw:


danwinship:
oh, I wasn't paying attention and filed #201. And this rebase is wrong anyway; it drops the Azure/GCP Ingress patches, and doesn't properly dual-stack-ify the host-network anti-MCS patches

PR #201, opened 2020-07-11
rebase to upstream 2020-07-11

danwinship:
Specifically to get all pending dual-stack work

Note in particular the three new local commits:
- [iptables-scripts: sync to upstream](https://github.com/openshift/ovn-kubernetes/commit/9d53b73fc902e44aa1fb00023d7cb2dcc08ef3dd), which just removes some unnecessary divergence
- [Revert IPv4-in-single-stack-IPv6 hack](https://github.com/openshift/ovn-kubernetes/pull/201/commits/dfb20b2be464a9205b51c392fcc901c6753f06a7), which reverts the single-stack Azure/AWS hacks, which never fully worked, and are no longer relevant; in the future, if clouds only support dual-stack, not single-stack IPv6, then we will only support dual-stack OCP on them, not single-stack IPv6.
- [Mark OCP-specific hacks more clearly](https://github.com/openshift/ovn-kubernetes/commit/3f708d6c24680d3475ca9e40fa854dd78a7cba57) which renames the existing OCP-local go files to `OCP_HACKS.go`, moves some more code into them, and adds `// OCP HACK` comments around the pieces of code in other files that refer to them. Keeping the code in other files helps to prevent merge conflicts, and adding the `// OCP HACK` comments help to make it clearer, when rebasing, that a merge conflict is being caused by an intentional divergence.

The previous rebase had some divergence from upstream around the delete conntrack stuff. I assume that this was either accidental or due to the fact that that code was still in flux. This merge puts us fully back in sync with upstream on that.

The MCS and Azure/GCP LoadBalancer hacks in `go-controller/pkg/node/OCP_HACKS.go` have now been updated for dual-stack.

PR #203, opened 2020-07-16
clean up local hacks

danwinship:


danwinship:
actually
/assign dcbw
I was testing this to figure out if the loadbalancer problem was in this part of #201 or the actual rebase, and it looks like it's in the rebase part and this is fine... so this could theoretically just merge now

go-controller/pkg/node/gateway_localnet.go
```diff
@@ -90,8 +90,9 @@ func delIptRules(ipt util.IPTablesHelper, rules []iptRule) {
 func generateGatewayNATRules(ifname string, ip net.IP) []iptRule {
 	// Allow packets to/from the gateway interface in case defaults deny
 	rules := make([]iptRule, 0)
-	// Block MCS
+	// OCP HACK: Block MCS Access. https://github.com/openshift/ovn-kubernetes/pull/170
```
trozet, line 93: what is MCS? I looked in the bug at that PR it still just points the acronym.

go-controller/pkg/node/gateway_localnet.go
```diff
@@ -271,30 +271,9 @@ func localnetIptRules(svc *kapi.Service, gatewayIP string) []iptRule {
...
-				chain: iptableNodePortChain,
-				args: []string{
-					"-d", ing.IP,
-					"-p", string(protocol), "--dport", ingPort,
-					"-j", "ACCEPT",
-				},
-			})
-		}
+		// OCP HACK: Fix Azure/GCP LoadBalancers. https://github.com/openshift/ovn-kubernetes/pull/112
+		rules = append(rules, getLoadBalancerIPTRules(svc, svcPort, gatewayIP, svcPort.NodePort)...)
```
trozet, line 275: we should go back and fix this upstream, is there a github issue already upstream?

go-controller/pkg/cni/helper_linux.go
```diff
@@ -315,16 +313,11 @@ func (pr *PodRequest) ConfigureInterface(namespace string, podName string, ifInf
...
-	err = wait.PollImmediate(100*time.Millisecond, 20*time.Second, func() (bool, error) {
-		stdout, err := ofctlExec("dump-flows", "br-int")
-		if err != nil {
-			return false, nil
-		}
-		return strings.Contains(stdout, ifInfo.IPs[0].IP.String()), nil
-	})
-	if err != nil {
+	// OCP HACK: wait for OVN to fully process the new pod
+	if err = waitForBrIntFlows(ifInfo.IPs[0].IP.String()); err != nil {
```
trozet, line 317: why does this exist? how is this required downstream only?

go-controller/pkg/node/gateway_localnet.go
danwinship: Machine Config Server. I don't know if there's a good primer on this but basically it is a security hole if pods can reach it, and they're supposed to fix it but it's complicated. (Basically, the MCS makes it easy to auto-scale-up a cluster with new nodes, but if you can pretend to be a newly-created node, you can eventually get cluster-admin credentials.)

go-controller/pkg/node/gateway_localnet.go
danwinship: yes, https://github.com/ovn-org/ovn-kubernetes/pull/1126. We haven't implemented it for shared gateway mode yet so it hasn't merged upstream. (Feel free to take over that PR if you understand shared gateway mode well at this point.)

danwinship:
I'm not sure what's going on here... I think something in CI is broken. If I use clusterbot to build an image from this PR, I can launch it with no problems. But the tests are reliably failing because they never reach the point of creating worker nodes, which does not seem to be related to any ovn-kube failures

danwinship:
hm... e2e-gcp-ovn passed that time, and e2e-aws-ovn successfully installed a cluster but flaked on a storage test. So maybe there was something infrastructural that was broken before that's fixed now?

/retest

go-controller/pkg/cni/helper_linux.go
dcbw: @trozet I wasn't sure whether Girish would like it but it's worth proposing upstream.

trozet: after discussing we need to put this upstream.

trozet:
Adding this comment for overall tracking purposes. All of the local go hacks here should go away except the MCS one, which is downstream only.

Upstream:
- https://github.com/ovn-org/ovn-kubernetes/pull/1126
- wait for br-int flows before CNI finishes

PR #219, opened 2020-07-29
hybrid-overlay-node: Windows service enablement

VaishnaviHire:
This PR introduces running hybrid overlay as a service
on Windows. This is achieved by adding a --windows-service
flag to enable running it as a Windows service. This flag
is ignored on Linux. This approach is based on kube-proxy
and kubelet services.

Upstream PR: ovn-org/ovn-kubernetes#1514

Signed-off-by: mansikulkarni96 <mankulka@redhat.com>

Co-authored-by: aravindhp <aravindh@redhat.com>

### DIFF
--- a/go-controller/go.mod
+++ b/go-controller/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/stretchr/testify v1.4.0
 	github.com/urfave/cli/v2 v2.2.0
 	github.com/vishvananda/netlink v0.0.0-20200625175047-bca67dfc8220
+	golang.org/x/sys v0.0.0-20200121082415-34d275377bf9
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
 	gopkg.in/fsnotify/fsnotify.v1 v1.4.7
 	gopkg.in/gcfg.v1 v1.2.3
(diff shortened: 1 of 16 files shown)


sdodson:
@vasukulkarni Needs a rebase.

danwinship:
/hold
why is this being submitted here? It has not yet merged upstream and is still under discussion. When it merges there it will make its way here via a rebase. We do not want to carry large divergences from upstream

aravindhp:
> /hold
> why is this being submitted here? It has not yet merged upstream and is still under discussion. When it merges there it will make its way here via a rebase. We do not want to carry large divergences from upstream

We did this after discussing with @dcbw. @dcbw please weigh in here.

aravindhp:
> > /hold
> > why is this being submitted here? It has not yet merged upstream and is still under discussion. When it merges there it will make its way here via a rebase. We do not want to carry large divergences from upstream
> 
> We did this after discussing with @dcbw. @dcbw please weigh in here.

To further iterate the plan... We need this feature for Windows Containers to GA in the 4.6 timeframe. We are unsure if the upstream PR will merge in time for the code freeze and downstream merge. Moreover the basic functionality in both places is same except that in upstream there is a request for a refactor. So the plan is to carry this patch so that we can land the Windows Containers feature and then land the refactor after the freeze.

I also want to call out that the way the patch stands here we are on par with how `kube-proxy` runs as a Windows service.

sdodson:
@VaishnaviHire Can you please reference the upstream PR so the two are linked?

VaishnaviHire:
> @VaishnaviHire Can you please reference the upstream PR so the two are linked?

Upstream PR : https://github.com/openshift/ovn-kubernetes/pull/219#issuecomment-666549381

dcbw:
I'm OK with this downstream for now given (a) it's Windows-only and (b) it's very isolated.

danwinship:
@VaishnaviHire it looks like `go-controller/hybrid-overlay/cmd/hybrid-overlay-node/hybrid-overlay-node.go` is the only existing go file that this touches. To ensure that the changes needed for this PR do not get accidentally reverted when rebasing from upstream, please wrap each modified section of that file in

    // OCP HACK: Windows service enablement https://github.com/openshift/ovn-kubernetes/pull/219
    ...
    // END OCP HACK

eg, see https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_localnet.go#L224

VaishnaviHire:
Closing this as the upstream PR(ovn-org/ovn-kubernetes#1514) is merged.

PR #224, opened 2020-07-31
[WIP] Fix leader election timers

trozet:
The current timers are backwards for RetryPeriod and RenewDeadline. The
RetryPeriod is the overall time it should wait to try to renew, while
RenewDeadline is the time each renew can take. RetryPeriod should be
much bigger than RenewDeadline, or else we will only try to renew one
time.

### DIFF
--- a/go-controller/pkg/ovn/pods.go
+++ b/go-controller/pkg/ovn/pods.go
@@ -112,7 +112,9 @@ func (oc *Controller) deleteLogicalPort(pod *kapi.Pod) {
 		// delete src-ip cached route to GR
 		nsInfo, err := oc.waitForNamespaceLocked(pod.Namespace)
 		if err != nil {
-			klog.Errorf(err.Error())
+			klog.Errorf("Unable to check port: %s, ip: %s for external gw route deletion: %v", portInfo.name,
+				podIPNet.IP, err)
+			continue
 		}
 		podIP := podIPNet.IP.String()
 		if gwToGr, ok := nsInfo.podExternalRoutes[podIP]; ok {


go-controller/pkg/config/config.go
```diff
@@ -94,8 +94,8 @@ var (
 	// MasterHA holds master HA related config options.
 	MasterHA = MasterHAConfig{
 		ElectionLeaseDuration: 60,
-		ElectionRenewDeadline: 30,
-		ElectionRetryPeriod:   20,
+		ElectionRenewDeadline: 5,
```
dcbw, line 97: @trozet why the lower Renew deadline? Docs say `RenewDeadline is the duration that the acting master will retry refreshing leadership before giving up.`; do we want the active master to give up so quickly?

trozet: from my reading of the leader election code, the renew deadline is the time a single renew iteration attempt will take. The retry period being larger than the renew deadline will allow us to attempt to renew multiple times:
Renew:
`// renew loops calling tryAcquireOrRenew and returns immediately when tryAcquireOrRenew fails or ctx signals done.
func (le *LeaderElector) renew(ctx context.Context) {
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
	wait.Until(func() {
		timeoutCtx, timeoutCancel := context.WithTimeout(ctx, le.config.RenewDeadline)
		defer timeoutCancel()
		err := wait.PollImmediateUntil(le.config.RetryPeriod, func() (bool, error) { ####<---wait maximum retry period
			return le.tryAcquireOrRenew(timeoutCtx), nil  <---try to renew with renew deadline maximum
		}, timeoutCtx.Done())

		le.maybeReportTransition()
		desc := le.config.Lock.Describe()
		if err == nil {
			klog.V(5).Infof("successfully renewed lease %v", desc)
			return
		}
		le.config.Lock.RecordEvent("stopped leading")
		le.metrics.leaderOff(le.config.Name)
		klog.Infof("failed to renew lease %v: %v", desc, err)
		cancel()
	}, le.config.RetryPeriod, ctx.Done())`

The problem we see in GCP job is:
`// tryAcquireOrRenew tries to acquire a leader lease if it is not already acquired,
// else it tries to renew the lease if it has already been acquired. Returns true
// on success else returns false.
func (le *LeaderElector) tryAcquireOrRenew(ctx context.Context) bool {
	now := metav1.Now()
	leaderElectionRecord := rl.LeaderElectionRecord{
		HolderIdentity:       le.config.Lock.Identity(),
		LeaseDurationSeconds: int(le.config.LeaseDuration / time.Second),
		RenewTime:            now,
		AcquireTime:          now,
	}

	// 1. obtain or create the ElectionRecord
	oldLeaderElectionRecord, oldLeaderElectionRawRecord, err := le.config.Lock.Get(ctx)
	if err != nil {
		if !errors.IsNotFound(err) {
			klog.Errorf("error retrieving resource lock %v: %v", le.config.Lock.Describe(), err)
			return false
		}`

So in the log we only try one time to renew:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/737/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn/1289172215102705664/artifacts/e2e-gcp-ovn/pods/openshift-ovn-kubernetes_ovnkube-master-9kx6z_ovnkube-master_previous.log

E0731 12:44:04.674901       1 leaderelection.go:320] error retrieving resource lock openshift-ovn-kubernetes/ovn-kubernetes-master: Get https://api-int.ci-op-h3cqnwy5-b3a20.origin-ci-int-gce.dev.openshift.com:6443/api/v1/namespaces/openshift-ovn-kubernetes/configmaps/ovn-kubernetes-master: context deadline exceeded
I0731 12:44:04.674963       1 leaderelection.go:277] failed to renew lease openshift-ovn-kubernetes/ovn-kubernetes-master: timed out waiting for the condition

I think we are not retrying because the renew deadline is longer than the retry period.

dcbw:
FWIW

```
	// RenewDeadline is the duration that the acting master will retry
	// refreshing leadership before giving up.
	//
	// Core clients default this value to 10 seconds.
	RenewDeadline time.Duration
	// RetryPeriod is the duration the LeaderElector clients should wait
	// between tries of actions.
	//
	// Core clients default this value to 2 seconds.
	RetryPeriod time.Duration
```

abhat:
`wait.Until()` takes the retry period as argument which is basically saying it will retry `f` every retry seconds. So the outer `wait.Until()` looks ok to me. Is the [`cancel()`](https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L279) an issue cuz it's really cancelling the outer `wait.Until`? There is already a `timeOutCancel` for the inner wait.PollImmediateUntil() which is for each renewal.

trozet:
> `wait.Until()` takes the retry period as argument which is basically saying it will retry `f` every retry seconds. So the outer `wait.Until()` looks ok to me. Is the [`cancel()`](https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L279) an issue cuz it's really cancelling the outer `wait.Until`? There is already a `timeOutCancel` for the inner `wait.PollImmediateUntil()` which is for each renewal.

maybe I have it backwards, and we are only trying one time because the channel is getting closed. I'll try to add debug logs so we can see whats happening.

trozet:
> > `wait.Until()` takes the retry period as argument which is basically saying it will retry `f` every retry seconds. So the outer `wait.Until()` looks ok to me. Is the [`cancel()`](https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L279) an issue cuz it's really cancelling the outer `wait.Until`? There is already a `timeOutCancel` for the inner `wait.PollImmediateUntil()` which is for each renewal.
> 
> maybe I have it backwards, and we are only trying one time because the channel is getting closed. I'll try to add debug logs so we can see whats happening.

Actually we are not closing cause of the channel, we are reaching the renewal deadline. We only try one time due to the timers I think. Trying a new patch to lower retry period.

abhat:
> Actually we are not closing cause of the channel, we are reaching the renewal deadline. We only try one time due to the timers I think. Trying a new patch to lower retry period.

How will a lower retry help? You think within the renew deadline, we aren't retrying more than once because the retry timer is too high? So setting it to lower will let it retry several times within the renewal time window?

trozet:
> > Actually we are not closing cause of the channel, we are reaching the renewal deadline. We only try one time due to the timers I think. Trying a new patch to lower retry period.
> 
> How will a lower retry help? You think within the renew deadline, we aren't retrying more than once because the retry timer is too high? So setting it to lower will let it retry several times within the renewal time window?

yeah that was my thought. I see in the latest log, it did try a bunch of times until it finally hit the renewal deadlne, but it didn't print any of the reasons why it could not renew. So I added some more debugs so we can hopefully see why it cant renew.

trozet:
The latest run shows some interesting info:
```
I0731 20:29:53.637490       1 leaderelection.go:263] TROZET running renew wait function
I0731 20:29:53.637532       1 leaderelection.go:316] TROZET attempting to acquire election record and lock
I0731 20:29:53.659622       1 leaderelection.go:364] TROZET DEFAULT return true
I0731 20:29:53.659646       1 leaderelection.go:269] TROZET DONE with trying to renew, err: %v<nil>
I0731 20:29:53.659654       1 leaderelection.go:273] successfully renewed lease openshift-ovn-kubernetes/ovn-kubernetes-master
I0731 20:29:58.659839       1 leaderelection.go:263] TROZET running renew wait function
I0731 20:29:58.659875       1 leaderelection.go:316] TROZET attempting to acquire election record and lock
E0731 20:30:28.660137       1 leaderelection.go:321] error retrieving resource lock openshift-ovn-kubernetes/ovn-kubernetes-master: Get https://api-int.ci-op-f54vs199-99b10.origin-ci-int-gce.dev.openshift.com:6443/api/v1/namespaces/openshift-ovn-kubernetes/configmaps/ovn-kubernetes-master: context deadline exceeded
I0731 20:30:28.660166       1 leaderelection.go:269] TROZET DONE with trying to renew, err: %vtimed out waiting for the condition
I0731 20:30:28.660178       1 leaderelection.go:278] failed to renew lease openshift-ovn-kubernetes/ovn-kubernetes-master: timed out waiting for the condition
```

What I see from the above is that we did renew the leader election at 20:29:53, and then 5 seconds later, we tried to renew again at I0731 20:29:58. I would have thought the renewal of election would hold for the leader election timeout which is 60 seconds. I think it is too much of a coincidence that exactly 5 seconds this happened and the retry timer is 5 seconds. So that is one thing we need to figure out. 

However, the other key piece of information from this log is that we tried to lock the resource at 20:29:58. That issues a HTTP request to a config map in the API server. However we can see that we did not finish that request until 20:30:28. That indicates to me that there might be a long HTTP timeout happening during some network problem to reach the API server. However, that HTTP timeout is chewing threw all of the retry time we have until the renewal deadline happens. I'm trying to add an HTTP timeout of 4 seconds to the next patch and see if it has any effect to test this theory.

abhat:
/test e2e-gcp-ovn

Starting the gcp ovn job with the new client set changes.

PR #226, opened 2020-08-01
Bug 1859883: Fix ovnkube-node aggressive memory allocation to prevent …

vishnoianil:
…memory bloating

    During the pod creation, ovnkube-node dump all the flows from br-int and grep
    for the ip address of the pod to verify that pod related flows are programmed.
    This can cause aggressive and big memory allocation to store the string output
    from ovs-ofctl command.
    This patch optimized the ovs-ofctl dump-flow query to dump flows only related
    to the pods. This will avoid dumping the massive string output and prevent
    memory bloating.

Signed-off-by: Anil Vishnoi <avishnoi@redhat.com>

### DIFF
--- a/go-controller/pkg/cni/OCP_HACKS.go
+++ b/go-controller/pkg/cni/OCP_HACKS.go
@@ -86,13 +86,13 @@ func ofctlExec(args ...string) (string, error) {
 	klog.V(5).Infof("exec: ovs-ofctl %s", cmdStr)
 
 	err := cmd.Run()
-	stdoutStr := string(stdout.Bytes())
-	stderrStr := string(stderr.Bytes())
 	if err != nil {
-		klog.V(5).Infof("exec: stderr: %q", stderrStr)
-		return "", fmt.Errorf("failed to run 'ovs-ofctl %s': %v\n  %q", cmdStr, err, string(stderr.Bytes()))
+		stderrStr := string(stderr.Bytes())
+		klog.Errorf("exec: ovs-ofctl %s : stderr: %q",cmdStr, stderrStr)
+		return "", fmt.Errorf("failed to run 'ovs-ofctl %s': %v\n  %q", cmdStr, err, stderrStr)
 	}
-	klog.V(5).Infof("exec: stdout: %q", stdoutStr)
+	stdoutStr := string(stdout.Bytes())
+	klog.V(5).Infof("exec: ovs-ofctl %s: stdout: %q", cmdStr, stdoutStr)
 
 	trimmed := strings.TrimSpace(stdoutStr)
 	// If output is a single line, strip the trailing newline
@@ -104,9 +104,9 @@ func ofctlExec(args ...string) (string, error) {
 
 func waitForPodFlows(mac string) error {
 	return wait.PollImmediate(200*time.Millisecond, 20*time.Second, func() (bool, error) {
-		//Query the flows by mac address and count the number of flows.
+		// Query the flows by mac address
 		query := fmt.Sprintf("table=9,dl_src=%s",mac)
-		//ovs-ofctl dumps error on stderr, so stdout will only dump flow data if matches the query.
+		// ovs-ofctl dumps error on stderr, so stdout will only dump flow data if matches the query.
 		stdout, err := ofctlExec("dump-flows", "br-int", query)
 		if err != nil {
 			return false, nil
--- a/go-controller/pkg/cni/OCP_HACKS.go
+++ b/go-controller/pkg/cni/OCP_HACKS.go
@@ -3,10 +3,13 @@
 package cni
 
 import (
+	"bytes"
 	"fmt"
 	"os/exec"
 	"strings"
 	"time"
+
+	"k8s.io/klog"
 
 	"github.com/containernetworking/plugins/pkg/ns"
 
@@ -73,29 +76,45 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
 
 // OCP HACK: wait for OVN to fully process the new pod
 func ofctlExec(args ...string) (string, error) {
-	args = append([]string{"--timeout=30"}, args...)
-	output, err := runner.Command("ovs-ofctl", args...).CombinedOutput()
-	if err != nil {
-		return "", fmt.Errorf("failed to run 'ovs-ofctl %s': %v\n  %q", strings.Join(args, " "), err, string(output))
-	}
+	args = append([]string{"--timeout=30", "--no-stats", "--strict"}, args...)
+	var stdout, stderr bytes.Buffer
+	cmd := runner.Command("ovs-ofctl", args...)
+	cmd.SetStdout(&stdout)
+	cmd.SetStderr(&stderr)
 
-	outStr := string(output)
-	trimmed := strings.TrimSpace(outStr)
+	cmdStr := strings.Join(args, " ")
+	klog.V(5).Infof("exec: ovs-ofctl %s", cmdStr)
+
+	err := cmd.Run()
+	stdoutStr := string(stdout.Bytes())
+	stderrStr := string(stderr.Bytes())
+	if err != nil {
+		klog.V(5).Infof("exec: stderr: %q", stderrStr)
+		return "", fmt.Errorf("failed to run 'ovs-ofctl %s': %v\n  %q", cmdStr, err, string(stderr.Bytes()))
+	}
+	klog.V(5).Infof("exec: stdout: %q", stdoutStr)
+
+	trimmed := strings.TrimSpace(stdoutStr)
 	// If output is a single line, strip the trailing newline
 	if strings.Count(trimmed, "\n") == 0 {
-		outStr = trimmed
+		stdoutStr = trimmed
 	}
-
-	return outStr, nil
+	return stdoutStr, nil
 }
 
-func waitForBrIntFlows(ip string) error {
-	return wait.PollImmediate(100*time.Millisecond, 20*time.Second, func() (bool, error) {
-		stdout, err := ofctlExec("dump-flows", "br-int")
+func waitForPodFlows(mac string) error {
+	return wait.PollImmediate(200*time.Millisecond, 20*time.Second, func() (bool, error) {
+		//Query the flows by mac address and count the number of flows.
+		query := fmt.Sprintf("table=9,dl_src=%s",mac)
+		//ovs-ofctl dumps error on stderr, so stdout will only dump flow data if matches the query.
+		stdout, err := ofctlExec("dump-flows", "br-int", query)
 		if err != nil {
 			return false, nil
 		}
-		return strings.Contains(stdout, ip), nil
+		if len(stdout) == 0 {
+			return false, nil
+		}
+		return true, nil
 	})
 }
 
--- a/go-controller/pkg/cni/helper_linux.go
+++ b/go-controller/pkg/cni/helper_linux.go
@@ -314,7 +314,7 @@ func (pr *PodRequest) ConfigureInterface(namespace string, podName string, ifInf
 	}
 
 	// OCP HACK: wait for OVN to fully process the new pod
-	if err = waitForBrIntFlows(ifInfo.IPs[0].IP.String()); err != nil {
+	if err = waitForPodFlows(ifInfo.MAC.String()); err != nil {
 		return nil, fmt.Errorf("timed out dumping br-int flow entries for sandbox: %v", err)
 	}
 	// END OCP HACK


vishnoianil:
@dcbw PTAL, Thanks
@trozet PTAL, i just did bit more optimization to ensure that it works for ipv4/ipv6 and tried to reduce some additional string creation. We can keep any of the PR (upstream or downstream), whatever works better and create less conflicts.

trozet:
We should merge this PR first as a fix that we can backport to 4.4. Then the upstream PR should come after.

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -73,29 +76,45 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
 
 // OCP HACK: wait for OVN to fully process the new pod
 func ofctlExec(args ...string) (string, error) {
-	args = append([]string{"--timeout=30"}, args...)
-	output, err := runner.Command("ovs-ofctl", args...).CombinedOutput()
+	args = append([]string{"--timeout=30", "--no-stats", "--strict"}, args...)
+	var stdout, stderr bytes.Buffer
+	cmd := runner.Command("ovs-ofctl", args...)
+	cmd.SetStdout(&stdout)
+	cmd.SetStderr(&stderr)
```
danwinship, line 83: you don't need all this; just use `Output` instead of `CombinedOutput` above and then you'll only get stdout rather than stdout+stderr

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -73,29 +76,45 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
...
-	return outStr, nil
+	return stdoutStr, nil
 }
 
-func waitForBrIntFlows(ip string) error {
-	return wait.PollImmediate(100*time.Millisecond, 20*time.Second, func() (bool, error) {
-		stdout, err := ofctlExec("dump-flows", "br-int")
+func waitForPodFlows(mac string) error {
+	return wait.PollImmediate(200*time.Millisecond, 20*time.Second, func() (bool, error) {
+		//Query the flows by mac address and count the number of flows.
```
danwinship, line 107: Comments should always have a space after the `//`. Also, this comment is incorrect; you don't count the number of flows.

go-controller/pkg/cni/OCP_HACKS.go
vishnoianil: @danwinship we are capturing the stderr so that we can report the error in case of failure, whether it's timeout or error in query. And the reason i separated them out is to avoid a grep (String operation which is not needed) on CombinedOut.

go-controller/pkg/cni/OCP_HACKS.go
vishnoianil: Thanks for catching this, i added it when i was trying to add some logic of counting the flows. Will fix it.

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -73,29 +76,45 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
...
+
+	cmdStr := strings.Join(args, " ")
+	klog.V(5).Infof("exec: ovs-ofctl %s", cmdStr)
+
+	err := cmd.Run()
+	stdoutStr := string(stdout.Bytes())
+	stderrStr := string(stderr.Bytes())
 	if err != nil {
-		return "", fmt.Errorf("failed to run 'ovs-ofctl %s': %v\n  %q", strings.Join(args, " "), err, string(output))
+		klog.V(5).Infof("exec: stderr: %q", stderrStr)
```
dcbw, line 92: @vishnoianil can we say *which* command failed here? eg ovs-ofctl

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -73,29 +76,45 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
...
+		return "", fmt.Errorf("failed to run 'ovs-ofctl %s': %v\n  %q", cmdStr, err, string(stderr.Bytes()))
 	}
+	klog.V(5).Infof("exec: stdout: %q", stdoutStr)
```
dcbw, line 95: Same here... at least let's say it was ofctl that got called.

aojea:
we can avoid allocating the strings just using a pipe and processing the output (code is an example)

```go
// OCP HACK: wait for OVN to fully process the new pod
func findIPonFlows(bridge, ip string) (bool, error) {
	cmd := runner.Command("ovs-ofctl", "dump-flows", bridge, "--timeout=30")
	output, err := cmd.StdoutPipe()
	if err != nil {
		return false, fmt.Errorf("failed to run 'ovs-ofctl dump-flows %s': %v", bridge, err)
	}
	if err = cmd.Start(); err != nil {
		return false, fmt.Errorf("failed to run 'ovs-ofctl dump-flows %s': %v", bridge, err)
	}

	scanner := bufio.NewScanner(output)
	found := false
	go func() {
		for scanner.Scan() {
			if strings.Contains(scanner.Text(), ip) {
				found = true
			}
		}
	}()

	if err := cmd.Wait(); err != nil {
		return false, fmt.Errorf("failed to run 'ovs-ofctl dump-flows %s': %v", bridge, err)
	}

	return found, nil

}

func waitForBrIntFlows(ip string) error {
	return wait.PollImmediate(100*time.Millisecond, 20*time.Second, func() (bool, error) {
		found, err := findIPonFlows("br-int", ip)
		if err != nil {
			return false, nil
		}
		return found, nil
	})
}

// END OCP HACK
```

dcbw:
@aojea @vishnoianil yeah, good point about scanner. Though I think doing less work on ofctl is probably better since even with the scanner we might be asking ofctl to dump tens or hundreds of thousands of flows...

go-controller/pkg/cni/OCP_HACKS.go
vishnoianil: done

go-controller/pkg/cni/OCP_HACKS.go
vishnoianil: done

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -73,29 +76,45 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
...
+	cmd := runner.Command("ovs-ofctl", args...)
+	cmd.SetStdout(&stdout)
+	cmd.SetStderr(&stderr)
+
+	cmdStr := strings.Join(args, " ")
+	klog.V(5).Infof("exec: ovs-ofctl %s", cmdStr)
+
+	err := cmd.Run()
+	stdoutStr := string(stdout.Bytes())
+	stderrStr := string(stderr.Bytes())
```
squeed, line 90: If you want to save another allocation, you could move this in to the `if err != nil` block.

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -73,29 +76,45 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
...
 	if err != nil {
-		return "", fmt.Errorf("failed to run 'ovs-ofctl %s': %v\n  %q", strings.Join(args, " "), err, string(output))
+		klog.V(5).Infof("exec (ovs-ofctl): stderr: %q", stderrStr)
+		return "", fmt.Errorf("failed to run 'ovs-ofctl %s': %v\n  %q", cmdStr, err, string(stderr.Bytes()))
```
squeed, line 93: the last parameter to `fmt.Errorf()` can be `stderrStr`

dcbw:
But... do we even care about optimizing the string operations down so much, if we're just making ofctl do the filtering for us with "table=9"? In that case we'll only ever get a couple lines of output from the binary so the string operation won't be expensive at all.

Like, literally, what if the patch was just:

```
- stdout, err := ofctlExec("dump-flows", "br-int")
+ stdout, err := ofctlExec("dump-flows", "br-int", "table=9")
```

go-controller/pkg/cni/OCP_HACKS.go
vishnoianil: fixed as well.

vishnoianil:
> But... do we even care about optimizing the string operations down so much, if we're just making ofctl do the filtering for us with "table=9"? In that case we'll only ever get a couple lines of output from the binary so the string operation won't be expensive at all.
> 
> Like, literally, what if the patch was just:
> 
> ```
> - stdout, err := ofctlExec("dump-flows", "br-int")
> + stdout, err := ofctlExec("dump-flows", "br-int", "table=9")
> ```
In my opinion, if the compute/memory cost is linearly increasing with number of pods, save as far as one is not impacting the others (cpu impacting memory or vice-versa). Just to understand the concern, is this optimization looking too complicated ? Just querying on table=9, will dump all the flows from table=9, which can increase based on number of pods, and than grep-ing on ip on that output in ovnkube-node is another costly string operation compared to filtering using ofctl datapath level.

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -73,29 +76,45 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
...
-	args = append([]string{"--timeout=30"}, args...)
-	output, err := runner.Command("ovs-ofctl", args...).CombinedOutput()
+	args = append([]string{"--timeout=30", "--no-stats", "--strict"}, args...)
+	var stdout, stderr bytes.Buffer
+	cmd := runner.Command("ovs-ofctl", args...)
+	cmd.SetStdout(&stdout)
+	cmd.SetStderr(&stderr)
+
+	cmdStr := strings.Join(args, " ")
+	klog.V(5).Infof("exec: ovs-ofctl %s", cmdStr)
```
dcbw, line 86: @vishnoianil "exec (ovs-ofctl):" like the others?

vishnoianil: this dumps the rest of the command string as well, so just kept in that format (easy to grep over logs etc)

dcbw:
The CI failures concern me a bit, doesn't seem to be a pattern with them. Same kind of storage-related failures happen on the 4.5 version. The code in question was added because it fixed a lot of flakes that didn't immediately seem to be networking, so we should be double-sure this is doing what we think it is.  @vishnoianil can you take a look and make sure the CNI bits are definitely waiting correctly?

danwinship:
> Just to understand the concern, is this optimization looking too complicated ?

Yes. We don't need to have the most optimal possible solution. We just need a solution that's good, as opposed to the current code which is terrible.

vishnoianil:
> > Just to understand the concern, is this optimization looking too complicated ?
> 
> Yes. We don't need to have the most optimal possible solution. We just need a solution that's good, as opposed to the current code which is terrible.

@danwinship can you please share your thoughts on which part of the patch is too complicated to understand? Also it would be helpful if you can point out which part of the patch is terrible, i can probably try to fix it.

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -73,29 +76,45 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
...
+
+	err := cmd.Run()
 	if err != nil {
-		return "", fmt.Errorf("failed to run 'ovs-ofctl %s': %v\n  %q", strings.Join(args, " "), err, string(output))
+		stderrStr := string(stderr.Bytes())
+		klog.V(5).Infof("exec (ovs-ofctl): stderr: %q", stderrStr)
```
abhat, line 91: well, we should be at least consistent?

vishnoianil: done

dcbw:
AWS failures seem like a whole lot of apiservers going down at 22:04:30-ish. That causes a lot of:

testcase:
```
Aug  5 22:17:55.955: FAIL: rpc error: code = Unavailable desc = transport is closing
```

apiserver:
```
I0805 22:15:15.573965      18 controlbuf.go:508] transport: loopyWriter.run returning. connection error: desc = "transport is closing"
I0805 22:15:15.574211      18 controlbuf.go:508] transport: loopyWriter.run returning. connection error: desc = "transport is closing"
I0805 22:15:15.574305      18 controlbuf.go:508] transport: loopyWriter.run returning. connection error: desc = "transport is closing"
I0805 22:15:15.576440      18 controlbuf.go:508] transport: loopyWriter.run returning. connection error: desc = "transport is closing"
```

on the part of the apiserver, which is apparently a gRPC server issue, but it's very unclear what the apiserver is suppose

dcbw:
So the test failures appear to be part ovnkube itself, part CI issues.  See https://github.com/ovn-org/ovn-kubernetes/issues/1588 for one instance. Those all appear unrelated to this PR, and we need this in 4.5 before 4.6 is going to get stable, so lgtm.

danwinship:
> Also it would be helpful if you can point out which part of the patch is terrible, i can probably try to fix it.

I meant the version _before_ your patch is terrible

danwinship:
(because it's using too much memory etc)

PR #228, opened 2020-08-02
Fix segfault in exgw route deletion

trozet:
We were missing a continue when we couldn't lock the namespace which
ended up segfaulting during port deletion..


trozet:
@dcbw @danwinship FYI

knobunc:
Overriding the gcp test since all of the failures are unrelated and seem to be due to the platform.

PR #238, opened 2020-08-22
Bug 1870867: Fix br-local duplicate MAC

trozet:
Using br-local with NORMAL action wasn't working because both ports have the same MACUsing.
    This PR sets the annotation so that the OVN GR will use a different MAC address


crawford:
/hold

I'm seeing a lot of very similar looking failures.

dcbw:
CI is passing; we don't yet expect gcp-ovn-upgrade to pass.
/hold cancel

dcbw:
We do not yet expect this job to pass.
/override ci/prow/e2e-gcp-ovn-upgrade

PR #261, opened 2020-09-09
Bug 1878295: Dockerfile: fix ART image builds by making RPM pinning agnostic

dcbw:
ART is still using a rhel7 base content set, while CI uses rhel8.
Which means we have to keep the RPM pinning agnostic.

@danwinship


danwinship:
ah, sneaky
/lgtm

PR #268, opened 2020-09-14
Bug 1868501: Do not default HO cluster subnets

trozet:
The hybrid-overlay-cluster-subnets argument is not required to have a
value for hybrid overlay to work. It is only required when using Windows
nodes that are part of the cluster. It is not required when sending to
3rd party gateways. This patch removes the default subnet value for the
arg.

(cherry picked from commit d117020089abafbc0ffa7b32a27e776efd18e963)

### DIFF
--- a/go-controller/hybrid-overlay/pkg/controller/node_linux_test.go
+++ b/go-controller/hybrid-overlay/pkg/controller/node_linux_test.go
@@ -317,7 +317,7 @@ var _ = Describe("Hybrid Overlay Node Linux Operations", func() {
 				// Assume fresh OVS bridge
 				Output: "",
 			})
-
+			config.HybridOverlay.RawClusterSubnets = "10.0.0.1/16/23"
 			_, err := config.InitConfig(ctx, fexec, nil)
 			Expect(err).NotTo(HaveOccurred())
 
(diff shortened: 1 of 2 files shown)


trozet:
@JacobTanenbaum @dcbw PTAL

JacobTanenbaum:
We need to make sure that the windows team is ready for this change, AFAIK the windows team tests and there documentation mention the default value

danwinship:
this was already fixed on 4.6 though; this is just a cherry-pick

markmc:
Seems bad to have an invalid route defined. Relatively minor fix for an important feature.

trozet:
@danwinship can you please override gcp upgrade?

trozet:
@danwinship why is it not showing as overridden?

danwinship:
It's not always reliably to `/override` a test while it's running. I don't know if that's a bug or if it's some oddity of the exact semantics of the operation. Sometimes it's helpful to `/hold` when you `/override`, so that if the override doesn't stick, you can re-override before the bot `/retest`s.

Anyway, it looks like it's going to merge now

alexanderConstantinescu:
> @danwinship can you please override gcp upgrade?

@trozet / @danwinship : why are we overriding the upgrade job? Is it a known problem on 4.5 / is there a bug opened for it?

PR #267, opened 2020-09-14
[WIP] dummy commit to create a PR that will run the tests against a 'clean' master

JacobTanenbaum:
**- What this PR does and why is it needed**

**- Special notes for reviewers**

**- How to verify it**

**- Description for the changelog**

PR #269, opened 2020-09-15
Bug 1872470: Upstream merge 9-14-2020

trozet:
Includes update to use local gateway mode.


dcbw:
Yeah, we have to figure this one out, but it was broken before this PR too, see https://github.com/openshift/ovn-kubernetes/pull/267
/override ci/prow/e2e-operator-with-custom-vxlan-port

PR #281, opened 2020-09-22
Bug 1880591: Allow local no bridge

trozet:
For OCP we upgrade CNO before MCO. This means that ovn-k8s will get upgraded first when 4.5->4.6 upgrade happens and will try to create a bridge for local gateway mode now. We have to have an intermediate step where we do not create a shared gateway bridge until after the node reboots with ovs-configuration. Therefore this downstream only patch allows us to launch local gateway mode with no shared gateway bridge, and it should still function during upgrade.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -522,7 +522,13 @@ func (n *OvnNode) initSharedGateway(subnets []*net.IPNet, gwNextHops []net.IP, g
 	var brCreated bool
 	var err error
 
-	if bridgeName, _, err = util.RunOVSVsctl("--", "port-to-br", gwIntf); err == nil {
+	// OCP HACK
+	// Do not configure OVS bridge for local gateway mode with a gateway iface of none
+	// For 4.5->4.6 migration, see 	https://github.com/openshift/ovn-kubernetes/pull/281
+	if gwIntf == "none" {
+		return n.initSharedGatewayNoBridge(subnets, gwNextHops, nodeAnnotator)
+		// END OCP HACK
+	} else if bridgeName, _, err = util.RunOVSVsctl("--", "port-to-br", gwIntf); err == nil {
 		// This is an OVS bridge's internal port
 		uplinkName, err = util.GetNicName(bridgeName)
 		if err != nil {
--- a/go-controller/pkg/ovn/endpoints.go
+++ b/go-controller/pkg/ovn/endpoints.go
@@ -114,9 +114,14 @@ func (ovn *Controller) AddEndpoints(ep *kapi.Endpoints) error {
 func (ovn *Controller) handleNodePortLB(node *kapi.Node) error {
 	gatewayRouter := gwRouterPrefix + node.Name
 	var physicalIPs []string
-	if physicalIPs, _ = ovn.getGatewayPhysicalIPs(gatewayRouter); physicalIPs == nil {
-		return fmt.Errorf("gateway physical IP for node %q does not yet exist", node.Name)
+	// OCP HACK - there will not be a GR during local gw + no gw interface mode (upgrade from 4.5->4.6)
+	// See https://github.com/openshift/ovn-kubernetes/pull/281
+	if !isGatewayInterfaceNone() {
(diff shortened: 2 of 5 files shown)


trozet:
@danwinship @pliurh @dcbw this will allow us to launch local gw mode with no shared gw bridge for upgrade and openshift-sdn-> ovn-k8s conversion. I have a corresponding PR for CNO:
https://github.com/openshift/cluster-network-operator/pull/801

and another PR that forces gateway-interface to be none just to ensure functionality still works in this mode:
https://github.com/openshift/ovn-kubernetes/pull/282

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -586,7 +620,7 @@ func (n *OvnNode) initSharedGateway(subnets []*net.IPNet, gwNextHops []net.IP, g
 
 	return func() error {
 		if config.Gateway.NodeportEnable {
-			if config.Gateway.Mode == config.GatewayModeLocal {
+			if config.Gateway.Mode == config.GatewayModeLocal && gwIntf != "None" {
```
trozet, line 623: this isn't right

trozet:
cluster-botAPP  5:26 PM
job test upgrade 4.5.8 openshift/ovn-kubernetes#281,openshift/cluster-network-operator#801 aws,ovn succeeded

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -522,7 +522,41 @@ func (n *OvnNode) initSharedGateway(subnets []*net.IPNet, gwNextHops []net.IP, g
 	var brCreated bool
 	var err error
 
-	if bridgeName, _, err = util.RunOVSVsctl("--", "port-to-br", gwIntf); err == nil {
+	// OCP HACK
```
danwinship, line 525: include a very brief description and a link to this PR please (in each added block of code)

go-controller/pkg/ovn/endpoints.go
```diff
@@ -3,6 +3,7 @@ package ovn
 import (
 	"fmt"
 
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
```
danwinship, line 6: Adding imports to existing files tends to cause rebase conflicts later. (And adding new functions to the end of the file does as well.) This is the big reason I tried to put most of the other local hacks into separate `OCP_HACKS.go` files, and just added single function calls as needed to the existing files.

go-controller/pkg/ovn/master.go
```diff
@@ -449,9 +449,20 @@ func (oc *Controller) syncGatewayLogicalNetwork(node *kapi.Node, l3GatewayConfig
...
-	err = gatewayInit(node.Name, clusterSubnets, hostSubnets, joinSubnets, l3GatewayConfig, oc.SCTPSupport)
-	if err != nil {
-		return fmt.Errorf("failed to init shared interface gateway: %v", err)
+	// OCP HACK
+	// GatewayModeLocal is only used if Local mode is specified and None shared gateway bridge is specified
+	// This is to allow local gateway mode without having to configure/use the shared gateway bridge
+	if l3GatewayConfig.Mode == config.GatewayModeLocal {
+		err = gatewayInitMinimal(node.Name, l3GatewayConfig, oc.SCTPSupport)
+		if err != nil {
+			return fmt.Errorf("failed to init shared interface gateway: %v", err)
```
danwinship, line 458: "local interface gateway" ?

trozet: yeah will fix

go-controller/pkg/node/gateway_shared_intf.go
trozet: will do

go-controller/pkg/ovn/endpoints.go
trozet: so do you prefer me moving this to another file? Just this occurrence or all places?

danwinship: yeah, I'd suggest for each directory you're adding patches to, put as much of the new code as possible into a new file, and just make as few changes as possible to the existing files, to minimize the chance that your modifications will create merge conflicts with future rebases

PR #296, opened 2020-09-28
[WIP] Don't exit db checker if the cluster isn't established

abhat:


abhat:
After openshift/cluster-network-operator#806 was fixed, this isn't needed. Closing the PR.

PR #302, opened 2020-10-05
[DNM]: Consider changed configuration when kicking out raft members

abhat:
In certain use-cases such as master replacement, we have seen
that the configuration of the db/master pods that holds the addresses
of the raft members isn't updated by the time the new master joins the
raft cluster. When this happens the db pods carrying the old address
configuration for nbdb and sbdb would kick the newly joined master out.
This commit addresses this issue by additionally checking existing master
pods' IPs against the server id list for determining whether a given
server id should be booted out of the raft cluster.

**- How to verify it**

### DIFF
--- a/dist/templates/ovnkube-db.yaml.j2
+++ b/dist/templates/ovnkube-db.yaml.j2
@@ -48,6 +48,7 @@ spec:
   template:
     metadata:
       labels:
+        app: ovnkube-db
         name: ovnkube-db
         component: network
         type: infra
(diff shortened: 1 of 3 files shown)


abhat:
/hold 
need to use this PR to dev-test in BM environment.

squeed:
Wow, this code needs some tests (sorry that this PR is what made me said this)

I suggest a function, `computeAction( ...state ... ) []string{} ` that returns the list of members to kick. Then you can write some simple unit tests.

PR #304, opened 2020-10-07
Bug 1885517: Clean up old local gw mode external ports

trozet:
When we upgrade from 4.5->4.6 we transition from the old local gateway
mode to the new. In this case, OVS is clean when 4.6 comes up, but OVN
DB still has a leftover "br-local" port that exists on the external
switch. ovn-controller will try to wire this port to br-int, creating
double patch ports between br-ex and br-int, thus causing a broadcast
storm.

This patch ensures we remove the old port from OVN NB DB, which
ovn-controller will then ensure is removed from the external switch.

### DIFF
--- a/go-controller/pkg/ovn/OCP_HACKS.go
+++ b/go-controller/pkg/ovn/OCP_HACKS.go
@@ -5,6 +5,7 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 	kapi "k8s.io/api/core/v1"
+	"k8s.io/klog"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


trozet:
/hold
testing upgrade from 4.5 to 4.6 to make sure this works correctly

go-controller/pkg/ovn/OCP_HACKS.go
```diff
@@ -95,3 +96,26 @@ func gatewayInitMinimal(nodeName string, l3GatewayConfig *util.L3GatewayConfig,
...
+		return
+	}
+	if len(stdout) == 0 {
+		return
+	}
+	klog.Infof("Found old local gateway interface to remove. Name: %s, ID: %s", ifaceID, stdout)
+	_, stderr, err = util.RunOVNNbctl("lsp-del", ifaceID)
+	if err != nil {
+		klog.Errorf("Unable to delete old local gateway interface: %s, stderr: %s, error: %v", ifaceID, stderr,
+			err)
```
danwinship, line 117: Should this be considered a bigger problem than just klog.Errorf? We expect there _will_ be problems with the cluster if this happens, right?

danwinship:
/lgtm
the lsp-del really won't plausibly fail I guess

PR #305, opened 2020-10-08
Bug 1880591: Fix service sync reject ACL name matching

trozet:
Fix service sync reject ACL name matching
    
    The generateACLName function was including backslash escapes for the ACL
    name for use in OVN commands that have trouble with literal ":". That
    was causing a mismatch when services were syncing because the name
    actually returned from an OVN command does not include any backslashes
    so the names would not match.
    

knobunc:
/hold

Pending upstream merge.  @trozet feel free to pull this hold when you are comfortable

PR #306, opened 2020-10-08
10-8-2020 merge

trozet:


dcbw:
All the failing tests are due to prometheus alerts about ovn components not running, likely due to https://github.com/openshift/cluster-network-operator/pull/834

PR #308, opened 2020-10-14
Bug 1887585: Avoid restarting ovn-dbchecker if the local db check-cluster fails

abhat:
Do not treat check-cluster error as fatal.


abhat:
Closing in favor of: ovn-org/ovn-kubernetes/pull/1763

PR #315, opened 2020-10-20
Klog v2

tfherbert:
fixes #1886062

k8s.io/klog moved to v2 in client-go and frieds. The operator does not use v2 yet and hence, we lose one half of the logging output (v2).

Logging is broken due to mix of k8s.io/klog v1 and v3


go-controller/vendor/k8s.io/apimachinery/pkg/api/meta/meta.go
```diff
@@ -25,7 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
```
juanluisvaladas: You can't do this.
This will be overwritten by go mod vendor (and is replaced in the following commit go mod vendor)

I think in go.mod you can do instead of `replace github.com/ebay/go-ovn v0.1.0 => github.com/ebay/go-ovn v0.1.1-0.20200810162212-30abed5fb968` you can do:

```
replace (
	github.com/ebay/go-ovn v0.1.0 => github.com/ebay/go-ovn v0.1.1-0.20200810162212-30abed5fb968
	k8s.io/klog => k8s.io/klog/v2 v2.3.0
)
```

And that would force klog v2. I  haven't actually tested this.

juanluisvaladas (changes requested):
I see 29 commits unrelated to this change, is that a deliberate rebase? If so it should probably be documented in the PR description.

go-controller/vendor/k8s.io/apimachinery/pkg/api/meta/meta.go
tfherbert: No, my fork and picked up the "extra" commits which should now be in master. The plan is for this PR to be merged into master and then later cherry-picked into 4.6.z.  Maybe I should rebase and generate new clean PR.

As for /v2 that seemed to work and is consistent with other projects merged into openshift. I had to run go mod vendor to suck in klog/v2 into vendor and it seemed to work but it seems messy to me.

tfherbert:
Converting this PR to draft for documentation.

tfherbert:
This is superceded by PR #1817 to ovn.org/ovn-kubernetes already merged

PR #329, opened 2020-10-27
Fix gateway tests for OCP hacks for cleaning up ports

abhat:
Fix unit-tests for cleaning up ports in local gateway mode


dcbw:
Is this patch appropriate for upstream? Or is this to handle OCP-specific hacks?

If the second, then we should mark it as such with OCP HACK or whatever in the Go code.

abhat:
This is only for OCP HACK on release-4.6 since we don't have this in upstream or in master branch downstream. @trozet does it make sense for the original patch to be upstream?

abhat:
@dcbw added comments around the changes to indicate that they are specific to OCP HACK that is only carried downstream in release-4.6
@trozet should the related patch be upstream at all?

abhat:
Close in favor of #347.

PR #332, opened 2020-10-28
WIP config: fix dual-stack config parsing

danwinship:
If you passed dual-stack cluster subnets with the IPv6 subnet first it would mistakenly apply the IPv6 validation rules to the IPv4 subnet too (which was guaranteed to fail).

(creating a PR for cluster-bot)


danwinship:
/close
this will get pulled in by a rebase

PR #333, opened 2020-10-29
[wip] test fix for GR router SNAT in host zone

dcbw:
rhbz#1892311


dcbw:
Still confirming from OpenFlow that the RPM is doing what we need it to do...

trozet:
@dcbw can you pull in these fixes to enable shared gw mode?

https://github.com/openshift/ovn-kubernetes/pull/325/commits

dcbw:
Closed in favor of Tim's

PR #344, opened 2020-11-13
Dockerfile: Add required packages to enable IPsec

markdgray:
Signed-off-by: Mark Gray <mark.d.gray@redhat.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,9 +40,10 @@ RUN INSTALL_PKGS=" \
 	libpcap iproute strace \
 	containernetworking-plugins \
 	tcpdump iputils \
+	libreswan \
 	" && \
 	yum install -y --setopt=tsflags=nodocs --setopt=skip_missing_names_on_install=False $INSTALL_PKGS && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -40,6 +40,7 @@ RUN INSTALL_PKGS=" \
 	libpcap iproute strace \
 	containernetworking-plugins \
 	tcpdump iputils \
+	openvswitch-ipsec libreswan \
 	" && \
 	yum install -y --setopt=tsflags=nodocs --setopt=skip_missing_names_on_install=False $INSTALL_PKGS && \
 	yum install -y --setopt=tsflags=nodocs --setopt=skip_missing_names_on_install=False "openvswitch2.13 == $ovsver" "openvswitch2.13-devel == $ovsver" "python3-openvswitch2.13 == $ovsver" && \
```
dcbw, line 46: @markdgray actually I think you need to add "openvswitch2.13-ipsec == $ovsver" to this line, and remove the "openvswitch-ipsec" from line 43 above. That should make things work. -72 is available in CI now.

markdgray: Thanks @dcbw, I was assuming the package name would be the same in RHEL and, of course, it is not. I will try this morning.

dcbw:
/test e2e-ovn-hybrid-step-registry
msg=Error: Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached.

PR #346, opened 2020-11-16
[WIP] DNM Changes local gateway flows to NORMAL action

trozet:
Following the changes for shared gw mode from:
https://github.com/ovn-org/ovn-kubernetes/pull/1774

This behavior will allow NORMAL action for all packets not destined to
the shared gw mac, thus allowing an non-ovn port to be attached to the
shared gw bridge and function normally.

**- How to verify it**

### DIFF
--- a/go-controller/pkg/node/gateway_localnet.go
+++ b/go-controller/pkg/node/gateway_localnet.go
@@ -353,7 +353,7 @@ func getLoadBalancerIPTRules(svc *kapi.Service, svcPort kapi.ServicePort, gatewa
 // -- to also connection track the outbound north-south traffic through l3 gateway so that
 //    the return traffic can be steered back to OVN logical topology
 // -- to also handle unDNAT return traffic back out of the host
-func addDefaultConntrackRulesLocal(nodeName, gwBridge, gwIntf string, stopChan chan struct{}) error {
+func addDefaultConntrackRulesLocal(nodeName, macAddress, gwBridge, gwIntf string, stopChan chan struct{}) error {
 	// the name of the patch port created by ovn-controller is of the form
 	// patch-<logical_port_name_of_localnet_port>-to-br-int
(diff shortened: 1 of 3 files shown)


go-controller/pkg/node/gateway_localnet.go
```diff
@@ -492,9 +491,18 @@ func addDefaultConntrackRulesLocal(nodeName, gwBridge, gwIntf string, stopChan c
 		nFlows++
 	}
 
-	// table 1, all other connections go to host
+	// table 1, we check to see if this dest mac is the shared mac, if so flood to both ports
```
aojea, line 494: is this not going to solve the icmp6 dup issue?

trozet: with this, incoming icmp and icmp6 will go to both, so there will be 2 replies (one from host and one from OVN)

trozet: we perhaps need to specifically steer icmp ping requests to only one of them

aojea: my question is if we can fix it here or it will need more work , doesn't seem efficient that I take on something that you can fix  here :)

trozet: sorry this comment is wrong. I think I forgot to change the comment before. This only will send to the local host, which is what we want if the traffic is IP and the dest mac is our host mac. So ICMP will only go to the host. However for icmp6 we need to edit our hack flow to only be for neighbor discovery response.

trozet: for the  icmp6 hack we need to look at icmp6_type in openflow and either:
1. match on ICMP request and only steer it to the local host OR
2. match on reply icmp6 reply traffic like neighbor discovery and send it to OVN

I think 1 is probably easier.

PR #347, opened 2020-11-16
Bug 1898269: Fix broken make check on 4.6

alexanderConstantinescu:
The release-4.6 branch are broken when running the command `make check`.
This commit fixes the build. 

I suspect this commit https://github.com/openshift/ovn-kubernetes/commit/15661c9d1d23fe0b30d266bda3668ee848884fb5 was the one that introduced the broken build. 

Tomorrow I am going to work on having `make check` be gating on release-4.6 and master (similar to `/test unit` on openshift/sdn) as to not have to do this again. 

/assign @trozet 

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/master_test.go
+++ b/go-controller/pkg/ovn/master_test.go
@@ -1045,6 +1045,9 @@ var _ = Describe("Gateway Init Operations", func() {
 			addNodeportLBs(fexec, nodeName, tcpLBUUID, udpLBUUID, sctpLBUUID)
 			fexec.AddFakeCmdsNoOutputNoError([]string{
 				"ovn-nbctl --timeout=15 --may-exist ls-add " + externalSwitchPrefix + nodeName,
+				// OCP HACK
+				"ovn-nbctl --timeout=15 --data=bare --no-headings --columns=_uuid find logical_switch_port name=br-local_" + nodeName,
+				// OCP HACK
 			})
 			fexec.AddFakeCmdsNoOutputNoError([]string{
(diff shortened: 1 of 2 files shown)


alexanderConstantinescu:
/retitle Bug 1898269: Fix broken `make check` on master

alexanderConstantinescu:
/retitle Bug 1898269: Fix broken make check on 4.6

abhat:
@alexanderConstantinescu I already had #329 to fix the same issue. I can close mine if you would like.

go-controller/pkg/ovn/master_test.go
```diff
@@ -3,14 +3,15 @@ package ovn
...
+	"net"
+	"strings"
+
 	goovn "github.com/ebay/go-ovn"
 	"github.com/urfave/cli/v2"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/tools/record"
-	"net"
```
trozet, line 12: dont move this

trozet (changes requested):

go-controller/pkg/ovn/master_test.go
alexanderConstantinescu: Why? Go standard library imports should be declared first right (according to the convention), so it's wrong today. Is it just as to not increase the diff?

trozet: it can crate conflicts when we merge later downstream. If its wrong fix it in upstream then downstream

dcbw:
/retest
`msg="Error: Error when reading or editing ComputeRegionBackendService \"projects/openshift-gce-devel-ci/regions/us-east1/backendServices/ci-op-iz744ttd-9e101-7qd2j-api-internal\": googleapi: Error 403: Quota exceeded for quota group 'ReadGroup' and limit 'Read requests per 100 seconds' of service 'compute.googleapis.com' for consumer 'project_number:1053217076791'., rateLimitExceeded`

PR #374, opened 2020-12-07
Bug 1899266: Fixes ICMPv6 flooding behavior

trozet:
Our flows for flooding icmpv6 to both OVN and the host were too generic.
This caused ICMPv6 echo requests to be sent to both, resulting in a
duplicate ack and other issues. This patch changes the flows for icmpv6
to be more specific and target only flooding for Neighbor Advertisements
and Route Advertisements. This will allow OVN to receive only the
packets it needs for peering/discovery, while all other ICMPv6 will go to
the host only.

(cherry picked from commit 1206bc2ba30d35a9c896b3cddf6f2f8fd6694583)

**- How to verify it**


trozet:
azure failure is a known bug that has nothing to do with this patch:
```
"Pod openshift-monitoring/cluster-monitoring-operator-7c545b7cc-9588z was pending entire time: unknown error",
```

/override ci/prow/e2e-azure-ovn

PR #366, opened 2020-12-01
12-01-2020 merge

alexanderConstantinescu:
**- What this PR does and why is it needed**

Master pull from upstream -> downstream

/assign @dcbw @trozet 

**- Special notes for reviewers**

**- How to verify it**

**- Description for the changelog**

PR #388, opened 2020-12-16
Bug 1908472: [4.6] Increase QPS and Burst setting

dcbw:
Our client will under high load exceed the default QPS (query per second)
and Burst setting when communicating with the API server, which will cause
the client request to get throttled by the API server. These default
values are currently 5 and 10 respectively. Scalability testing has shown that
25 and 25 are values which lead to reasonable pod ready latency times
under high load.

@alexanderConstantinescu @trozet


markmc:
(patch manager) Fixes a serious issue uncovered in a scale test. Backport of ovn-org/ovn-kubernetes#1878 which is already on master as of #366

PR #389, opened 2020-12-17
Bug 1908803: [4.6] fix network policy when the namespace label changes

JacobTanenbaum:
Currently when using a namespaceSelector and podSelector in the same network
policy when changing the labels of the namespace nothing happens.
Correct the operation so that when updating the label removes a
namespace from a network policy the pods are correctly removed

**- How to verify it**


markmc:
(patch manager) backport of ovn-org/ovn-kubernetes#1850 (already merged in 4.7) for long-standing issue

PR #390, opened 2020-12-18
Bug 1908921: ovn/node: use node name as chassis hostname

dcbw:
Backport of https://github.com/ovn-org/ovn-kubernetes/pull/1653

In some cloud environments the Kube node name is a short name, while
the node's "hostname" is an FQDN. eg:

node.Name: ci-op-h2qtbcsi-99b10-qhxj2-worker-d-6wxxl
hostname:  ci-op-h2qtbcsi-99b10-qhxj2-worker-d-6wxxl.c.openshift-gce-devel-ci.internal

syncNodesPeriodic() assumed they were going to be the same, which
resulted in deleting all chassis records from the SBDB every five
minutes, because ovnkube-node was setting the node's FQDN as the

@trozet @fepan @fedepaol

### DIFF
--- a/go-controller/pkg/node/node.go
+++ b/go-controller/pkg/node/node.go
@@ -47,11 +47,6 @@ func NewNode(kubeClient kubernetes.Interface, wf *factory.WatchFactory, name str
 func setupOVNNode(node *kapi.Node) error {
 	var err error
 
-	nodeName, err := util.GetNodeHostname(node)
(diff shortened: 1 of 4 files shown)


markmc:
(patch manager) backport of a patch that's been in 4.6 and 4.7 for a while now, fixes a serios issue seen by a real-world user

PR #391, opened 2020-12-18
Bug 1886062: Backport klog v2 changes to 4.6

abhat:
Rebased and re-vendorized after commit 4efbb599 already brought
in klog v2 into vendor.

Signed-off-by: Thomas F Herbert <therbert@redhat.com>
(cherry picked from commit 40ec945bffd7282171a031c6e192ca1e8903897b)


go-controller/vendor/modules.txt
```diff
@@ -586,9 +588,11 @@ k8s.io/client-go/util/testing
 k8s.io/client-go/util/workqueue
 # k8s.io/klog v1.0.0
 k8s.io/klog
+# k8s.io/klog/v2 v2.0.0
```
aojea, line 591: :thinking: why is still the klog v1.0.0 dependency there? 
I think it should go away after the update, no?

tfherbert: No, I don't think so. Much  vendor code is still using klog v1. v2 is in a subdirectory so both dependencies should go-exist.

russellb:
(patch manager) 4.6 blocker identified, so pulling unmerged patches from the queue

tfherbert:
What is the status of this backport to 4.6? Can it be merged yet?
The bugzilla 1886062 has a needinfo flag.

PR #392, opened 2020-12-21
[WIP][TEST] New service controller using endpoint slices

aojea:
test only


aojea:
this doesn't make any sense, now the gcp job works without any change, it was failing to install before

aojea:
> this doesn't make any sense, now the gcp job works without any change, it was failing to install before

ok, I think that this may explain it, the kubernetes version has to be bumped independently in each component
https://github.com/openshift/cluster-dns-operator/pull/222
and then accepted in the release
https://origin-release.apps.ci.l2s4.p1.openshiftapps.com/releasetag/4.7.0-0.okd-2020-12-21-154920?from=4.7.0-0.okd-2020-12-17-131741

aojea:
Ok, this time it went through the 3 cloud providers,

the `reject packet for service with endpoints` test keeps failing though on the 3 jobs

aojea:
this is pretty good @dcbw 2/2 passes with the new ovn version and the new feature in OVN that reject the services without endpoints
https://github.com/openshift/ovn-kubernetes/pull/392/commits/81751428800b371a7f4a15bce9888212300d4769
https://github.com/openshift/ovn-kubernetes/pull/392/commits/a15efbd032166639fe580d89ca98859101307e1c

What do you think if instead of trying to fix the ACLs with the new service endpoint slice controller I wait for this version?

PR #399, opened 2021-01-11
Dockerfile: bump OVN to 20.09.0-23

aojea:
OVN implements the option to reject packets destined to LoadBalancers
without endpoints, simplifying OVN-Kubernetes, since it doesn´t have to
handle the reject traffic with ACLs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ RUN yum install -y  \
 	yum clean all
 
 ARG ovsver=2.13.0-72.el8fdp
-ARG ovnver=20.09.0-21.el8fdn
+ARG ovnver=20.09.0-23.el8fdn
 
 RUN INSTALL_PKGS=" \
 	openssl python3-pyOpenSSL firewalld-filesystem \


Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN yum install -y  \
 	yum clean all
 
 ARG ovsver=2.13.0-72.el8fdp
-ARG ovnver=20.09.0-21.el8fdn
+ARG ovnver=20.09.0-22.el8fdn
```
aojea, line 36: seems this is not available

alexanderConstantinescu: Indeed that version does not exist, here is the list of all available versions (need to be on the VPN):

https://brewweb.engineering.redhat.com/brew/packageinfo?packageID=74031

And this is not the right way of bumping packages. You need to

1. Tag the brew build you want with a specific OCP tag recognized by ART so that the proper image version can be pulled in, these tags are: `rhaos-4.6-rhel-8-candidate`, `rhaos-4.7-rhel-8-candidate`
2. Wait a couple of hours for everything to get propagated (this might not be required anymore)
3. Create this PR and specify the version you tagged in step 1. 

I guess what you want is:

```
* Mon Dec 14 2020 Lorenzo Bianconi <lorenzo.bianconi@redhat.com> - 20.09.0-22
- Backport "northd: add reject action for lb with no backends" (#1883957)
```

Which means you need `>= 20.09.0-23.el8fdn || 20.09.0-23.el8fdp`

alexanderConstantinescu: Sorry, I forgot a key piece of information: this is how you tag brew builds; https://source.redhat.com/personal_blogs/how_to_tag_brewweb_rpm_builds

aojea: The docs in https://brewweb.engineering.redhat.com/brew/packageinfo?packageID=74031 were useful, thank you very much

alexanderConstantinescu: Actually I saw you tagged the build with both `rhaos-4.6-rhel-8-candidate` and  `rhaos-4.7-rhel-8-candidate`. Don't do that, use only the tag corresponding to the branch you target (in this case `rhaos-4.7-rhel-8-candidate`), otherwise you'll bump 4.6 too, which we don't want. I removed the 4.6 tag from that build

aojea: upps, thanks, I took the comment literally :)

alexanderConstantinescu:
Also, I forgot to ask: @aojea  this is missing a BZ. This will not go in on master until Friday without a BZ, so the question is: should it? 

A problem arising from the way we are pinning versions in these images is that we are now effectively breaking all our builds on master because ART is now taking 20.09.0-23 (which we've tagged) but the images still pin to 20.09.0-21.el8fdn (because this PR has not merged). 

If this can wait until 4.8, then we should un-tag that brew build as to stop breaking master.

aojea:
> If this can wait until 4.8, then we should un-tag that brew build and stop breaking master. If it's required for 4.7, then we need a BZ.

This will allow us to get rid of all the ACL logic for services in ovn-kube, I have a pr working upstream https://github.com/ovn-org/ovn-kubernetes/pull/1928 using this ovn version
Let's ask @danwinship and @dcbw  what do they think

danwinship:
yeah, the plan was to get this in for Friday (so it needs a bz)

danwinship:
> yeah, the plan was to get this in for Friday (so it needs a bz)

hm... I guess I misunderstood the current state of things when we talked yesterday... I had thought that the new reject ACL code was already merged upstream. But it looks like it's not yet, that's https://github.com/ovn-org/ovn-kubernetes/pull/1928, right?

This is basically a new feature, and feature freeze was a long time ago, so it would be better to figure out why the old code isn't working, rather than totally rewriting it all at the last minute

aojea:
> This is basically a new feature, and feature freeze was a long time ago, so it would be better to figure out why the old code isn't working, rather than totally rewriting it all at the last minute

heh, I´m sure that I didn´t explain it correctly, I wasn´t expecting to include this neither :smile:

PR #400, opened 2021-01-11
Bug 1915027: Fix MCS-blocking iptables rules

danwinship:
The MCS-blocking rules were too aggressive, and would block connections using 22623/22624 as a source port in some cases.

`--syn` means "SYN and not ACK", so this should make it only block attempts to establish connections to port 22623/4, and not any other traffic involving those ports.

### DIFF
--- a/go-controller/pkg/cni/OCP_HACKS.go
+++ b/go-controller/pkg/cni/OCP_HACKS.go
@@ -13,10 +13,10 @@ import (
 // OCP HACK: block access to MCS/metadata; https://github.com/openshift/ovn-kubernetes/pull/19
 var iptablesCommands = [][]string{
 	// Block MCS
-	{"-A", "OUTPUT", "-p", "tcp", "-m", "tcp", "--dport", "22623", "-j", "REJECT"},
-	{"-A", "OUTPUT", "-p", "tcp", "-m", "tcp", "--dport", "22624", "-j", "REJECT"},
-	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22623", "-j", "REJECT"},
-	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22624", "-j", "REJECT"},
+	{"-A", "OUTPUT", "-p", "tcp", "-m", "tcp", "--dport", "22623", "--syn", "-j", "REJECT"},
+	{"-A", "OUTPUT", "-p", "tcp", "-m", "tcp", "--dport", "22624", "--syn", "-j", "REJECT"},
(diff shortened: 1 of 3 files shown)


danwinship:
/hold
@trozet pointed out a problem

danwinship:
Updated (hackily) to delete the old rules on upgrade.

We should make ovn-kubernetes use its own tables the way that openshift-sdn does. (ie, rather than adding to `OUTPUT` directly, it should create an `OVN-KUBERNETES-OUTPUT` chain and add things there, and only add a single `-A OUTPUT -j OVN-KUBERNETES-OUTPUT` rule to the main chain, so then it's easier for us to manage our rules on upgrade

abhat:
@danwinship do we need to be more granular and add the rules only for node IPs on which the MCS listens? Right now the rules will deny request to a pod that runs on any of those ports at a different cluster IP.

trozet:
I agree with @abhat here. I think we are just solving part of the problem with this. If anyone runs an external server with these same ports, or another pod uses these ports as a server. The pods wont be able to talk to it. It seems like we need a better solution here...or we need to document that you cannot use these ports in your openshift deployment. Either MCS needs better security, or we need to watch for nodes, and add rules based on node ips.

trozet:
@danwinship couldn't we just add iptables rules to the nodes where MCS runs to reject any connection from the pods subnet instead?

danwinship:
> I think we are just solving part of the problem with this.

We are absolutely only solving part of the problem. This was always known to be totally half-assed.

> or we need to watch for nodes, and add rules based on node ips.
...
> couldn't we just add iptables rules to the nodes where MCS runs to reject any connection from the pods subnet instead?

part of the problem is that we need to block pods from using multus IPs to connect to the MCS as well. That means adding rules based on node IPs is difficult, because the blocking has to happen in the pod network namespace, so if the set of blocked destinations changes, we would have to update every pod on every node with the new destinations. And it means the MCS can't just reject connections from the pod subnet, because multus IPs wouldn't be from the pod subnet.

If some customer complains that they have an external server using port 22624 that they can't access, then we'll reassign that bug to the MCO team and force them to prioritize fixing the MCS faster.

/hold cancel

trozet:
/hold

let's add  ! -d <pod subnet> to the rule so at least pods can use the ports between each other

danwinship:
> let's add ! -d pod subnet to the rule so at least pods can use the ports between each other

I had agreed to that initially because I thought it was going to be simpler, but I don't think it's worth adding code to look up the value of the pod subnet in the CNI plugin just so that we can do this. No one has ever complained about not being able to access any services on that port before, and the hack will eventually go away, so it's not worth spending too much time trying to make it perfect.

PR #407, opened 2021-01-14
Bug 1908888: Check IP family for exgws

rcarrillocruz:
We need to ensure we only add the proper family routes when exgw
annotations are used.

Signed-off-by: Ricardo Carrillo Cruz <ricarril@redhat.com>


rcarrillocruz:
@trozet here it is the backport for the check ip family fix

PR #408, opened 2021-01-14
Bug 1903651: Add clusterIP to ingress policy AS for SNAT-ed hairpin Traffic Cherry Pick 

astoycos:
This PR fixes the problem found here https://bugzilla.redhat.com/show_bug.cgi?id=1903651

When a POD sends traffic to a service it is loadbalanced(DNAT-ed) and sometimes the same pod is chosen as the backend to the service. To ensure this traffic travels through OVN-K8s rather than the POD2POD network since (srcIP==dstIP), the traffic is SNAT-ed to the VIP of the service's loadbalancer, which results in some packet loss with hairpinned traffic since this is not currently accounted for in the network Policy code

Temporary solution:

In the long term, a feature is being created that will allow for us to specify the loadbalancer's SNAT IP in OVN see -> https://bugzilla.redhat.com/show_bug.cgi?id=1908540. This will allow us to make a few rules to cover all the possible SNAT IPs, which is not possible now since the SNAT ip selection process for hairpin traffic is automatic

However for shorter term solution this PR will add the ClusterIP, node IPS, external IPs and Ingress IPS of any services that are applied in the same namespace as a NetworkPolicy to the policy's ingress rule's address set.

This involves adding...

a new service handler that only looks at services for specific namespaces
a function in service.go to get the possible VIPs of the Service
some work in policy.go to add service handlers in the same namespace as the network policy whenever ingress rules with selectors are applied
Thankfully this will not result in any security gaps since the only time a packet will have a src.ip of the SVC's cluster IP is in with the hairpin case.

### DIFF
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -399,6 +399,11 @@ func (wf *WatchFactory) AddServiceHandler(handlerFuncs cache.ResourceEventHandle
 	return wf.addHandler(serviceType, "", nil, handlerFuncs, processExisting)
 }
 
+// AddFilteredServiceHandler adds a handler function that will be executed on all Service object changes for a specific namespace
+func (wf *WatchFactory) AddFilteredServiceHandler(namespace string, handlerFuncs cache.ResourceEventHandler, processExisting func([]interface{})) *Handler {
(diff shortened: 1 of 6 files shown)


astoycos:
@trozet

PR #412, opened 2021-01-18
[wip] Upstream merge 2021-01-18

dcbw:
@trozet soak test for now, but maybe we want it for 4.7?


trozet:
i dont think we want to do this right now, we can talk offline

dcbw:
> i dont think we want to do this right now, we can talk offline

@trozet no problem; we don't have to. Soak test might still be interesting though. It's not as smooth as current master, I think.

PR #418, opened 2021-01-20
Bug 1915295: fix service controller repair loop race

aojea:
The new service controller for endpoint slices has a repair
loop that runs once, during the controller startup process,
so we can delete out of sync data between OVN and Kubernetes,
specially to deal with upgrades and OVN new features.

The repair loop was using the controller service tracker as a
source of true, but this is racy because the service tracker
is updated just after the OVN Load Balancer has been configured,
so it can happen that the Service has an OVN loadbalancer, but
doesn´t exist on the service tracker. As a consequence the repair
loops things that is a stale loadbalancer and deletes it.

The repair loop must use the informer cache as source of true, also
it must run before starting the informers workers, so we avoid any
races.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1915295


aojea:
/retest
the dual stack job failed with a different error and is not using this controller, for what I see in the logs

PR #419, opened 2021-01-20
Bug 1914250: support multipath to detect default gateway

aojea:
We were detecting the default gateway based in the assumption that
the netlink library was populating the de Gw field with it, however,
it turns out that when multipath is enabled that field is empty and
the corresponding information is in the MultiPath field.

Also, in order to obtain the default gateway, we were dumping all
routes filtering by FAMILY_ALL, and iterating on the results to get
the ones we are interested. We can filter directly by the IP Family instead,
to avoid future possibles problems with the FAMILY_ALL interpretation.

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1914250

### DIFF
--- a/go-controller/pkg/node/helper_linux.go
+++ b/go-controller/pkg/node/helper_linux.go
@@ -5,12 +5,13 @@ package node
 import (
 	"fmt"
 	"net"
-	"syscall"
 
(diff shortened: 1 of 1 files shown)


vpickard:
cc @astoycos This may help with the issue you were seeing in KIND with your CNO work.

PR #420, opened 2021-01-20
Bug 1917605: Fixes deleting exgw pod

trozet:
The ovn_cluster_router policy was being removed for a pod anytime an
exgw was deleted. This is incorrect, because a pod may have more than
one exgw configured. The policy should only be removed for a pod if it
no longer has any exgws configured.

Additionally, the policy was being removed for potentially the wrong
node. This patch corrects the node to be the one where the affected pod
lives.


trozet:
@abhat @alexanderConstantinescu PTAL

PR #424, opened 2021-01-26
Bug 1920482: Configure GARP for egress IP re-assignment

alexanderConstantinescu:
NOTE: back-port of #422

Egress IP assignments can change depending on the health check detection
of routes to egress nodes. Once that re-assignment happens we previously
experienced several minutes of packet loss between the source pods and
any external services they were targetting because no GARP was configured
on the external gateway routers by OVN. This lead to several minutes of
interrupted traffic until the ARP cache expired between client and server
for the egress pods.

I am only configuring GARP on egress-assignable nodes, and removing the
configuration once the egress node is flagged as not usable anymore.

/assign @danwinship 

**- How to verify it**


jwforres:
(Patch Manager) merge window is closed, removing cherry pick approval

PR #428, opened 2021-01-29
Bug 1892472: Run linter, utests, and gofmt in container

abhat:
Cherry-pick changes from github.com/ovn-org/ovn-kubernetes#1935 and github.com/ovn-org/ovn-kubernetes#1976 for running lint, gofmt, and utest jobs in containers.

### DIFF
--- a/go-controller/Makefile
+++ b/go-controller/Makefile
@@ -9,6 +9,18 @@ PKGS ?=
 GOPATH ?= $(shell go env GOPATH)
 TEST_REPORT_DIR?=$(CURDIR)/_artifacts
 export TEST_REPORT_DIR
+GO_DOCKER_IMG = quay.io/giantswarm/golang:1.15.7
+# CONTAINER_RUNNABLE determines if the tests can be run inside a container. It checks to see if
+# podman/docker is installed on the system and also confirms that it is not running in a CI environment by checking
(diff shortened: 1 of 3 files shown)


go-controller/Makefile
```diff
@@ -9,6 +9,18 @@ PKGS ?=
 GOPATH ?= $(shell go env GOPATH)
 TEST_REPORT_DIR?=$(CURDIR)/_artifacts
 export TEST_REPORT_DIR
+GO_DOCKER_IMG = quay.io/giantswarm/golang:1.15.7
```
aojea, line 12: do we trust this?

squeed: The standard is to use `golang:1.15-alpine`, which I realize is DockerHub. We could make this overridable with a `?=` and point to a CI-hosted image for automated runs.

abhat: Yeah, we can make this a downstream specific change. I think anything that is not hosted in Dockerhub which is relatively "official" will work.

Do you feel ok with merging this for now since this is a cherry-pick of upstream and I will create a downstream only PR to update that image to CI-host one.

aojea: it was mostly curiosity, not a blocking question :)

PR #431, opened 2021-02-05
Create a podman image to use with CI so linter/utests can be run

abhat:
For OpenShift CI, we need to create a custom image that has podman
installed in it to run the containerized linter, gofmt, and utests
There is no base image that will provide podman by default. So the
alternative is to create our own image and use it in the tests.

### DIFF
--- a/Dockerfile.utest
+++ b/Dockerfile.utest
@@ -1,0 +1,14 @@
+FROM registry.centos.org/centos/centos:7
+
+# base: EPEL repo for extra tools
+RUN yum -y install epel-release
+
+# build: system utilities and libraries
+RUN yum update -y && \
+    yum -y install epel-release && \
+    yum -y groupinstall 'Development Tools' && \
+    yum -y install openssl-devel protobuf-compiler jq podman && \
+    yum -y install yamllint && \
+    yum -y install cmake elfutils-libelf-devel libcurl-devel binutils-devel elfutils-devel && \
+    yum clean all
+


abhat:
This will eventually be used with: openshift/release#13218

Dockerfile.utest
```diff
@@ -0,0 +1,14 @@
...
+RUN yum -y install epel-release
+
+# build: system utilities and libraries
+RUN yum update -y && \
+    yum -y install epel-release && \
+        yum -y groupinstall 'Development Tools' && \
+	    yum -y install openssl-devel protobuf-compiler jq skopeo buildah podman &&\
+	        yum -y install yamllint && \
+		    yum -y install cmake elfutils-libelf-devel libcurl-devel binutils-devel elfutils-devel && \
+		        yum clean all
```
aojea, line 13: is the indentantion on purpose?

Dockerfile.utest
```diff
@@ -0,0 +1,14 @@
+FROM registry.centos.org/centos/centos:7
+
+# base: EPEL repo for extra tools
+RUN yum -y install epel-release
+
+# build: system utilities and libraries
+RUN yum update -y && \
+    yum -y install epel-release && \
+        yum -y groupinstall 'Development Tools' && \
+	    yum -y install openssl-devel protobuf-compiler jq skopeo buildah podman &&\
```
aojea, line 10: I think the backslash needs an space `&& \`

rcarrillocruz: is skopeo and buildah really needed?

Dockerfile.utest
abhat: i think its just readability. not something i consciously did either.

Dockerfile.utest
abhat: perhaps not. copy-paste of another dockerfile from Cincinnati :).

aojea:
/lgtm
let´s see how it behaves in the CI :crossed_fingers:

PR #433, opened 2021-02-12
Bug 1918442: Fix idling and rejects for 4.7

aojea:
Temporary fix idling and reject with the new service controller.
In 4.8 we leverage OVN for doing that, in 4.7 we use ACLs


aojea:
/retest
@trozet the reject ACLs are working :clap:

PR #435, opened 2021-02-16
[wip] [4.7] 2020-02-16 scale rollup

dcbw:
OVS 2.15 for column diffs; latest OVN for logical datapath groups and hairpin LB flow reductions, enable memory trimming on compaction, etc.


dcbw:
most of these are backported to 4.7 already

PR #436, opened 2021-02-17
Bug 1929314: detect if the cluster has endpoint slices

aojea:
the endpoint slice controller was checking if the kubernetes
version was greater than 1.20 in order to run, so it can verify
the dual-stack api was implemented.

Since the endpoint slice controller in ovn can run with or without
the dualstack api, we only need to check that the API for slices
is enabled.


aojea:
Merged during the rebase

PR #442, opened 2021-02-24
Bug 1932268: detect if the cluster has endpoint slices

aojea:
the endpoint slice controller was checking if the kubernetes
version was greater than 1.20 in order to run, so it can verify
the dual-stack api was implemented.

Since the endpoint slice controller in ovn can run with or without
the dualstack api, we only need to check that the API for slices
is enabled.

Backport https://github.com/openshift/ovn-kubernetes/commit/cdabdc3244de873beaac168b7f613f445bdda380

### DIFF
--- a/go-controller/pkg/util/kube.go
+++ b/go-controller/pkg/util/kube.go
@@ -3,7 +3,6 @@ package util
 import (
 	"encoding/json"
 	"fmt"
-	"strconv"
 	"strings"
 
 	kapi "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


go-controller/pkg/util/kube.go
```diff
@@ -240,22 +239,11 @@ func EventRecorder(kubeClient kubernetes.Interface) record.EventRecorder {
 	return recorder
 }
 
-// UseEndpointSlices if the EndpointSlice API is available
-// and if the kubernetes versions supports DualStack (Kubernetes >= 1.20)
+// UseEndpointSlices detect if Endpoints Slices are enabled in the cluster
 func UseEndpointSlices(kubeClient kubernetes.Interface) bool {
-	endpointSlicesEnabled := false
 	if _, err := kubeClient.Discovery().ServerResourcesForGroupVersion(discovery.SchemeGroupVersion.String()); err == nil {
```
dcbw, line 244: @aojea I don't understand how this works, but I'm sure I'm missing something. I see this pattern in some other places, but how does it actually ensure that EndpointSlices are present?

aojea:
```
discovery.SchemeGroupVersion.String()
```
returns the client group-version for endpoint slices (discovery is the group for slices)

And the ServerResourcesForGroupVersion() ask the apiserver if it is supported

https://github.com/kubernetes/client-go/blob/8f8241f9ef2dc939ecaf3cbdc80f6eb58227efd9/discovery/discovery_client.go#L88-L90
```
// ServerResourcesInterface has methods for obtaining supported resources on the API server
type ServerResourcesInterface interface {
	// ServerResourcesForGroupVersion returns the supported resources for a group and version.
	ServerResourcesForGroupVersion(groupVersion string) (*metav1.APIResourceList, error)
```

dcbw: @aojea aha, that's what I was missing, that Discovery is the group for slices. Totally non-obvious.

aojea: yeah, that is the complicated part with slices, is not easy to map slices to them, and in addition here, the method to know if the version is supported it called `.Discovery()`

trozet:
@aojea do we have the service reject fixes in 4.7?

Dualstack job keeps failing on:
Services should be rejected when no endpoints exist [

aojea:
> @aojea do we have the service reject fixes in 4.7?
> 
> Dualstack job keeps failing on:
> Services should be rejected when no endpoints exist [

yeah, seems the dual-stack jobs now run all tests , grest
It was missing your fix for reject acls

aojea:
> @aojea do we have the service reject fixes in 4.7?
> 
> Dualstack job keeps failing on:
> Services should be rejected when no endpoints exist [

@trozet PTAL, it was missing the service reject fixes

aojea:
weird, 
```
[sig-network] Services should be rejected when no endpoints exist [Skipped:ibmcloud] [Suite:openshift/conformance/parallel] [Suite:k8s]
``` 
in the dualstack job

PR #467, opened 2021-03-16
Bug 1939488: Backport Handle Multus network-status annotations on pod update

vpickard:
Currently we require that if you specify an external gateway that corresponds
to a multus network, then you must do so as a pod update to add the exgw
annotations. This is due to the fact that multus will not add the network
attachment definition until after OVN CNI runs, and therefore OVN will not be
able to locate the exgw information. The current workflow is this:

1. Create exgw pod (OVN will add pod, multus will add network attachment def)
2. Once created, edit the pod and annotate it with the exgw annotations that
link to a multus network:

k8s.v1.cni.cncf.io/network-status: "[{\"name\":\"blah\",\"interface\":\"net1\",
\"ips\":[\"172.18.0.4\"],\"mac\":\"62:ff:69:f0:83:03\",\"dns\":{}}]"

k8s.ovn.org/routing-namespaces: exgw
k8s.ovn.org/routing-network: blah

However, we should be able to handle including the routing annotations on the
pod during pod add. When multus runs after OVN and annotates the network
attachment definition, OVN-Kube will get an update event from k8s and should
be able to then reconcile the routes.

As it turns out, this was already handled in addPodExternalGW() logic.
This change will handle the case where the network-status annotatations
change, for example, if the IP is changed by manually editing the pod
spec and changing the IP in the network-status annotation.

Signed-off-by: vpickard <vpickard@redhat.com>
(cherry picked from commit d6db5ea310ade159ba97fda81d1eef78735a8540)

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/ovn.go
+++ b/go-controller/pkg/ovn/ovn.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	goovn "github.com/ebay/go-ovn"
+	nettypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	hocontroller "github.com/ovn-org/ovn-kubernetes/go-controller/hybrid-overlay/pkg/controller"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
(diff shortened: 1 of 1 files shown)


knobunc:

We keep flaking on infrastructure problems (image pull failures, etc.) and have previously passed that job with this PR.

PR #470, opened 2021-03-18
Bump OVN to ovn2.13-20.12.0-25.el8fdp

trozet:
Includes a required fix for force_snat_ip=router_ip for upcoming shared
gw.


trozet:
@dcbw PTAL

PR #478, opened 2021-03-25
Bug 1943316: [4.7] master: enable logical datapath groups for OVN >= 20.12

dcbw:
Logical datapath groups reduce the size of the southbound
database in large clusters. The option is ignored if unsupported
by ovn-northd. ovn-controllers should be upgraded to a version
that supports this option before it is turned on.

Cherry-pick of https://github.com/ovn-org/ovn-kubernetes/pull/2074 that was pulled into 4.8 as part of https://github.com/openshift/ovn-kubernetes/pull/440

@trozet

PR #490, opened 2021-04-07
Bug 1946899: master: enable logical datapath groups for OVN >= 20.12

alexanderConstantinescu:
Since the CNO patch is already on 4.7, this back-port should work

------

Logical datapath groups reduce the size of the southbound
database in large clusters. The option is ignored if unsupported
by ovn-northd. ovn-controllers should be upgraded to a version
that supports this option before it is turned on.

/assign @dcbw 

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -211,6 +211,31 @@ func (oc *Controller) upgradeOVNTopology(existingNodes *kapi.NodeList) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


dcbw:
@alexanderConstantinescu this is a dupe of https://github.com/openshift/ovn-kubernetes/pull/478 which already has approve/lgtm right?

alexanderConstantinescu:
> @alexanderConstantinescu this is a dupe of #478 which already has approve/lgtm right?

Thanks for letting me know, I didn't think to look during my back-port frenzy this morning

/close

PR #494, opened 2021-04-09
Changes to support network policy for host network traffic

abhat:
This commit adds support for enabling network policy on host
network traffic pods such as the router pod in OpenShift.

Currently, if you have a default deny network policy on a namespace,
services within that namespace are not reachable via host network
using external routes. This feature will make that possible by allowing
users to create a separate network policy with a host network namespace selector.
   
The expectation is that the namespace you want to use to classify host network
traffic to is pre-created with the desired labels. These labels will then
be used as match fields for the network policy object's ingress selector.
It is also necessary that the users pass the host network namespace
as a command line parameter for ovnkube when it starts.
   
**- How to verify it**

### DIFF
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -175,9 +175,11 @@ parse_args() {
             -dl  | --ovn-loglevel-nbctld )      shift
                                                 OVN_LOG_LEVEL_NBCTLD=$1
                                                 ;;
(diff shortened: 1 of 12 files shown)


trozet:
/hold

we will just do a merge

abhat:
Superseded by #496.

PR #497, opened 2021-04-12
Bug 1930248: Sync management port for an existing node

alexanderConstantinescu:
This adds a commit which should be adding correctness to the management port synchronization. It does not fix the bug mentioned in the title, but it should improve the general fix for what concerns syncing the state w.r.t to this specific OVN-Kubernetes setup. I am however using the bug, as to get the fix in on 4.8 and then back-port both the commit that did fix this issu, and this one: which improves the general fix.  

/assign @dcbw @trozet 

**- How to verify it**


trozet (approved):
/lgtm

PR #498, opened 2021-04-14
Bug 1945948: Fixes local node IP reachability in shared gateway mode

trozet:
Ovnkube-node will now detect IPs local to the node. Any unicast IP addresses that are non-loopback are added to an annotation on the node: k8s.ovn.org/host-addresses

These are then read by ovnkube-master who will add new policy routes on ovn_cluster_router with priority 1004 in order to allow pods to reach the IPs on the node.


trozet:
@alexanderConstantinescu @dcbw PTAL

dcbw:
gcp failure is `: [sig-api-machinery] API priority and fairness should ensure that requests can be classified by testing flow-schemas/priority-levels [Suite:openshift/conformance/parallel] [Suite:k8s]`
dualstack didn't even get the platform set up

/retest

PR #500, opened 2021-04-14
Bug 1947712: pods: bind pod logical switch ports to the node's chassis with requested-chassis

trozet:
To prevent port binding ping-pongs between ovn-controllers, if an
ovnkube-node isn't running or hasn't been able to clear the iface-id
for an old instance of a pod, but the pod has been rescheduled,
request the chassis the pod has been scheduled on.


trozet:
@alexanderConstantinescu @dcbw

PR #502, opened 2021-04-15
Bug 1947917: fix deadlock in EgressFirewall DNS code

JacobTanenbaum:
Currently when adding EgressFirewalls with 5 or so dnsNames (observed
with 5 dnsNames) it is probable that a deadlock will occur.

thread #1 calls (e *EgressDNS) Add() grabs lock e.lock and calls
signalAdded(dnsName) and continues looping to the next dnsName calling
(e *EgressDNS) Add() grabbing e.lock and trying to write to the channel

thread #2 recieves the signalAdded(dnsName) and is waiting to grab
e.lock in updateEntry for name

thread #1 cannot write to the channel because the channel blocks waiting
for the write and holds e.lock

thread #2 cannot continue with updateEntryForName() because it is wating
for e.lock

Solve this problem by taking the add goroutine out of the long running
update goroutine. By spawning a goroutine for every DNS name the
goroutines can wait until it is there turn to update.

includes a small e2e test that should exercise the deadlocking codepaths

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/egressfirewall_dns.go
+++ b/go-controller/pkg/ovn/egressfirewall_dns.go
@@ -24,7 +24,7 @@ type EgressDNS struct {
 	addressSetFactory addressset.AddressSetFactory
 
 	// Report change when Add operation is done
-	added          chan string
+	added          chan struct{}
 	stopChan       chan struct{}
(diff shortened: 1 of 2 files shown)


trozet (approved):
/lgtm

PR #505, opened 2021-04-19
Bug 1951064: master: cancel leader election on exit

trozet:
When the master is terminated instead of just exiting, give leader
election a chance to release the leader lock so another ovnkube
master can grab it while this instance is down.

(cherry picked from commit 9c1d0c193d736d2d8180e5b0e5dc3c87f0b9ff03)


trozet:
@dcbw PTAL

dcbw:
/retest
`Error creating: pods "aws-ebs-csi-driver-controller-7b4bf6cdc8-" is forbidden: unable to validate against any security context constraint: [provider restricted: .spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.containers[0].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used `
(not network related)
/lgtm

PR #509, opened 2021-04-21
Bug 1943334: Taint node with NoSchedule effect when ovnkube pod is down

tssurya:

UseCase: During upgrades, when the ovnkube pod is deleted and
before/while it gets recreated if a new pod request comes, it
will fail.

When node networking is down, we should not allow pods to be created
on that node. This PR fixes this by adding a taint on the node with
the NoScedule effect when the ovnkube pod receives a SIGTERM and then
removes the taint when the ovnkube pod finishes setting up networking
and initializing the CNI config.

fixes #1943334

Unit test has not been added since it is difficult to simulate the plugin exec and start functions.

**- How to verify it**
The added taint will be visible on the node for the brief period of time between the deletion of ovnkube-node pod and creation of the new one that takes its place which roughly takes ~60sec.

`Taint node with NoSchedule effect when ovnkube pod is down and remove the taint when the pod is available again`


go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -121,6 +133,26 @@ func main() {
 		select {
 		case s := <-exitCh:
 			klog.Infof("Received signal %s. Shutting down", s)
+			if s == syscall.SIGTERM {
```
tssurya, line 136: note: There are a lot of nested ifs that I am not a fan of. However since we shouldn't really care to error out if we fail in one of the steps, I did this. But if folks feel I need to split them out to make it eye-friendly I can do that as well.

trozet:
@tssurya did you already post this upstream?

tssurya:
> @tssurya did you already post this upstream?

https://github.com/ovn-org/ovn-kubernetes/pull/2183

trozet:
normally we dont cherry pick downstream before merge upstream (unless you just want to WIP test it). This commit ID is different from the one upstream

/hold

jluhrsen:
@tssurya , @trozet , anything happening with this PR. I see the /hold so maybe it's not gonna make it. Asking becuase I'm
looking at this [issue](https://issues.redhat.com/browse/OCPBUGSM-27020), which is claiming to be blocked by the bug that this PR is for (among a few others)

tssurya:
> normally we dont cherry pick downstream before merge upstream (unless you just want to WIP test it). This commit ID is different from the one upstream
> 
> /hold

It has merged upstream: https://github.com/ovn-org/ovn-kubernetes/pull/2183.

I am closing this PR since the weekly merge PR should pick it up and put it downstream.

PR #515, opened 2021-04-24
Bug 1953169: endpoint slice controller doesn't handle services target port correctly

aojea:
Service targetPort is a selector for the endpoints/endpointslices
controller to create the endpoints based on that container port name.
It is not meant to be used in the Service implementation.
The relation is ServicePort.Name - EndpointPort.Name, however,
ServicePort.Name is only required for multiple ports and it may be empty.
If the endpoint matches the service and there is no name,
that means that is a single port service and there is only one endpoint.


aojea:
/retest
🤔

aojea:
/retest
the dual stack job doesn't seem to work

trozet (approved):
/lgtm

aojea:
> /override ci/prow/e2e-metal-ipi-ovn-dualstack

@dcbw it seems didn't override it

PR #519, opened 2021-05-03
Bug 1956270: service controller should not use the target-port

aojea:
Cherry-pick from https://github.com/openshift/ovn-kubernetes/pull/515

Service targetPort is a selector for the endpoints/endpointslices
controller to create the endpoints based on that container port name.
It is not meant to be used in the Service implementation.
The relation is ServicePort.Name - EndpointPort.Name, however,
ServicePort.Name is only required for multiple ports and it may be empty.
If the endpoint matches the service and there is no name,
that means that is a single port service and there is only one endpoint.


mrunalp:
@aojea @dcbw Please take a look at the test failures

PR #535, opened 2021-05-10
Bug 1947098: [4.6z] Ensure no SNAT on GR for DisableSNATMultipleGws

trozet:
When we have this configuration enabled, there should be no SNAT entry
for the pod subnet on the GR. Instead there should be per pod SNAT
entries added. On upgrade or enabling this feature we were failing to
remove any previous SNAT entries here.

(cherry picked from commit a9166cb0a998dd1d9c887269ec74ce2680bf6ff6)


alexanderConstantinescu (approved):
/lgtm

PR #538, opened 2021-05-12
[release-4.7] Bug 1959737: Fix: egress IP route health check detection state on restart

alexanderConstantinescu:
Egress IP initializes a "watcher" which health checks connectivity to egress
nodes every 5 seconds. This is done in its sync function. However, due to a
bug in the state initialization in WatchEgressNodes we don't set the
different egress node fields correctly in our ADD func which screws up the health
check detection mechanism upon a ovnkube-master restart.

Essentially it does like this:

1) egress setup is made
2) there's a problem with connectivity to the egress node(s)
3) ovnkube-master is restarted

Once step 3) is made: it is expected that ovnkube-master will perform
health checks continously and detect when the connectivity in step 2) is
back up. However because of this problem that is not the case and even
if further connectivity is successful, we never flag the node as such in
our internal data.

/assign @dcbw 

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/ovn.go
+++ b/go-controller/pkg/ovn/ovn.go
@@ -879,12 +879,16 @@ func (oc *Controller) WatchEgressNodes() {
 				klog.Error(err)
 			}
 			nodeLabels := node.GetLabels()
-			if _, hasEgressLabel := nodeLabels[nodeEgressLabel]; hasEgressLabel && oc.isEgressNodeReady(node) && oc.isEgressNodeReachable(node) {
+			if _, hasEgressLabel := nodeLabels[nodeEgressLabel]; hasEgressLabel {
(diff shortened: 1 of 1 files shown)


sdodson:
Persistently failing test

sdodson:
[patch-manager] This did not merge due to persistent CI failures, please investigate the failing tests.

alexanderConstantinescu:
@sdodson, FYI: I asked @trozet to override the vsphere job. It's been almost perma failing for a while, see: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-ovn-kubernetes-release-4.7-e2e-vsphere-ovn and this change/feature is not tested in CI, so it could not have impacted the CI tests.

PR #539, opened 2021-05-13
Bug 1960057: Fix potential mutex deadlock

msherif1234:
there is ordering issue in createNamespaceLocked() vs getNamespaceLocked(), the issue was flagged
while running test code using [deadlock](https://github.com/sasha-s/go-deadlock)

Attached a log while running with deadlock package and how it flagged the deadlock below for reference.

[test.log](https://github.com/openshift/ovn-kubernetes/files/6472632/test.log)
[instrumentation diffs to turn on deadlock mutex usage](https://paste.centos.org/view/833229e9)

### DIFF
--- a/go-controller/pkg/ovn/namespace.go
+++ b/go-controller/pkg/ovn/namespace.go
@@ -48,10 +48,12 @@ func (oc *Controller) syncNamespaces(namespaces []interface{}) {
 
 func (oc *Controller) addPodToNamespace(ns string, portInfo *lpInfo) error {
 	nsInfo, err := oc.waitForNamespaceLocked(ns)
+	if nsInfo != nil {
+		defer nsInfo.Unlock()
+	}
 	if err != nil {
 		return err
 	}
-	defer nsInfo.Unlock()
 
 	if nsInfo.addressSet == nil {
 		nsInfo.addressSet, err = oc.createNamespaceAddrSetAllPods(ns)
@@ -341,7 +343,6 @@ func (oc *Controller) getNamespaceLocked(ns string) *namespaceInfo {
 // with its mutex locked.
 func (oc *Controller) createNamespaceLocked(ns string) *namespaceInfo {
 	oc.namespacesMutex.Lock()
-	defer oc.namespacesMutex.Unlock()
 
 	nsInfo := &namespaceInfo{
 		networkPolicies:       make(map[string]*namespacePolicy),
@@ -349,9 +350,10 @@ func (oc *Controller) createNamespaceLocked(ns string) *namespaceInfo {
 		multicastEnabled:      false,
 		routingExternalPodGWs: make(map[string][]net.IP),
 	}
-	nsInfo.Lock()
 	oc.namespaces[ns] = nsInfo
+	oc.namespacesMutex.Unlock()
 
+	nsInfo.Lock()
 	return nsInfo
 }
 


go-controller/pkg/ovn/namespace.go
```diff
@@ -48,10 +48,12 @@ func (oc *Controller) syncNamespaces(namespaces []interface{}) {
 
 func (oc *Controller) addPodToNamespace(ns string, portInfo *lpInfo) error {
 	nsInfo, err := oc.waitForNamespaceLocked(ns)
+	if nsInfo != nil {
```
JacobTanenbaum, line 51: the way the function is written if there is an error the lock was not grabbed, the only time that nsInfo is nil is in the case of an error

go-controller/pkg/ovn/namespace.go
```diff
@@ -341,17 +343,17 @@ func (oc *Controller) getNamespaceLocked(ns string) *namespaceInfo {
...
 		podExternalRoutes:     make(map[string]map[string]string),
 		multicastEnabled:      false,
 		routingExternalPodGWs: make(map[string][]net.IP),
 	}
-	nsInfo.Lock()
 	oc.namespaces[ns] = nsInfo
+	oc.namespacesMutex.Unlock()
 
+	nsInfo.Lock()
 	return nsInfo
```
JacobTanenbaum, line 357: I don't understand what changing this locking is hoping to accomplish?

go-controller/pkg/ovn/namespace.go
msherif1234: when I ran UT I hit a panic with nil access hence fixed this occurance

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x16f5e57]

goroutine 9479 [running]:
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn.(*Controller).addPodToNamespace(0xc001099400, 0x1b38e22, 0xa, 0xc001054c80, 0x0, 0x0)
        /home/mmahmoud/go/src/ovn-kubernetes-1/go-controller/pkg/ovn/namespace.go:51 +0x77
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn.(*Controller).addLogicalPort(0xc001099400, 0xc0008a8400, 0x0, 0x0)
        /home/mmahmoud/go/src/ovn-kubernetes-1/go-controller/pkg/ovn/pods.go:438 +0xdd4
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn.(*Controller).ensurePod(0xc001099400, 0x0, 0xc0008a8400, 0x1, 0x0)
        /home/mmahmoud/go/src/ovn-kubernetes-1/go-controller/pkg/ovn/ovn.go:626 +0x35e
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn.(*Controller).WatchPods.func2(0x1b0f620, 0xc0008a8400)
        /home/mmahmoud/go/src/ovn-kubernetes-1/go-controller/pkg/ovn/ovn.go:659 +0x5c
k8s.io/client-go/tools/cache.ResourceEventHandlerFuncs.OnAdd(...)
        /home/mmahmoud/go/pkg/mod/k8s.io/client-go@v0.20.0-rc.0/tools/cache/controller.go:231
k8s.io/client-go/tools/cache.FilteringResourceEventHandler.OnAdd(0xc001218640, 0x1dc0df0, 0xc000470390, 0x1b0f620, 0xc0008a8400)
        /home/mmahmoud/go/pkg/mod/k8s.io/client-go@v0.20.0-rc.0/tools/cache/controller.go:264 +0x6a
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory.(*Handler).OnAdd(...)
        /home/mmahmoud/go/src/ovn-kubernetes-1/go-controller/pkg/factory/handler.go:38
```

go-controller/pkg/ovn/namespace.go
msherif1234: currently in set we do the following :-
lock namespaces
lock  nsinfo
unlock namespaces
and later unlock nsinfo

while we do the opposite when doing get this change attempt to avoid this out of order

go-controller/pkg/ovn/namespace.go
dcbw: @msherif1234 I agree with Jacob; if there is *any* way that `err != nil && nsInfo != nil` that would be a pretty big issue and we should make sure that never happens. Then we don't need this chunk?

PR #540, opened 2021-05-14
Bug 1958958: NewAddressSet: return nil in case of error

fedepaol:
newOvnAddressSets returns a concrete type, that gets assigned to an
interface. This makes the interface not being nil (even if the backing
value is).  In particular, having
nsInfo.addressSet, err = oc.createNamespaceAddrSetAllPods(ns)
called twice would result in having the interface not being nil but the
value behind it being nil and it will make sequent use of
nsInfo.addressSet refer to a nil value.


trozet (approved):
/lgtm

PR #545, opened 2021-05-19
Bug 1956955: Reduces number of OVN operations in services #2201

trozet:
Changes-Include:
- Implement a services cache to avoid calling remove vip so many times in OVN for different load balancers which may not have the vip on them.
- Fixes leaving around unidling VIPs in OVN accidentally.
- Fixes initial sync of ovnkube-master starting, where it would not repair/resync certain services. This would result in some services not working across the cluster on ovnkube-master restart.
- Batches OVN service transactions: Large scale service operations result in many ovn-nbctl commands. This is addressed by batching as many as possible into a single nbctl transaction.

### DIFF
--- a/go-controller/pkg/ovn/controller/services/repair.go
+++ b/go-controller/pkg/ovn/controller/services/repair.go
@@ -56,11 +56,11 @@ func (r *Repair) runOnce() error {
 	for _, p := range protocols {
 		lbUUID, err := loadbalancer.GetOVNKubeLoadBalancer(p)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


trozet:
@alexanderConstantinescu PTAL

alexanderConstantinescu (approved):
I have had a quick look at this, but I didn't have time to really dive in deeply. However, seeing as how:

- The upstream PRs received thorough review
- CI seems to be passing well
- The commits were cleanly cherry-picked 

/lgtm

PR #548, opened 2021-05-20
Bug 1962819: CNI cmdCheck: treat ingress_policing_rate=0 as not found

fedepaol:
When the cni is not chained with multus, the container runtime invokes
the check command. The check command checks the bandwith requirements
annotated on the pod, vs what's being configured in ovs. Ovs sets
ingress_policing_rate=0 by default, so the check always fails.

Here, when ingress_policing_rate is 0, we treat it as not found.

Manual cherry pick of https://github.com/openshift/ovn-kubernetes/pull/522


derekwaynecarr:

approved

PR #549, opened 2021-05-20
Bug 1959597: upgrade ovs to version 2.15 to support column diffs.

msherif1234:
Update 4.7 release to use OVS 2.15 so we can enable column diff feature starting from release 4.8
Note: after this commit is merged we no longer can downgrade to older releases.
New ovsdb-server will be able to read old database format, but older ovsdb-server will fail to read database created by this new OVS version.

changes in ovs2.15: https://github.com/openvswitch/ovs/blob/0b3ff31d35f56c9401d868613581b4d35bb42724/NEWS#L17
package details : https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=1584628

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN yum install -y  \
 	selinux-policy && \
 	yum clean all
 
-ARG ovsver=2.13.0-79.el8fdp
+ARG ovsver=2.15.0-15.el8fdp
 ARG ovnver=20.12.0-24.el8fdp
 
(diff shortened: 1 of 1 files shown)


squeed:
Because we upgraded 4.8 to OVS 2.15 via the normal development process, this will never have a 4.8 bug.

Thus, I'm manually adding the valid-bug label. @knobunc, feel free to yell at me for this.

squeed:
/lgtm cancel

some debate about if we want this.

PR #551, opened 2021-05-21
Bug 1962392: CARRY: fix missed learn for hybrid exgw

trozet:
We are intentionally ignoring learn when something critical to the learn
flow changes like pod ip, mac, exgw annotations. However, once we sync
flows we should reset the pod to learn the flow on the next sync.
However, we were only resetting pods for which we found flows already in
OVS. This fixes that by restting ignoreLearn for all pods as long as we
have a successful sync.

### DIFF
--- a/go-controller/hybrid-overlay/pkg/controller/node_linux.go
+++ b/go-controller/hybrid-overlay/pkg/controller/node_linux.go
@@ -610,7 +610,7 @@ func (n *NodeController) syncFlows() {
 			// and we accidentally pick up the old vtep flow and cache it. This should only ever happen on a pod update
 			// with an NS annotation VTEP change. We only need to ignore it for one iteration of sync.
 			if cacheEntry.ignoreLearn {
-				klog.V(5).Infof("Ignoring learned flow to add to hybrid cache for this iteration: %s", line)
+				klog.Infof("Ignoring learned flow to add to hybrid cache for this iteration: %s", line)
 				cacheEntry.ignoreLearn = false
 				cacheEntry.learnedFlow = ""
 				continue
@@ -635,6 +635,13 @@ func (n *NodeController) syncFlows() {
 	_, _, err = util.ReplaceOFFlows(extBridgeName, flows)
 	if err != nil {
 		klog.Errorf("Failed to add flows, error: %v, flows: %s", err, flows)
+		return
+	}
+
+	// handle resetting ignore learn as we may not have encountered some pods while
+	// iterating through the flows in OVS
+	for _, entry := range n.flowCache {
+		entry.ignoreLearn = false
 	}
 }
 


trozet:
@dcbw @fedepaol @dave-tucker PTAL

abhat:
LGTM.

go-controller/hybrid-overlay/pkg/controller/node_linux.go
```diff
@@ -635,6 +635,13 @@ func (n *NodeController) syncFlows() {
 	_, _, err = util.ReplaceOFFlows(extBridgeName, flows)
 	if err != nil {
 		klog.Errorf("Failed to add flows, error: %v, flows: %s", err, flows)
+		return
+	}
+
+	// handle resetting ignore learn as we may not have encountered some pods while
+	// iterating through the flows in OVS
+	for _, entry := range n.flowCache {
+		entry.ignoreLearn = false
```
trozet, line 644: I put this here to only reset it if we had a successful flow sync, but after thinking about this some more I dont think it really matters since we are handling pods that had no learn flow during the current sync. I think it should be safe to put it in the other loop on line 629. Thoughts?

go-controller/hybrid-overlay/pkg/controller/node_linux.go
```diff
@@ -635,6 +635,13 @@ func (n *NodeController) syncFlows() {
 	_, _, err = util.ReplaceOFFlows(extBridgeName, flows)
 	if err != nil {
 		klog.Errorf("Failed to add flows, error: %v, flows: %s", err, flows)
+		return
+	}
+
+	// handle resetting ignore learn as we may not have encountered some pods while
```
fedepaol, line 641: By doing this, it means that we need to wait for up to 30 seconds for the right flows to get in right?

Because what will happen is:
- the pod gets created -> sync with empty cache
- the pod gets the annotations from the namespace -> sync with filled cache, ignoreLearn = true, set to false
- the timer calls another syncFlows, where the learned flows are then added

go-controller/hybrid-overlay/pkg/controller/node_linux.go
fedepaol: I think they are syntactically equivalent. ignoreLearn will hit the next round

go-controller/hybrid-overlay/pkg/controller/node_linux.go
trozet: It could be up to 30 seconds before we get the learned flow in the cache, but that's not the same as the datapath having the flows:
1. pod gets created ->sync (we dont program any of the flows anyway because we dont have annotations yet)
2. pod update with annotations (we now program the flows to learn, and set ignoreLearn to false)
3. pod comes up and tries to arp, forces exgw to send arp, learn flow is created in OVS (pod can now send traffic)
4. sometime later (up to 30 seconds) ovnkube syncs again, and populates the learn flow in its cache so it wont be removed for this pod

fedepaol: Oh, now I see it. The difference is setting it to false in 2. 
Before we had `ignoreLearn = true` so in step 4 we would remove the learned flows.

fedepaol:
LGTM

dcbw:
Looks like Federico's hold didn't work, so until he's OK:
/hold

dcbw:
@fedepaol and I reviewed and chatted on slack
/hold cancel

dcbw:
One of the dual-stack failures is apparently https://github.com/openshift/machine-config-operator/pull/2580

dcbw:
ovn-ipv6 failure is pulling images from:

`[36mINFO[0m[2021-05-24T17:08:31Z] error: gather did not start for pod must-gather-cltw9: unable to pull image: ImagePullBackOff: Back-off pulling image "registry.redhat.io/openshift4/ose-must-gather:latest" `

/retest

dcbw:
ci/prow/4.8-upgrade-from-stable-4.7-e2e-aws-ovn-upgrade failing on capacity issues:
```
 INFO[2021-05-24T19:27:27Z] level=fatal msg=failed to fetch Cluster: failed to fetch dependency of "Cluster": failed to generate asset "Platform Quota Check": error(MissingQuota): ec2/L-0263D0A3 is not available in us-west-2 because the required number of resources (2) is more than the limit of 0, vpc/L-A4707A72 is not available in us-west-2 because the required number of resources (1) is more than the limit of 0, vpc/L-E79EC296 is not available in us-west-2 because the required number of resources (2) is more than the limit of 0, vpc/L-F678F1CE is not available in us-west-2 because the required number of resources (1) is more than the limit of 0, vpc/L-FE5A380F is not available in us-west-2 because the required number of resources (1) is more than the limit of 0 
```

dcbw:
https://bugzilla.redhat.com/show_bug.cgi?id=1960758

https://github.com/openshift/machine-config-operator/pull/2580

PR #554, opened 2021-05-25
Bug 1958972: Fixes adding nodes to use new Services controller

trozet:
Accidentally the old services/endpoints watcher code was being used to handle service sync for when a node is added/updated. During scale deployments we would notice this would create deadlock. This PR migrates the code to only use the new services controller and removes the legacy code.


trozet:
dualstack currently broken:
https://bugzilla.redhat.com/show_bug.cgi?id=1966602

dcbw:
dualstack currently broken:
https://bugzilla.redhat.com/show_bug.cgi?id=1966602

PR #559, opened 2021-06-01
Bug 1927047: Handling packet sizes greater than pod MTU

trozet:
With OVN-kubernetes we set the MTU of the pods to be 100 less than
physical network MTU. This becomes a problem when something outside of
the cluster tries to access a pod (local or shared gw mode) or via
service (shared gateway mode) with a packet larger than the pod MTU.
The packet will be dropped by OVS because OVS will not re-fragment
at the lower pod MTU when trying to send the packet to the pod.

This solves the problem by checking packet size for packets destined
towards OVN. If they are > pod MTU + 12 bytes (eth overhead) they are
forwarded to the kernel. The kernel has a route for pods via mp0
interface, which is a -100 byte MTU interface. This means the kernel
will automatically send ICMP needs frag back to the client.

For services, we add back the PREROUTING iptables rule to DNAT nodeport
towards the cluster IP service. This means incoming nodeport packets
that are too large are forwarded into the kernel, and then DNAT'ed to
cluster ip. After DNAT, kernel looks up the routing table and finds the
cluster IP route has an MTU of -100 on it, which triggers the kernel to
also send ICMP needs frag back to the client.


dcbw:
failing due to https://github.com/openshift/machine-config-operator/pull/2580

PR #560, opened 2021-06-01
Bug 1961852: policy: reduce lock contention on policy object

squeed:
When a policy selects many pods, there can be lock contention. This doesn't need to be the case - all the pod creation wants to do is update a hash.

So, instead, take a read lock and use a sync-safe map instead.

Signed-off-by: Casey Callendrello <cdc@redhat.com>

This is a cherry-pick of upstream PR https://github.com/ovn-org/ovn-kubernetes/pull/2228

### DIFF
--- a/go-controller/pkg/ovn/policy.go
+++ b/go-controller/pkg/ovn/policy.go
@@ -21,7 +21,10 @@ import (
 )
 
 type networkPolicy struct {
-	sync.Mutex
+	// RWMutex synchronizes operations on the policy.
+	// Operations that change local and peer pods take a RLock,
+	// whereas operations that affect the policy take a Lock.
+	sync.RWMutex
 	name            string
 	namespace       string
 	policyTypes     []knet.PolicyType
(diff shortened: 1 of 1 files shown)


squeed:
test failures seem like total flakes...
/retest

go-controller/pkg/ovn/policy.go
```diff
@@ -679,17 +687,17 @@ func (oc *Controller) handleLocalPodSelectorDelFunc(
...
-	if np.deleted {
+	// If we never saw this pod, short-circuit
+	if _, ok := np.localPods.LoadAndDelete(logicalPort); !ok {
 		return
 	}
 
-	if _, ok := np.localPods[logicalPort]; !ok {
+	np.RLock()
+	defer np.RUnlock()
+	if np.deleted {
```
msherif1234, line 697: what is the use now of the above deleted flag after the switch to sync map ?

astoycos: I had the same thought... but it accounts for the case where a namespace deletion event is received before the policy deletion event. In that case we have to clean everything up from the namespace code before the policy code executes. see https://github.com/ovn-org/ovn-kubernetes/pull/1665

astoycos: An unfortunate fact of our controller design..

squeed: we hold the read lock so we're sure the NP can't be deleted out from under us. The sync map is just for the list of pods.

astoycos:
/lgtm

Fair warning CI has been rough, `ci/prow/e2e-metal-ipi-ovn-dualstack` is permafailing

PR #530, opened 2021-05-10
Bug 1953680: policy: handle NetworkPolicies single "from" multiple "ipBlock"

danwinship:
We were handling

    ingress:
    - from:
      - ipBlock: ...
    - from:
      - ipBlock: ...

but not

    ingress:
    - from:
      - ipBlock: ...
      - ipBlock: ...

PR #562, opened 2021-06-02
Bug 1967132: policy: handle NetworkPolicies single "from" multiple "ipBlock"

danwinship:
backport of #530 ; we only handled NPs with multiple `ipBlock` entries if they were written in a particular way

/hold
Are there upgrade concerns here? Will this change leave stray ACLs around? Will they ever get cleaned up? (EDIT: Yes, Yes, No; fixed by second commit)

### DIFF
--- a/go-controller/pkg/ovn/gress_policy.go
+++ b/go-controller/pkg/ovn/gress_policy.go
@@ -282,8 +282,8 @@ func (gp *gressPolicy) localPodAddACL(portGroupName, portGroupUUID string) {
 		if len(gp.ipBlock) > 0 {
 			// Add ACL allow rule for IPBlock CIDR
 			cidrMatches = gp.getMatchFromIPBlock(lportMatch, l4Match)
-			for _, cidrMatch := range cidrMatches {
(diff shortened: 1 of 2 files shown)


trozet:
@danwinship it looks like this could potentially be a problem. We update the acl using:
```
func (gp *gressPolicy) modifyACLAllow(oldMatch, newMatch string) error {
	uuid, stderr, err := util.RunOVNNbctl("--data=bare", "--no-heading",
		"--columns=_uuid", "find", "ACL", oldMatch,
		fmt.Sprintf("external-ids:namespace=%s", gp.policyNamespace),
		fmt.Sprintf("external-ids:policy=%s", gp.policyName),
		fmt.Sprintf("external-ids:%s_num=%d", gp.policyType, gp.idx),
		fmt.Sprintf("external-ids:policy_type=%s", gp.policyType))

```

basically using the old match to find the old policy and update it...not great. We probably need a sync method on upgrade to search for all the buggy acls, then call modifyACLAllow and fix them with the right match criteria

danwinship:
/hold cancel
forgot to get back to this backport after we fixed the problem in master...

ecordell:
[patch-manager] :rocket: Approved for z-stream by score: 1.50

picked

PR #565, opened 2021-06-09
[DNM] Hack: fall back to a direct retrieval of annotations on failure

squeed:
There seems to be a problem where the Pod informer is somehow behind when starting up during upgrades. So, try and fall back.

As a debugging measure, dump stack when fallback succeeds.

/cc @trozet


squeed:
/hold
this is for playing around

PR #567, opened 2021-06-09
Bug 1956955: Batching: Fixes finding maximum bash arguments

trozet:
It turns out what the system reports for maximum arguments may not be
the real amount of arguments that are usable in a shell. Different
kernels calculate it differently, but the gist is that the only reliable
way to know is to test for it. This patch does an initial search based
on the given max args in the kernel to find the actual usable amount.

The search is only done one time during initialization, so it shouldn't
impact performance.

### DIFF
--- a/go-controller/pkg/util/ovs.go
+++ b/go-controller/pkg/util/ovs.go
@@ -103,11 +103,17 @@ func findMaxArgsUsable(estimatedMax int) int {
 	for i := range args {
 		args[i] = "a"
 	}
-	if estimatedMax <= minOSArgs {
-		return minOSArgs
+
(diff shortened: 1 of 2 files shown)


aojea:
this is not the one we want, right?

trozet:
> this is not the one we want, right?

@aojea right, pulling in the Adrian's fix as well

trozet:
windows job is failing because of aws resouce issue and failed waiting for the windows server to come up:

level=error msg=Error: error creating Route53 Hosted Zone: TooManyHostedZones: Limits Exceeded: MAX_HOSTED_ZONES_BY_OWNER - Cannot create more hosted zones.

PR #568, opened 2021-06-09
Bug 1970068: Fixes getNodeIPs assuming dualstack

trozet:
The function was assuming dualstack, which was causing "Failed to find
node ips for gateway" error to appear over 14k times in our scale
test runs downstream.


trozet:
@kedark3 fyi

alexanderConstantinescu (approved):
/lgtm

PR #571, opened 2021-06-11
[release-4.7] Bug 1970779: Remove getDefaultIfAddr and use getNetworkInterfaceIPAddresses

alexanderConstantinescu:
As seen in the associated bug report: https://bugzilla.redhat.com/show_bug.cgi?id=1970779#c3 , there can be situations where the egress IP related code for finding the default IP address can sometimes pick up the virtual IPs on bare-metal clusters. Hence back-port the following commit which uses `getNetworkInterfaceIPAddresses` instead and has dedicated code to avoid that: https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/util/net_linux.go#L370-L376

/assign @trozet 

**- How to verify it**

### DIFF
--- a/go-controller/pkg/node/gateway_init.go
+++ b/go-controller/pkg/node/gateway_init.go
@@ -178,11 +178,15 @@ func (n *OvnNode) initGateway(subnets []*net.IPNet, nodeAnnotator kube.Annotator
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


mfojtik:

approved

PR #573, opened 2021-06-14
Bug 1966833: BACKPORT Add a cluster-wide group with node ls-to-cluster-router ports.

astoycos:
Backport to fix multicast on 4.6

Commit 40a90f06644a removed the multicast deny port group and used
instead the cluster port group.  However, this breaks pod-to-pod
multicast when pods reside on different nodes.  That is because OVN ACLs
are applied on all logical switch ports, including logical switch ports
connected to router ports.

Hence, an ACL of the form "if ip.mcast then deny" applied on the
clusterPortGroup will drop all multicast traffic that would normally be
routed by the cluster router even when multicast is allowed for a
namespace.

Instead, add a new (smaller) cluster wide group that only contains the
node logical switch ports connected to the cluster router.  This allows us
to define two allow ACLs for multicast traffic from/to node switches to/from
cluster router, therefore not breaking the namespace multicast allow policy
if pods reside on different nodes.

Fixes: 40a90f06644a ("Migrate default deny multicast policy to port-group")
Signed-off-by: Dumitru Ceara <dceara@redhat.com>

### DIFF
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -27,7 +27,6 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/informer"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/metrics"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/ipallocator"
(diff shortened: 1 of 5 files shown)


dceara:
I only did a visual review and compiled the changes.  I didn't actually test if multicast works with this in 4.6 although it should.  The backport looks ok to me.

trozet (approved):
/lgtm

PR #584, opened 2021-06-23
Bug 1961506: Detect if OVS supports check pkt length

trozet:
In older kernels like RHEL 7, OVS does not support chk pkt length in
kernel datapath. This will cause all packets hitting these flows to go
to userspace, which is the majority of OVN destined ingress traffic into
the node. This will result in severe performance degradation. To avoid
this, detect and disable using these flows if the kernel datapath does
not support the action.

Note this will result in these nodes not being able to handle packets
larger than pod MTU and send ICMP frag needed, but the trade off vs.
performance impact is worth it.

(cherry picked from commit 7e832266167b518fdb9f6eba63623e7753e753f5)

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -229,6 +229,20 @@ func gatewayInitInternal(nodeName, gwIntf string, subnets []*net.IPNet, gwNextHo
 		return bridgeName, uplinkName, nil, nil, err
 	}
 
+	if !config.Gateway.DisablePacketMTUCheck {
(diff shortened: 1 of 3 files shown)


knobunc:

Skipping because we passed on mainline, and all of the recent runs of this test have failed due to resource problems.

We are working to see what needs to be done to fix, or skip, the tests.

knobunc:
/override ci/prow/4.8-upgrade-from-stable-4.7-e2e-aws-ovn-upgrade

PR #589, opened 2021-06-25
[wip] pod add latency test

dcbw:


dcbw:
/retest
azure and gcp had resource issues on install

PR #590, opened 2021-06-25
Bug 1973286: 6-25-2021 merge

trozet:


trozet:
@squeed @jcaamano FYI

squeed:
Failing tests:

- NetworkPolicy between server and client should enforce multiple egress policies with egress allow-all policy taking precedence
- console not available
- monitoring stack didn't roll out

Definitely worth looking in to.

PR #614, opened 2021-07-20
Add an optional second bridge for ex gateway traffic

fedepaol:
We optionally want to have the ex gateway traffic going through a bridge different from the default one. Here we add the bridge and we wire it to the GR via a different external switch.

This is the cherry pick of https://github.com/ovn-org/ovn-kubernetes/pull/2318


fedepaol:
superseeded by https://github.com/openshift/ovn-kubernetes/pull/618

PR #615, opened 2021-07-20
[wip] cni: create veth peer in hostns rather than moving

dcbw:


dcbw:
lots of `Step e2e-aws-ovn-*-ipi-conf failed after 10s.` which happens long before any OVN components run.
/retest

PR #628, opened 2021-07-26
Bug 1986443: Fix pod handler race downstream

trozet:
Fixes another case where update pod and adding pods to retry could race. Also a trivial fix to skipping the pod when it is not scheduled.


trozet:
@dcbw PTAL

PR #632, opened 2021-07-28
Bug 1986708: Add routes for pod: fail only after checking all the gw addresses / ips

fedepaol:
When adding routes to pod, we fail inside the inner loop returning an
error. What happens is that if the pod has two ip addresses, and the gw
is set only for the second address, the function will mistakenly return
an error.


dcbw:
Closing in preference to https://github.com/openshift/ovn-kubernetes/pull/774 which already merged.

PR #633, opened 2021-07-28
[WIP] Scale debug 2021-07-28

trozet:


go-controller/vendor/github.com/ebay/go-ovn/ovnimp.go
```diff
@@ -147,8 +150,12 @@ func (odbi *ovndb) getRowsMatchingUUID(table, field, uuid string) ([]string, err
 
 func (odbi *ovndb) transact(db string, ops ...libovsdb.Operation) ([]libovsdb.OperationResult, error) {
 	// Only support one trans at same time now.
-	odbi.tranmutex.Lock()
-	defer odbi.tranmutex.Unlock()
+	start := time.Now()
+	defer func() {
+		klog.Infof("ovsdb transaction took %v", time.Since(start))
+	}()
+	//odbi.tranmutex.Lock()
```
abhat, line 157: this will violate cache consistency esp with multiple handlers using the single go-ovn client instance. ok for debug purposes I suppose.

dcbw: Note that latest libovsdb itself has an "rpc mutex" that protects its internal Transact() call. So just removing from go-ovn won't save us when we revendor soon...

rpc2 grabs a mutex around writes to the socket, but with this change (or removing the rpc lock in new libovsdb) we won't be guaranteed that the reply we receive is for the right transaciton that's in the pending queue in rpc2, right? Is that a problem?

trozet: we are guaranteed that the update for the cache happens for the transaction in question before the execute finishes.  This mutex doesnt do anything except block multiple transactions in flight, which rpc2 already serializes. If we have 2 threads T1 and T2, that execute transactions A and B, respectively and simultaneously:

T1.execute cannot return until A is complete, and the cache has been updated with A outcome
T2.execute cannot return until B is complete, and the cache has been updated with B outcome

However, there is no guarantee that when T2 execute completes that the cache is up to date with whatever T1 did, and vice versa.

PR #640, opened 2021-07-30
[wip] Scale fixes downstream

dcbw:
test some stuff out


trozet:
There are a bunch of sandbox failures here, which are most likely related to lowering the number of pod handlers, attempting to test 15 (previous default) and 50:
https://github.com/openshift/ovn-kubernetes/pull/644
https://github.com/openshift/ovn-kubernetes/pull/645

PR #647, opened 2021-08-06
Bug 1976399: DBChecker: reconcile the election timeouts when specified

fedepaol:
We may want to update the ovn election timers, and in order to do that
we need to act asynchronously because the election must be completed and
the change must be performed on the leader node. So, we add two optional
parameters for the desired election timer and we change that as part of
the db checker timer.


trozet:
/hold

I think we need https://github.com/ovn-org/ovn-kubernetes/pull/2447

fedepaol:
> /hold
> 
> I think we need [ovn-org/ovn-kubernetes#2447](https://github.com/ovn-org/ovn-kubernetes/pull/2447)

Indeed! Thanks for spotting that

dcbw:
/test e2e-ovn-hybrid-step-registry
terraform setup problems

PR #653, opened 2021-08-11
WIP: Scale drop 20210811

abhat:
WIP: Scale fixes from upstream and PR#2408 and PR#2385


dcbw:
I think this is quite likely out-of-date :)

PR #661, opened 2021-08-17
[release 4.8] Bug 1994624: Infer subnet for node /128 IPv6 addresses

jcaamano:
When node addresses are allocated through DHCPv6 IA_NA, they might get
a /128 prefix address. These node addresses are set on the gateway
router port as well. OVN routers infers on-link routes from subnet
information on the address itself. Since /128 addresses are not on a
subnet, cluster traffic is routed through the node gatway instead of
directly.

Note that OVN does not support learning routes from ICMPv6 route
advertisement and does not support adding on-link static routes.

This patch infers the subnet of the interface looking at the node
routing information and replaces the /128 prefix with the subnet
prefix.

(cherry picked from commit 1db8f65a29d7e917de4007a613dd1c21c6f10c19)


jcaamano:
@dcbw PTAL, thanks!

dhellmann:
[patch-manager] :rocket: Approved for z-stream by score: 1.50

picked

PR #666, opened 2021-08-18
Bug 1995330: Cherry-pick of per-service loadbalancers

squeed:
This is a cherry-pick of https://github.com/ovn-org/ovn-kubernetes/pull/2294

This removes all of the complexity around managing vip membership.
Instead, it creates one or more load balancers per service.
Additionally, it tags those load balancers with the GroupKind and
NamespacedName of the "owning" Kubernetes object, so we can easily
find stale objects and reconcile.

It includes an idempotent load-balance reconciliation engine, so we can
generate just the ideal state in the ServiceController and let the
reconciler take care of the details. This includes a read-through cache,
to minimize unnecessary reads from the database.

A much-simpler repair engine handles cleaning up any stale Load_Balancer
objects on startup.

This also implements session affinity, which is just an option on the
Load_Balancer object.


go-controller/pkg/ovn/loadbalancer/loadbalancer.go
```diff
@@ -173,6 +173,8 @@ func lbToColumns(lb *LB) []string {
 	}
 
 	skipSNAT := "false"
+	// HACK(cdc)
+	// re-enable when BZ 1995326 is fixed
```
dcbw, line 177: @squeed how does this disable the option?

go-controller/pkg/ovn/master.go
```diff
@@ -598,10 +598,7 @@ func (oc *Controller) syncGatewayLogicalNetwork(node *kapi.Node, l3GatewayConfig
 	// See https://github.com/openshift/ovn-kubernetes/pull/281
 	drLRPIPs, _ := oc.joinSwIPManager.getJoinLRPCacheIPs(types.OVNClusterRouter)
 	if l3GatewayConfig.Mode == config.GatewayModeLocal {
-		err = gatewayInitMinimal(node.Name, l3GatewayConfig, oc.SCTPSupport)
-		if err != nil {
-			return fmt.Errorf("failed to init local gateway with no OVS bridge: %v", err)
-		}
+		// do nothing, now that we have the services controller.
```
trozet, line 601: i think you can just remove the entire if statement now right? This was only for 4.5->4.6 and i think SDN->OVN migration. It's not applicable anymore as SDN->OVN migration uses a different method and doesnt rely on using "none" for the bridge name.

@pliurh can correct me if I'm wrong.

go-controller/pkg/ovn/loadbalancer/loadbalancer.go
squeed: it doesnt :facepalm:.

go-controller/pkg/ovn/master.go
squeed: Oh interesting - I wonder if I can just revert #281 entirely?

squeed: If so I'll file a separate PR first.

pliurh: Yeah, the SDN migration doesn't rely on #281 anymore.

trozet:
dualstack is a known failure

trozet:
Latest local gw job failures have nothing to do with this pr

dcbw:
/override ci/prow/okd-images
The images exist for OCP since prow/images worked; apparently OKD mirrors are borked.

squeed:
Many many thanks, @trozet @dcbw for pushing this over the line :heart:

PR #671, opened 2021-08-19
Bug 1943334: Taint node with NoSchedule when ovnkube pod is down

tssurya:
UseCase: During upgrades, when the ovnkube pod is deleted and
before/while it gets recreated if a new pod request comes, it
will fail.

When node networking is down, we should not allow pods to be created
on that node. This PR fixes this by adding a taint on the node with
the NoSchedule effect when the ovnkube pod receives a SIGTERM and then
removes the taint when the ovnkube pod finishes setting up networking
and initializing the CNI config.

(cherry picked from commit 7e958e0aa8f3ef3816bd9a1f6ac56a3082598807)


dcbw:
/retest-required
AWS failing with terraform LB setup issues

PR #676, opened 2021-08-23
Bug 1988512: [release-4.6] Fix LGW externalIP

astoycos:
Make Iptables rule for every ExternalIP or ingress IP, this will
result in shortcirciting of node -> ExternalIP service traffic

Remove code to manage host routes

Ensure we cleanup any stale routes on upgrade

Signed-off-by: astoycos <astoycos@redhat.com>


hardys:
[patch-manager] Approving for cherry pick with a score of 0.70

PR #686, opened 2021-08-24
Bug 1996201: Fixes cases of timed out while waiting for OVS port binding

trozet:
1. Add back checking for pod flows (to ensure the port that gets ovn_installed is the latest port for the pod)
2. Cancel oldest sandbox request if the pod's MAC or UUID changes. Basically if ovnkube is serving a CNI request and the mac was changed due to a pod delete/add event, cancel this request and move onto the newer pod.


jluhrsen:
/retest

the gcp-ovn-upgrade job failures look a lot like what we see in the periodic job
the metal jobs have failures that we see in the periodic jobs, but those should be able to pass once in a while
not sure about the hybird-step-registry job. it looks like it fails 50% of the time, although I didn't dig in to why
each job is failing.

dcbw:
/lgtm
retest

mfojtik:

adding cherry-pick approved

PR #698, opened 2021-08-30
Bug 1998146: Fix lb delete during node deletion

tssurya:
When a node gets deleted, the logical switch and
gateway router get removed first. However we do
not update the LBCache and this causes dependency
failures during deletion leaving behind wrong ip->vip
pairing in the lbs.

When the pods on this node go away, ensureLBs
gets triggered and it tries to update endpoints across
load balancers. This set load_balancer command is
batched with ls-lb-del & ls-lr-del commands to remove
association of the lbs from the removed switches and
routers. These commands fail together because the switches
and routers are not present in ovn.

This PR updates the LBcache as soon as we remove the
routers and switches.

(cherry picked from commit 6de80e8dbb2c865209d6dffd32cd4d93eb46b77e)

### DIFF
--- a/go-controller/pkg/ovn/gateway_cleanup.go
+++ b/go-controller/pkg/ovn/gateway_cleanup.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


squeed:
https://bugzilla.redhat.com/show_bug.cgi?id=2000086 - permafailing

PR #609, opened 2021-07-16
Merge 2021-07-18

trozet:
@fedepaol @jcaamano @astoycos @abhat @alexanderConstantinescu @dcbw FYI

PR #700, opened 2021-08-31
[release-4.8] Bug 1999638: Fix duplicate incrementing of subnet allocation metric

abhat:
When the master is restarted, all the nodes are fed using
both a fetch of existing nodes and by the add node event.
In case the node already had subnet annotations, we were
updating the subnets allocated metric again. This commit
checks if new subnets were allocated and only then calls
the code to update metrics associated with allocated
subnets.

(cherry picked from commit 4de9fa6a60e4f103939a880f1a0c5a6beb5d8afe)

### DIFF
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -1074,14 +1074,15 @@ func (oc *Controller) addNode(node *kapi.Node) ([]*net.IPNet, error) {
 	// delete stale chassis in SBDB if any
 	oc.deleteStaleNodeChassis(node)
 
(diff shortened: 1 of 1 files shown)


mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 1.00

adding cherry-pick approved

mfojtik:
taking CPA back as the jobs seems to be permanently failing

hardys:
[patch-manager] Approving cherry-pick with score 1.0, high prio bz, I see the release-4.9 change landed via #609

PR #701, opened 2021-08-31
[wip] Scale fixes3 leader

dcbw:


go-controller/pkg/ovn/pods.go
```diff
@@ -239,6 +239,15 @@ func (oc *Controller) addRoutesGatewayIP(pod *kapi.Pod, podAnnotation *util.PodA
 	return nil
 }
 
+func combineOvnCmdOps(dst, src *goovn.OvnCommand) {
```
abhat, line 257: Isn't it better to just add another `LSPAddObject` api? This feels hack-ish.

PR #709, opened 2021-08-31
[WIP] Monitor cancel/ensure handling

trozet:

### DIFF
--- a/go-controller/vendor/github.com/ebay/go-ovn/client.go
+++ b/go-controller/vendor/github.com/ebay/go-ovn/client.go
@@ -289,6 +289,7 @@ type ovndb struct {
 	tableCols    map[string][]string
 	tlsConfig    *tls.Config
 	reconn       bool
+	monitorSem   uint32
 }
 
 func connect(c *ovndb) (err error) {
@@ -357,6 +358,8 @@ func NewClient(cfg *Config) (Client, error) {
 }
 
 func (c *ovndb) reconnect() {
+	c.tranmutex.Lock()
+	defer c.tranmutex.Unlock()
 	ticker := time.NewTicker(500 * time.Millisecond)
 	go func() {
 		log.Printf("%s disconnected. Reconnecting ... \n", c.addr)
--- a/go-controller/vendor/github.com/ebay/go-ovn/client.go
+++ b/go-controller/vendor/github.com/ebay/go-ovn/client.go
@@ -289,7 +289,6 @@ type ovndb struct {
 	tableCols    map[string][]string
 	tlsConfig    *tls.Config
 	reconn       bool
-	monitorSem   uint32
 }
 
 func connect(c *ovndb) (err error) {
--- a/go-controller/vendor/github.com/ebay/go-ovn/ovnimp.go
+++ b/go-controller/vendor/github.com/ebay/go-ovn/ovnimp.go
@@ -19,8 +19,10 @@ package goovn
 import (
 	"errors"
 	"fmt"
+	"log"
 	"reflect"
 	"strings"
+	"sync/atomic"
 
 	"github.com/ebay/libovsdb"
 )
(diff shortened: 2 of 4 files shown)


trozet:
@tssurya

go-controller/vendor/github.com/ebay/go-ovn/client.go
```diff
@@ -289,6 +289,7 @@ type ovndb struct {
 	tableCols    map[string][]string
 	tlsConfig    *tls.Config
 	reconn       bool
+	monitorSem   *uint32
```
dcbw, line 292: @trozet stylistically this should be a plain `uint32` and passed as a pointer to `atomic.*Uint32(&monitorSet)`; but Go probably doesn't care that it's defined as a pointer not an actual int? Maybe?

go-controller/vendor/github.com/ebay/go-ovn/ovnimp.go
```diff
@@ -174,6 +176,10 @@ func (odbi *ovndb) transact(db string, ops ...libovsdb.Operation) ([]libovsdb.Op
 	if len(reply) < len(ops) {
 		return reply, fmt.Errorf("Number of Replies should be atleast equal to number of operations")
 	}
+	if atomic.LoadUint32(odbi.monitorSem) != 0 {
+		// we did not receive an Update for this transaction, disconnect
+		odbi.reconnect()
```
dcbw, line 181: Should really have a log message here so we know how often this is a problem (or so we know when our assumption about Request-Update-Response ordering isn't true).

dcbw: @trozet also, this doesn't disconnect/close the original client. I think you want to call `odbi.Close()` instead, which will call libovsdb.Client's Disconnect() which in turn will signal the disconnect handler in go-ovn and then call reconnect(). Otherwise you're leaking a libovsdb.Client that's still connected to the DB right?

go-controller/vendor/github.com/ebay/go-ovn/client.go
trozet: no you are right, should have done a plain uint32. updating

go-controller/vendor/github.com/ebay/go-ovn/client.go
```diff
@@ -357,6 +357,8 @@ func NewClient(cfg *Config) (Client, error) {
 }
 
 func (c *ovndb) reconnect() {
+	c.tranmutex.Lock()
```
tssurya, line 360: q: why do we need this mutex? We are not calling the `reconnect` function from anywhere right?.

```
odbi.Close() 
```
will go to
```
// TODO return proper error
func (c *ovndb) Close() error {
	c.client.Disconnect()
	return nil
}
```
will go to
```
// Disconnect will close the OVSDB connection
func (ovs OvsdbClient) Disconnect() {
	ovs.rpcClient.Close()
}
```
will go to

```
// Close waits for active calls to finish and closes the codec.
func (c *Client) Close() error {
	c.mutex.Lock()
	if c.shutdown || c.closing {
		c.mutex.Unlock()
		return ErrShutdown
	}
	c.closing = true
	c.mutex.Unlock()
	return c.codec.Close()
}
```
will go to
```
func (c *gobCodec) Close() error {
	return c.rwc.Close()
}
```
and here
```
type Closer interface {
	Close() error
}
```
after this point, I can't follow the code flow to `odbi.reconnect` function.

tssurya: I am guessing somehow it triggers
```
func (notify ovnNotifier) Disconnected(client *libovsdb.OvsdbClient) {
	if notify.odbi.reconn {
		notify.odbi.reconnect()
	} else if notify.odbi.disconnectCB != nil {
		notify.odbi.disconnectCB()
	}
}
```

PR #717, opened 2021-09-03
Bug 1998614: Ensure client handling of canceled/dropped OVSDB monitor

tssurya:
Under heavy load, monitor gets dropped silently and we stop receiving `update` messages from ovsdbserver. This leads to stale cache and inconsistency with what's in the DB leading to transaction errors. While this is originally a ovsdbserver bug, a workwround fix here would be to trigger reconnect to recreate the monitor and start receiving update messages, when a transaction error occurs.

Co-Authored-By: trozet@redhat.com

### DIFF
--- a/go-controller/vendor/github.com/ebay/go-ovn/client.go
+++ b/go-controller/vendor/github.com/ebay/go-ovn/client.go
@@ -281,6 +281,7 @@ type ovndb struct {
 	client       *libovsdb.OvsdbClient
 	cache        map[string]map[string]libovsdb.Row
 	cachemutex   sync.RWMutex
+	tranmutex    sync.RWMutex
 	signalCB     OVNSignal
 	disconnectCB OVNDisconnectedCallback
 	db           string
@@ -359,6 +360,8 @@ func (c *ovndb) reconnect() {
 	ticker := time.NewTicker(500 * time.Millisecond)
 	go func() {
 		log.Printf("%s disconnected. Reconnecting ... \n", c.addr)
+		c.tranmutex.Lock()
+		defer c.tranmutex.Unlock()
 		retry := 0
 		for range ticker.C {
 			if err := connect(c); err != nil {
@@ -391,7 +394,7 @@ func (c *ovndb) filterTablesFromSchema() []string {
 		tables = SBTablesOrder
 	}
 
-	dbSchema := c.GetSchema()
+	dbSchema := c.getSchema(c.db)
 	schemaTables := make([]string, 0)
 	for _, table := range tables {
 		if _, ok := dbSchema.Tables[table]; ok {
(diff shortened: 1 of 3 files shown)


go-controller/vendor/github.com/ebay/go-ovn/client.go
```diff
@@ -281,6 +281,7 @@ type ovndb struct {
 	client       *libovsdb.OvsdbClient
 	cache        map[string]map[string]libovsdb.Row
 	cachemutex   sync.RWMutex
+	tranmutex    sync.Mutex
```
trozet, line 284: use an RWMutex, and call Lock like you are in 362 thats good. But you need to RLock in transact. If you dont then the mutex isnt really doing anything. The purpose of adding the mutex back here is so that we avoid issuing a transaction while the client could be nil:
https://github.com/openshift/ovn-kubernetes/pull/717/files

That would cause a panic and crash ovnkube. On the other hand if the client is not nil, but closed, that will simply return an error "error: connection is shut down"

go-controller/vendor/github.com/ebay/go-ovn/client.go
```diff
@@ -358,6 +359,8 @@ func NewClient(cfg *Config) (Client, error) {
 func (c *ovndb) reconnect() {
 	ticker := time.NewTicker(500 * time.Millisecond)
 	go func() {
+		c.tranmutex.Lock()
```
trozet, line 362: move this under line 364 so we get the log message before we lock the mutex

go-controller/vendor/github.com/ebay/go-ovn/ovnimp.go
```diff
@@ -164,7 +164,8 @@ func (odbi *ovndb) transact(db string, ops ...libovsdb.Operation) ([]libovsdb.Op
 			if i < len(ops) {
 				opsInfo = fmt.Sprintf("%v", ops[i])
 			}
-			return nil, fmt.Errorf("Transaction Failed due to an error: %v details: %v in %s",
+			odbi.Close()
```
trozet, line 167: add 
```
c.tranMutex.RLock() 
defer c.tranMutexRUnlock()
```
at the start of this transact function

trozet (changes requested):

trozet:
you also need to pull in my commit for monitor_cancel support:
https://github.com/eBay/libovsdb/pull/41

trozet:
/retitle Bug 1998614: Ensure client handling of canceled/dropped OVSDB monitor

go-controller/vendor/github.com/ebay/go-ovn/client.go
trozet: https://github.com/openshift/ovn-kubernetes/blob/41ad158235fcb290bab580a3bade8b521c732377/go-controller/vendor/github.com/ebay/go-ovn/client.go#L302

go-controller/vendor/github.com/ebay/go-ovn/ovnimp.go
dcbw: @trozet FWIW Surya and I decided we do not need to protect any other uses of `c.client` (GetSchema(), Close()) in 4.7.

But @tssurya since this PR is for master/4.9, we do need to protect Close() (because updateSslKeyPair() calls Close()) and we also need to split GetSchema() into a locked and unlocked variant because deleteChassis() calls it.

trozet: good catch. We can backport this protection as well for Close and GetSchema just in case. So Close can use tranMutex.Lock() and tranMutex.RUnlock() should be used in GetSchema(). Like dcbw suggested we need internal/external function because Transact will already have the lock, so it can just call ```c.client.Disconnect()```

go-controller/vendor/github.com/ebay/go-ovn/client.go
tssurya: ack. done.

go-controller/vendor/github.com/ebay/go-ovn/client.go
tssurya: done.

tssurya:
> you also need to pull in my commit for monitor_cancel support:
> [eBay/libovsdb#41](https://github.com/eBay/libovsdb/pull/41)

done!

trozet:
/lgtm

thanks @tssurya !

PR #718, opened 2021-09-05
Bug 2001363: [4.8z] Ensure client handling of canceled/dropped OVSDB monitor

trozet:
Under heavy load, monitor gets dropped silently and we stop receiving update messages from ovsdbserver. This leads to stale cache and inconsistency with what's in the DB leading to transaction errors. While this is originally a ovsdbserver bug, a workwround fix here would be to trigger reconnect to recreate the monitor and start receiving update messages, when a transaction error occurs.


mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 2.50

adding cherry-pick approved

PR #724, opened 2021-09-07
[release-4.7] Bug 2001998: egressfirewall not set after upgrade

ricky-rav:
(backport of downstream PR 716 from master branch)

Fixes #2001641, which is a clone of #1997438, where any update to an EgressFirewall (including a cluster upgrade) causes an error as soon as more than one ACL UUID are retrieved and parsed, leading to the EgressFirewall eventually not being correctly set. This is easily reproducible in local gateway mode.

This patch proposes the following changes:

- Added `--format=table` to all `ovn-nbctl find` calls in order to have entries separated by a newline character and not by two newline characters, which is the default if `--format` is not specified;
- Replaced calls to `Strings.Split` on `sep=\n` with a safer `Strings.Fields`, which splits based on whitespace characters and trims them too.

    (cherry picked from commit 0106ec77896e7d492a737e790b635e52fe72b57c)
    (cherry picked from commit d6127e143618224edec2b036fbd2e1605e3e8fb6)
    (cherry picked from commit 33bc9b0eba880b54b16c94081bb65be22f5b8b48)

### DIFF
--- a/go-controller/pkg/ovn/egressfirewall.go
+++ b/go-controller/pkg/ovn/egressfirewall.go
@@ -434,7 +434,7 @@ func (oc *Controller) deleteEgressFirewallRules(externalID string) error {
 	} else {
 		logicalSwitches = []string{types.OVNJoinSwitch}
 	}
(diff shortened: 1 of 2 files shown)


hardys:
[patch-manager] Approving cherry pick, high severity score 2.0

PR #741, opened 2021-09-14
Bug 2000057: panic after EgressFirewall deletion and DNS record expiration

ricky-rav:
(backport of upstream https://github.com/ovn-org/ovn-kubernetes/pull/2471)

- Add a new channel named "deleted" to the EgressDNS struct, so that the domain name (if any) of a newly deleted EgressFirewall can be passed over to the Run() loop, which keeps track of DNS records and updates them as soon as they expire. This prevents inconsistencies between current EgressFirewalls and corresponding DNS records, which eventually led to a panic in a customer case.

- The aforementioned panic is now avoided by checking for the presence of the given DNS domain to delete in a map, before proceeding to update its (pointer) value.
- Fix type assertion when deleting EgressFirewalls, which resulted in DNS records never being deleted from the Run() loop.
(- Make error messages start with lowercase letters to comply with style guidelines, as a CI test was failing)

This patch fixes #2000057.

Signed-off-by: Riccardo Ravaioli <rravaiol@redhat.com>
(cherry picked from commit 12ff248bd2878f98bfae0dae5e207352d7def96b)

PR #749, opened 2021-09-15
[release-4.8] Bug 2004488: panic after EgressFirewall deletion and DNS record expiration

ricky-rav:
(backport from https://github.com/openshift/ovn-kubernetes/pull/741)

- Added a new channel named "deleted" to the EgressDNS struct, so tha…t the domain name (if any) of a newly deleted EgressFirewall can be passed over to the Run() loop, which keeps track of DNS records and updates them as soon as they expire. This prevents inconsistencies between current EgressFirewalls and corresponding DNS records, which eventually led to a panic in a customer case.

- The aforementioned panic is now avoided by checking for the presence of the given DNS domain to delete in a map, before proceeding to update its (pointer) value.

(- Made error messages start with lowercase letters to comply with style guidelines, as a CI test was failing)

This patch fixes bug BZ 2004488.

(cherry picked from commit 12ff248bd2878f98bfae0dae5e207352d7def96b)

### DIFF
--- a/go-controller/pkg/ovn/egressfirewall_dns.go
+++ b/go-controller/pkg/ovn/egressfirewall_dns.go
@@ -25,6 +25,7 @@ type EgressDNS struct {
 
 	// Report change when Add operation is done
 	added          chan struct{}
+	deleted        chan string
 	stopChan       chan struct{}
 	controllerStop <-chan struct{}
 }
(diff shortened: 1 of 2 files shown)


hardys:
[patch-manager] Approving for cherry-pick with a score of 2.50 - urgent bz

PR #753, opened 2021-09-17
Bug 2005357: [4.8z] Fixes misuse of pod annotations during update event

trozet:
In the update pod logic, we pass the current pod event to
addLogicalPort. In addLogicalPort we assume that if the annotations
exist for the pod mac/ifaddr, then we use those and do not update
annotations on the pod. This assumption is invalid, because this event
may not be the current state of the pod. In other words we could have a
situation where:

1. A pod add event comes we annotate with 10.0.0.2, assume OVN execute
   failure
2. Before the annotate is done, the pod is modified in some other way
   signaling another pod update event
3. A pod update event comes for 2, the pod is annotated with 10.0.0.3
   because this was an update to the original pod, before it was
   annotated with 10.0.0.2, assume OVN execute failure
4. A pod update event comes for 1, since annotations existed, nothing is
   annotated and 10.0.0.2 is found to be used. OVN logical port is
   configured with 10.0.0.2. addLogicalPort succeeds.
5. Now the pod has 10.0.0.3 annotated, and 10.0.0.2 in OVN. CNI openflow
   check will fail and the pod will never come up.

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/ovn.go
+++ b/go-controller/pkg/ovn/ovn.go
@@ -518,6 +518,22 @@ func (oc *Controller) WatchPods() {
 		UpdateFunc: func(old, newer interface{}) {
 			oldPod := old.(*kapi.Pod)
 			pod := newer.(*kapi.Pod)
+
+			// there may be a situation where this update event is not the latest
+			// and we rely on annotations to determine the pod mac/ifaddr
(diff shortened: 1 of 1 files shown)


hardys:
[patch-manager] Approving for cherry-pick with a score of 1.70

PR #766, opened 2021-09-21
Bug 2005519: [4.7z] Remove waiting for namespace and namespace lock contention

trozet:
Removes waiting for namespace event to show up. Now it is just ensured when needed. Also changes nsinfo to use a RWMutex so that pod handlers can all simultaneously RLock and not be blocked trying to add many pods for a single namespace.


dcbw:
/retest

PR #769, opened 2021-09-24
Bug 2007665: [4.7z] Ensure host interfaces are deleted by CNI

trozet:


go-controller/pkg/cni/cni.go
```diff
@@ -213,6 +213,12 @@ func HandleCNIRequest(request *PodRequest, podLister corev1listers.PodLister) ([
 func (pr *PodRequest) getCNIResult(podInterfaceInfo *PodInterfaceInfo) (*current.Result, error) {
 	interfacesArray, err := pr.ConfigureInterface(pr.PodNamespace, pr.PodName, podInterfaceInfo)
 	if err != nil {
+		for _, iface := range interfacesArray {
```
trozet, line 216: @dcbw fyi i had to add/rework this a little bit due to some missing patches in 4.7z compared to other branches

dcbw: @trozet I don't think this is right; if there are errors in ConfigureInterface() after the veth (or sriov) has been created, then the ports won't get deleted. ConfigureInterface() is designed to only return stuff in interfacesArray on success.

Why not just unconditonally delete ports on error?

trozet: oh crap, I forgot to make it return the interfaces even on error. If its fine to unconditionally do it, then thats even easier. Will make that change.

dcbw: Worst that happens is we print some errors in the logs.

dhellmann:
[patch-manager] :rocket: Approved for z-stream by score: 2.00

picked

PR #750, opened 2021-09-15
Bug 1959352: phase 2 scale improvements

trozet:
Changes Include:

1. batch pod operations into a single operation
2. remove unnecessary locking around address sets to avoid lock contention during pod operations
3. combine address set operations for add/remove IP into logical port OVN transactions
4. use uuids for OVSDB operations involving conditions like "where x == y"
5. update 2/3 json rpc methods for ovsdb
6. leader only connections for go-ovn
7. temporarily revert libovsdb clients until they have support for 5 and 6
8. fix an issue where reconnect in go-ovn client can hang forever if ovsdb server does not respond with a reply

PR #774, opened 2021-09-29
Bug 1986708: Add routes for pod: fail only after checking all the gw addresses / ips

trozet:
When adding routes to pod, we fail inside the inner loop returning an
error. What happens is that if the pod has two ip addresses, and the gw
is set only for the second address, the function will mistakenly return
an error.

Signed-off-by: Federico Paolinelli <fpaoline@redhat.com>
(cherry picked from commit d2e05937f75c067bee3de3b2cdbea9a554116da2)


dcbw:
/retest

dcbw:
https://bugzilla.redhat.com/show_bug.cgi?id=1970331

dcbw:
/override ci/prow/e2e-vsphere-ovn
sandbox failures are due to lack of go-ovn following leader which will get fixed when https://github.com/openshift/ovn-kubernetes/pull/750 gets backported to 4.9

PR #778, opened 2021-10-01
Bug 1997072: [4.9] phase 2 scale improvements

dcbw:
Cherry-pick of https://github.com/openshift/ovn-kubernetes/pull/750

Changes Include:

1. batch pod operations into a single operation
2. remove unnecessary locking around address sets to avoid lock contention during pod operations
3. combine address set operations for add/remove IP into logical port OVN transactions
4. use uuids for OVSDB operations involving conditions like "where x == y"
5. update 2/3 json rpc methods for ovsdb
6. leader only connections for go-ovn
7. temporarily revert libovsdb clients until they have support for 5 and 6
8. fix an issue where reconnect in go-ovn client can hang forever if ovsdb server does not respond with a reply

Also pulls in https://github.com/openshift/ovn-kubernetes/pull/781 and https://github.com/openshift/ovn-kubernetes/pull/787

Note that we might want https://github.com/openshift/ovn-kubernetes/pull/777 to land first, or we suck that into this PR; it makes the rebase/cherry-pick easier but isn't necessary. It should certainly be fixed in 4.9 regardless.

PR #781, opened 2021-10-04
pods: remove unnecessary LSPGet() calls

dcbw:
LSPGet() already returns the LSP's Options so just copy them
instead of calling LSPGetOptions() which has to lock the go-ovn
cache and look up the LSP.

Second, change the after-OVN-execute LSPGet() to LSPGetUUID()
after reading the UUID from the OVN execute response. go-ovn
uses the UUID as a cache index, so LSPGetUUID() is much faster
than getting an LSP by name which requires walking the cache
row-by-row.

Third, eliminate the LSPGet() that GetPortAddresses() does.
We already know the addresses from the first LSPGet() so we
only need to parse them. If the port didn't exist in the
first place, we know there are no addresses to parse so
don't even bother.

Finally, don't repeatedly access two levels of maps to get the
cache row when building a LSP object. Micro-benchmarks
indicate that's about 2x as slow as just getting the row once
and reading fields out of it.

@trozet

### DIFF
--- a/go-controller/pkg/ovn/pods.go
+++ b/go-controller/pkg/ovn/pods.go
@@ -295,28 +295,33 @@ func (oc *Controller) addLogicalPort(pod *kapi.Pod) (err error) {
 	var cmd *goovn.OvnCommand
 	var podCmd *goovn.OvnCommand
 	var releaseIPs bool
-	needsIP := true
+
+	opts := make(map[string]string)
 
 	// Check if the pod's logical switch port already exists. If it
 	// does don't re-add the port to OVN as this will change its
 	// UUID and and the port cache, address sets, and port groups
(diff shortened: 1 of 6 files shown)


trozet:
/lgtm

we should include it in https://github.com/openshift/ovn-kubernetes/pull/778

dcbw:
/test e2e-aws-ovn-local-gateway
`alert etcdGRPCRequestsSlow fired for 120 seconds with labels: {endpoint="etcd-metrics", grpc_method="Status", grpc_service="etcdserverpb.Maintenance", instance="10.0.180.119:9979", job="etcd", namespace="openshift-etcd", pod="etcd-ip-10-0-180-119.ec2.internal", service="etcd", severity="critical"}`

PR #796, opened 2021-10-13
[DownstreamMerge] Merge 2021-10-13

trozet:
Passes unit tests :disappointed_relieved:


dave-tucker:
10 seconds is the current timeout
https://github.com/openshift/ovn-kubernetes/blob/a8c57cf18434b01948225163e49b8bb2af5c73e9/go-controller/pkg/libovsdb/libovsdb.go#L25

I think I'll need to take a look at the way we do reconnect handling to be more robust.

PR #798, opened 2021-10-15
Bug 2014332: [4.8z] Scale fixes for pods/exgws

trozet:
Performance and scale fixes with pods and multiple external gateways.


mffiedler:
Tested on 4.8 cluster-bot cluster using the workload from https://bugzilla.redhat.com/show_bug.cgi?id=2014332#c6
/label qe-approved

PR #800, opened 2021-10-15
CARRY: go-ovn: prevent deadlock processing Updates during initial DB dump

dcbw:
It appears that ovsdb-server will sometimes send updates after we register
our database monitors, but before we've received and processed the initial
database dump. This causes a deadlock because during the initial monitor
setup we must lock the database cache, but the update processing also
locks the database cache.

Avoid this by making the notifier defer updates during the initial
database dump, and processing them only after the dump is done.

@trozet @dave-tucker


trozet (commented):
@dcbw seems sane to me. I was worried about blocking the notifier while processing all the updates, but @dave-tucker pointed out the rpc2 lib creates a map to handle pending incoming messages. So the buffering of incoming requests is only limited by the amount of memory the process can allocate. I basically copied your approach in:

https://github.com/ovn-org/libovsdb/pull/254

I don't think we need this commit in openshift master though. I think you just want to add this to the 4.9 PR.

PR #811, opened 2021-11-02
block metadata service for Alibaba Cloud

Lyt99:
block metadata service '100.100.100.200' on Alibaba Cloud

Signed-off-by: xiayu.lyt <xiayu.lyt@alibaba-inc.com>


go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -25,6 +25,12 @@ var iptables4OnlyCommands = [][]string{
 	{"-A", "OUTPUT", "-p", "udp", "-m", "udp", "-d", "169.254.169.254", "!", "--dport", "53", "-j", "REJECT"},
 	{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "-d", "169.254.169.254", "!", "--dport", "53", "-j", "REJECT"},
 	{"-A", "FORWARD", "-p", "udp", "-m", "udp", "-d", "169.254.169.254", "!", "--dport", "53", "-j", "REJECT"},
+	// Block cloud provider metadata IP for Alibaba Cloud
```
abhat, line 28: I think we need to figure out if OCP is running on Ali platform and then add these rules. I was thinking if we could somehow populate the infrastructure type from the infrastructure config object into a config map entry that we can read, and only add the rules for specific platform, that will be better. @trozet, @danwinship  wdyt?

kwoodson: Any updates on this @abhat @trozet @danwinship?

abhat: Closing in favor of https://github.com/openshift/ovn-kubernetes/pull/873

Lyt99:
Closed for #873

PR #816, opened 2021-11-04
Bug 2017909: EgressGW: only return unique elements from getRouteInfosForGateway()

tssurya:
If a single routeInfo had multiple external routes that matched the
same gateway IP, that routeInfo could be added to the returned list
multiple times. It was only locked once, however.

But deletePodGWRoutesForNamespace() iterates over the list returned
by getRouteInfosForGateway() and calls Unlock() on each element in
the list, which can create an unbalanced locking situation (multiple
Unlocks for a single Lock) in the above situation where a routeInfo
has multiple external routes matching the same gatewayIP.

(cherry picked from commit b5bb0d46d28d742b0e388cdefe0116b70048e432)


tssurya:
@dcbw : Seems like `e2e-metal-ipi-ovn-ipv6 ` needs an override.

dcbw:
Failures to contact image registry on v4 addresses are unrelated to this PR

PR #817, opened 2021-11-05
Send host traffic via overlay

mccv1r0:
[WIP] Send host traffic via overlay

Signed-off-by: Michael Cambria <mcambria@redhat.com>


go-controller/pkg/ovn/gateway_init.go
```diff
@@ -536,40 +541,45 @@ func syncPolicyBasedRoutes(nodeName string, matches sets.String, priority, nexth
 	// sync and remove unknown policies for this node/priority
 	// also flag if desired policies are already found
 	for _, policyCompare := range policiesCompare {
-		if strings.Contains(policyCompare, fmt.Sprintf("%s\"", nodeName)) {
+		if strings.Contains(policyCompare, nodeName) {
```
dcbw, line 544: @mccv1r0 this was probably in the original code, but I'll bet this is using the trailing " to make sure that two nodes with similar/overalpping names don't match here?

mccv1r0: Correct.  I put it back in?

PR #820, opened 2021-11-08
Bug 2019093: Not exit when ovnkube runs with '--gateway-interface none'

pliurh:
During SDN migration, there is an intermediate phase, ovn-kube runs with '--gateway-interface none' before br-ex is not created. This PR allows ovn-kube to run in such a scenario.

Fix the regression introduced by https://github.com/openshift/ovn-kubernetes/commit/1c28b968f1b0fb6252e7f4d3061b107d5cc498e0. 

**- How to verify it**
Trigger SDN migration

### DIFF
--- a/go-controller/pkg/node/gateway_init.go
+++ b/go-controller/pkg/node/gateway_init.go
@@ -179,7 +179,9 @@ func (n *OvnNode) initGateway(subnets []*net.IPNet, nodeAnnotator kube.Annotator
 	}
 
 	ifAddrs, err := getNetworkInterfaceIPAddresses(gatewayIntf)
-	if err != nil {
+	// Downstream-only hack: Do not return error when gatewayIntf is 'none', this is only for the intermediate
+	// phase of SDN migration when ovnkube runs with '--gateway-interface none'
+	if err != nil && gatewayIntf != "none" {
 		return err
 	}
 


tssurya (commented):
@trozet : PTAL. We broken the migration code during this commit: https://github.com/openshift/ovn-kubernetes/commit/1c28b968f1b0fb6252e7f4d3061b107d5cc498e0#diff-e9790f3508dd0c36e1fd2[…]370d7c9e276ddb8e2278b21L181. Although what that commit did is correct, we still rely on no error returns to then proceed with `newLocalGateway` creation.

tssurya:
@pliurh : Did you test this fix with the migration code and it worked?

tssurya:
> > @pliurh : Did you test this fix with the migration code and it worked? -> Asking since we don't have a CI job doing this.
> 
> Yes, I did test it. It works. We only have CI job for 4.8 and later branches.

Is this only an issue on 4.7 then? I just noticed the PR is against 4.7. In 4.8 and above we do things differently?

pliurh:
In 4.8 and later, the SDN migration has a different procedure. The br-ex bridge will be created before ovnkube starts. So there is no such a phase when ovnkube runs with '--gateway-interface none' anymore.

tssurya:
> In 4.8 and later, the SDN migration has a different procedure. The br-ex bridge will be created before ovnkube starts. So there is no such a phase when ovnkube runs with '--gateway-interface none' anymore.

ack thanks Peng! sg.

tssurya:
/lgtm

cc @trozet for approval.

pliurh:
@trozet Could you approve this PR?

trozet (approved):
/approve

PR #824, opened 2021-11-09
Bug 2015512: run RunDBChecker as goroutine

jluhrsen:
when the ovn-dbchecker was receiving a SIGTERM it would
catch it, but the channel never closed so it never
quits.

Signed-off-by: Jamo Luhrsen <jluhrsen@gmail.com>

**- How to verify it**


trozet:
/hold

I dont think this is needed because it will be picked up as part of the downstream merge.

jluhrsen:
> /hold
> 
> I dont think this is needed because it will be picked up as part of the downstream merge.

ok

PR #841, opened 2021-11-17
Bug 2024313: run RunDBChecker as goroutine

jluhrsen:
when the ovn-dbchecker was receiving a SIGTERM it would
catch it, but the channel never closed so it never
quits. Fix that.

Signed-off-by: Jamo Luhrsen <jluhrsen@gmail.com>

**- How to verify it**


jluhrsen:
this made it in as part of a u/s->d/s merge

PR #847, opened 2021-11-23
[WIP] syncEgressfirewall should not be limited to the current gateway mode

ricky-rav:
At startup, part of syncEgressfirewall method matches EgressFirewall ACLs in OVN with EgressFirewalls in the API server and removes spurious ACLs in OVN from the logical switches that are relevant to the current gateway mode. This patch makes sure that all logical switches are taken into account in syncEgressfirewall, to be aligned with 4.10 and also to ensure that an upgrade from one gateway mode to another doesn't bring along any trailing ACLs from previously failed executions (e.g. BZ 2011666)

removing ACLs from all switches to be aligned with 4.10

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/egressfirewall.go
+++ b/go-controller/pkg/ovn/egressfirewall.go
@@ -94,6 +94,9 @@ func newEgressFirewallRule(rawEgressFirewallRule egressfirewallapi.EgressFirewal
 
 // NOTE: Utilize the fact that we know that all egress firewall related setup must have a priority: types.MinimumReservedEgressFirewallPriority <= priority <= types.EgressFirewallStartPriority
 func (oc *Controller) syncEgressFirewall(egressFirwalls []interface{}) {
+	klog.Errorf("[syncEgressFirewall] start")
+	defer func() { klog.Errorf("[syncEgressFirewall] end") }()
+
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/egressfirewall.go
```diff
@@ -456,36 +488,48 @@ func (oc *Controller) createEgressFirewallRules(priority int, match, action, ext
...
 		for _, uuid := range uuids {
 			_, stderr, err = txn.AddOrCommit([]string{"remove", "logical_switch", logicalSwitch, "acls", uuid})
 			if err != nil {
 				return fmt.Errorf("failed to commit db changes for egressFirewall stderr: %q, err: %+v", stderr, err)
 			}
 		}
 	}
+	for _, uuid := range uuids {
+		klog.Errorf("[deleteEgressFirewallRules] removing ACL %s", uuid)
+		_, stderr, err = txn.AddOrCommit([]string{"destroy", "acl", uuid})
```
flavio-fernandes, line 528: quick thought on this: if you are worried about deleting a row that no longer exists, you can use the `--if-exists` flag to make it idem-potent. That would make the transaction not fail on you.

PR #856, opened 2021-11-30
Bug 2023220: [4.9.z] removing ACLs from all switches in deleteEgressFirewallRules

ricky-rav:
This patch makes sure that all logical switches are taken into account in deleteEgressFirewallRules and (consequently) syncEgressfirewall, regardless of the current gateway mode. This has to benefits:
- it aligns the code with 4.10 (with libovsdb)
- it ensures that syncEgressfirewall correctly removes any trailing ACLs that might be hanging around after an upgrade from an OCP version affected by the bug in BZ 2023220 (i.e. a 4.8.z < 4.8.14), where temporary ACLs marked with external ID: egressFirewall=.*-blockAll were never removed.

This fixes #2023220 for 4.9.z.

### DIFF
--- a/go-controller/pkg/ovn/egressfirewall.go
+++ b/go-controller/pkg/ovn/egressfirewall.go
@@ -107,7 +107,7 @@ func (oc *Controller) syncEgressFirewall(egressFirwalls []interface{}) {
 			fmt.Sprintf("priority>=%s", types.MinimumReservedEgressFirewallPriority),
 		)
 		if err != nil {
-			klog.Errorf("Unable to list egress firewall logical router policies, cannot cleanup old stale data, stderr: %s, err: %v", stderr, err)
+			klog.Errorf("Unable to list egress firewall ACLs, cannot cleanup old stale data, stderr: %s, err: %v", stderr, err)
 			return
 		}
 		if egressFirewallACLIDs != "" {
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/egressfirewall.go
```diff
@@ -456,22 +458,30 @@ func (oc *Controller) createEgressFirewallRules(priority int, match, action, ext
 	return nil
 }
 
-// deleteEgressFirewallRules delete the specific logical router policy/join switch Acls
+// deleteEgressFirewallRules delete the ACL identified by the given externalID from all logical switches.
 func (oc *Controller) deleteEgressFirewallRules(externalID string, txn *util.NBTxn) error {
 	logicalSwitches := []string{}
-	if config.Gateway.Mode == config.GatewayModeLocal {
```
tssurya, line 462: why are we removing the gateway modes? The ACLs should only be on node switches for lgw mode and join switches for sgw mode. `syncEgressFirewall` code had already implemented the path for removing acls from node switches for shared gateway when we moved the acls to join switch for performance.

tssurya (commented):
Could you please explain how they found left over ACLs on join switches even if it was local gateway mode (ACLs were supposed to be created only on the node switches right?) since I am a bit confused on how removing the gateway mode can help + PR says `it aligns the code with 4.10 (with libovsdb)`, 

Also there is work already in upstream to support the migration between gateway modes that uses this logic.

Is this a backport only PR? I can't seem to find the upstream equivalent...

PR #873, opened 2021-12-13
Block access to metadata service based on platform type

abhat:
This commit adds support for blocking access from workload pods
to Alibaba cloud platform's metadata service.

### DIFF
--- a/go-controller/pkg/cni/OCP_HACKS.go
+++ b/go-controller/pkg/cni/OCP_HACKS.go
@@ -4,29 +4,16 @@ package cni
 
 import (
 	"fmt"
-	"github.com/containernetworking/plugins/pkg/ns"
 	"os/exec"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 
 	utilnet "k8s.io/utils/net"
 )
 
 // OCP HACK: block access to MCS/metadata; https://github.com/openshift/ovn-kubernetes/pull/19
-var iptablesCommands = [][]string{
-	// Block MCS
-	{"-A", "OUTPUT", "-p", "tcp", "-m", "tcp", "--dport", "22623", "--syn", "-j", "REJECT"},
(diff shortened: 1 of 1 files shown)


go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -38,6 +25,25 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
...
 
+		var iptablesCommands = [][]string{
+			// Block MCS
+			{"-A", "OUTPUT", "-p", "tcp", "-m", "tcp", "--dport", "22623", "--syn", "-j", "REJECT"},
+			{"-A", "OUTPUT", "-p", "tcp", "-m", "tcp", "--dport", "22624", "--syn", "-j", "REJECT"},
+			{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22623", "--syn", "-j", "REJECT"},
+			{"-A", "FORWARD", "-p", "tcp", "-m", "tcp", "--dport", "22624", "--syn", "-j", "REJECT"},
+		}
+		metadataServiceIP := "169.254.169.254"
+		if config.Kubernetes.PlatformType == string(configv1.AlibabaCloudPlatformType) {
```
alexanderConstantinescu, line 36: I don't think this is going to work. For `config` to actually be valid we need to initialize it and have it read a config file. This is not done for the CNI server because the OVN-Kubernetes config file is not mounted on the host (it's a config map that we mount in the pod, see: https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/ovnkube-master.yaml#L835-L837) and there will be nothing in the default directory specified either: https://github.com/openshift/ovn-kubernetes/blob/3e33b921cda7c3d924079ac555343591bebbb937/go-controller/pkg/config/config.go#L1551 

I think we need to employ the same procedure done for openshift-sdn, i.e: have ovnkube-node read the ovnkube-config when it starts, write the platform type in the CNI configuration file which can then be read by the CNI server. 

I happened to write the code for this earlier today, see https://github.com/openshift/ovn-kubernetes/pull/874

abhat: Got it. Thanks. I will close my PR then.

abhat:
Closed in favor of #874

trozet (approved):
/lgtm

PR #874, opened 2021-12-13
OCP downstream only: block alibaba metadata service IP

alexanderConstantinescu:
This adds OCP specific code to block the Alibaba cloud metadata service IP.  

/assign @trozet @abhat 

**- How to verify it**

### DIFF
--- a/go-controller/pkg/cni/helper_linux.go
+++ b/go-controller/pkg/cni/helper_linux.go
@@ -369,7 +369,7 @@ func (pr *PodRequest) ConfigureInterface(podLister corev1listers.PodLister, kcli
 	}
 
 	// OCP HACK: block access to MCS/metadata; https://github.com/openshift/ovn-kubernetes/pull/19
-	err = setupIPTablesBlocks(netns, ifInfo)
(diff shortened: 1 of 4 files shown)


go-controller/pkg/cni/helper_linux.go
```diff
@@ -369,7 +369,7 @@ func (pr *PodRequest) ConfigureInterface(podLister corev1listers.PodLister, kcli
 	}
 
 	// OCP HACK: block access to MCS/metadata; https://github.com/openshift/ovn-kubernetes/pull/19
-	err = setupIPTablesBlocks(netns, ifInfo)
+	err = setupIPTablesBlocks(netns, ifInfo, pr.CNIConf.PlatformType)
```
trozet, line 372: what is the value in writing this as part of the CNI config? Why not just always use the config.Kubernetes.PlatformType?

abhat: @trozet https://github.com/openshift/ovn-kubernetes/pull/873#discussion_r768095484. Basically the CNI plugin executable call doesn't seem to have access to `config.Kubernetes.PlatformType` unless we pass it in via cni config.

PR #877, opened 2021-12-17
Bug 2033672: Multiple exgw - Use --may-exist on hybrid policy and correct hybrid policy cleanup

trozet:
Ensures we use --may-exist on 501 policies as it is fixed in OVN. Also ensures that hybrid policies are cleaned up correctly in the sync method.

### DIFF
--- a/go-controller/pkg/ovn/egressgw.go
+++ b/go-controller/pkg/ovn/egressgw.go
@@ -620,6 +620,14 @@ func (oc *Controller) delHybridRoutePolicyForPod(podIP net.IP, node string, forc
 	return nil
 }
 
+func delAllHybridRoutePolicies() {
+	_, stderr, err := util.RunOVNNbctl("lr-policy-del", types.OVNClusterRouter, types.HybridOverlayReroutePriority)
+	if err != nil {
+		klog.Errorf("Failed to remove all hybrid policies on: %s, stderr: %s, err: %v",
+			types.OVNClusterRouter, stderr, err)
+	}
+}
+
 // cleanUpBFDEntry checks if the BFD table entry related to the associated
 // gw router / port / gateway ip is referenced by other routing rules, and if
 // not removes the entry to avoid having dangling BFD entries.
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/egressgw.go
```diff
@@ -574,15 +574,11 @@ func (oc *Controller) addHybridRoutePolicyForPod(podIP net.IP, node string) erro
 		// traffic destined outside of cluster subnet go to GR
 		matchStr := fmt.Sprintf(`inport == "%s%s" && %s.src == %s`, types.RouterToSwitchPrefix, node, l3Prefix, podIP)
 		matchStr += matchDst
-		_, stderr, err := util.RunOVNNbctl("lr-policy-add", types.OVNClusterRouter, types.HybridOverlayReroutePriority, matchStr, "reroute",
+		_, stderr, err := util.RunOVNNbctl("--may-exist", "lr-policy-add", types.OVNClusterRouter, types.HybridOverlayReroutePriority, matchStr, "reroute",
```
tssurya, line 577: ++

go-controller/pkg/ovn/egressgw.go
```diff
@@ -783,8 +797,7 @@ func (oc *Controller) cleanExGwECMPRoutes() {
 		}
 
 		// if pod had no ECMP routes we need to make sure we remove any logical route policy for local gw mode
-		// for shared gateway mode, these LRPs shouldn't exist, so delete them all
-		if !podHasAnyECMPRoutes || config.Gateway.Mode == config.GatewayModeShared {
+		if !podHasAnyECMPRoutes && config.Gateway.Mode == config.GatewayModeLocal {
```
tssurya, line 800: so before we were removing LRPs; if it was either shared gateway mode (where we don't want them) OR if no ecmp routes were found for the pod - that means it was for LGW.

Now: we remove them if no ecmp routes were found for the pod AND its LGW. - and this is fine since we already nuke all policies for SGW, much better logic I'd say;

we should have done this before instead of adding the force option to `delHybridRoutePolicyForPod` to handle the SGW case which now looks redundant, but since this is sort of a backport, let's keep changes to minimal.

tssurya (approved):
/lgtm

trozet:
@anuragthehatter can you please add cherry pick approved

trozet:

PR #886, opened 2021-12-21
Bug 2034669:  Fix node connectivity to service backed by egress IP pods

alexanderConstantinescu:
Legacy commit using ovn-nbctl for the implementation. This commit will
be used for backports to older versions.

The host network couldn't connect to services backed by egress IP
matching pods since the return packets would get forwarded to the egress
nodes for these pods because of the routing policies specified for the
egress matching pods. This would in turn break the connection. Hence
what we need is to specify routing policies which forces the packet to
go to the originating GR instead of getting routed to the egress node
GR.

(cherry picked from commit 993e7d21250c7f0c2c9836c61daba40873e42a9d)

cherry-pick of https://github.com/ovn-org/ovn-kubernetes/pull/2580

/assign @dcbw @trozet 

**- How to verify it**


trozet:

trozet:
@anuragthehatter PTAL

PR #892, opened 2022-01-04
Bug 2036977: Fixes for shared to local gateway migration

trozet:
Changes-Include:
1. updating service route correctly
2. removing egress firewalls from join switch on sgw -> local

### DIFF
--- a/go-controller/pkg/node/management-port_linux.go
+++ b/go-controller/pkg/node/management-port_linux.go
@@ -56,6 +56,18 @@ func newManagementPortIPFamilyConfig(hostSubnet *net.IPNet, isIPv6 bool) (*manag
 	if config.Gateway.Mode != config.GatewayModeShared {
 		for _, subnet := range config.Kubernetes.ServiceCIDRs {
 			if utilnet.IsIPv6CIDR(subnet) == isIPv6 {
+				// ensure the route is not already configured if this was a shared gw mode -> local gw mode transition
+				if existingRoutes, err := util.GetRoutesBySubnet(subnet); err == nil {
+					for _, route := range existingRoutes {
+						if route.Gw.Equal(cfg.gwIP) {
+							continue
+						}
+						klog.Infof("Removing stale service route: %+v", route)
+						if err := util.RouteDel(&route); err != nil {
+							klog.Errorf("Failed to remove stale service route: %+v, error: %v", route, err)
+						}
+					}
+				}
 				cfg.allSubnets = append(cfg.allSubnets, subnet)
 			}
 		}
--- a/go-controller/pkg/ovn/egressfirewall.go
+++ b/go-controller/pkg/ovn/egressfirewall.go
@@ -92,28 +92,34 @@ func newEgressFirewallRule(rawEgressFirewallRule egressfirewallapi.EgressFirewal
 // -	Cleanup the old implementation (using LRP) in local GW mode -> shared GW mode implementation (using ACLs on the join switch)
 //  	For this it just deletes all LRP setup done for egress firewall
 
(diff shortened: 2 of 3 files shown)


go-controller/pkg/node/management-port_linux.go
```diff
@@ -56,6 +56,18 @@ func newManagementPortIPFamilyConfig(hostSubnet *net.IPNet, isIPv6 bool) (*manag
 	if config.Gateway.Mode != config.GatewayModeShared {
 		for _, subnet := range config.Kubernetes.ServiceCIDRs {
 			if utilnet.IsIPv6CIDR(subnet) == isIPv6 {
+				// ensure the route is not already configured if this was a shared gw mode -> local gw mode transition
+				if existingRoutes, err := util.GetRoutesBySubnet(subnet); err == nil {
+					for _, route := range existingRoutes {
+						if route.Gw.Equal(cfg.gwIP) {
```
tssurya, line 62: `if route.Gw.Equal(cfg.gwIP)`
here cfg.gwIP = ".1" IP of the podSubnet for that node. So we first check if 
```
10.96.0.0/16 via 10.244.2.1 dev ovn-k8s-mp0 
```
is configured and if not we delete the old route and re-add the new one towards mp0... since if the sgw route existed, then `route.Gw` would be the nodeSubnet.1 and not podSubnet.1.
logic lgtm.

tssurya (commented):
lgtm, I'll wait for the other fixes if any.

trozet:
tested this with cluster bot and ran some e2e network tests. Looks good. Would be nice to have a CI job to test this against before we merge (@tssurya is working on it)

tssurya:
> tested this with cluster bot and ran some e2e network tests. Looks good. Would be nice to have a CI job to test this against before we merge (@tssurya is working on it)

https://github.com/openshift/release/pull/24985 - what that job would do is,:

1- bring up an aws cluster on 4.9 shared gateway mode, 
2- create the `gateway-mode-config` map
3- wait for CNO to roll out the new daemon sets in LGW mode
4- run the e2e tests that are run in the normal e2e-aws-ovn job.

Does that sound good @trozet?

tssurya:
Once we get openshift/release#24985 in, and then trigger that here (maybe a new PR is needed to pick up new jobs?) and ensure it passes we should be good to go.

tssurya:
```
alert TargetDown fired for 60 seconds with labels: {job="machine-config-daemon", namespace="openshift-machine-config-operator", service="machine-config-daemon", severity="warning"}
```

but looking at the must-gather, I am not seeing any continuous failures, looking at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_ovn-kubernetes/892/pull-ci-openshift-ovn-kubernetes-release-4.9-e2e-aws-ovn-shared-to-local-gateway-mode-migration/1481278392148955136/artifacts/e2e-aws-ovn-shared-to-local-gateway-mode-migration/openshift-e2e-test/build-log.txt the tests are starting after ensuring all the operators are healthy and at that point the "machine-config" operator seems healthy. I am not exactly sure when this targetDown alert is getting fired and how long this is lasting,

trozet:
ci/prow/e2e-aws-ovn-shared-to-local-gateway-mode-migration  job is failing because an alert fires that MCD loses http connectivity to the kapi-server for a brief time during the test. This coincides with when ovnkube-masters are rolling out. During SGW-> LGW migration there will be some temporary network outage for the following cases:
1. pod egress: ovnkube-master has to reconfigure the routes for external traffic to go via mp0 instead of GR
2. host -> svc -> hairpin back to host: ovnkube master has to create the DGP and routes to allow hairpin for LGW

PR #896, opened 2022-01-10
[WIP]: Hacky way of doing ITP:preferLocal for openshift-dns:default

tssurya:

**- How to verify it**


go-controller/pkg/ovn/controller/services/load_balancer.go
```diff
@@ -298,6 +302,7 @@ func buildPerNodeLBs(service *v1.Service, configs []lbConfig, nodes []nodeInfo)
 					switchV4targetips = util.FilterIPsSlice(switchV4targetips, node.nodeSubnets(), true)
 					switchV6targetips = util.FilterIPsSlice(switchV6targetips, node.nodeSubnets(), true)
 				}
+
```
tssurya, line 305: ugh, go away

tssurya:
https://bugzilla.redhat.com/show_bug.cgi?id=2038386 is the ACTIVE bug for the failing job `4.10-upgrade-from-stable-4.9-e2e-aws-ovn-upgrade` and `e2e-metal-ipi-ovn-dualstack` seems to be PERMA failing.

go-controller/pkg/ovn/controller/services/load_balancer_test.go
```diff
@@ -1369,3 +1369,126 @@ func Test_buildPerNodeLBs(t *testing.T) {
...
+			expected: []ovnlb.LB{
+				{
+					Name:        "Service_openshift-dns/dns-default_TCP_node_router_node-a_merged",
+					ExternalIDs: defaultExternalIDs,
+					Routers:     []string{"gr-node-a", "gr-node-b"},
+					Protocol:    "TCP",
+					Rules: []ovnlb.LBRule{
+						{
+							Source:  ovnlb.Addr{"192.168.1.1", 80},
+							Targets: []ovnlb.Addr{{"10.128.0.2", 8080}, {"10.128.1.2", 8080}},
```
trozet, line 1449: thanks for the nice test case. I guess the one caveat here is that if a host -> DNS service it could go to any node rather than it's own node. That's probably OK?

tssurya: I think that should be fine, according to the bz, they are worried only about egress traffic from the pods. Maybe if we get yelled at we can do the `host->clusterIP->pod` then? Because for that I'd need to also hack the vip:ip on the GR-routers which probably isn't that hard to do, just that we'd have twice as many more LBs.

tssurya: also, the `dns-default` clsuterIP svc is mainly used to dns resolution for stuff in overlay. So I think we are ok if we don't cover the host->svc->pod case since that's technically "outside the overlay".

abhat:
job is perma-failing.

PR #900, opened 2022-01-12
[DNM] PR to test models generated with copy/equal methods

jcaamano:
/hold


flavio-fernandes:
Please see: https://github.com/ovn-org/ovn-kubernetes/pull/2764

PR #907, opened 2022-01-14
Bug 2025467: ETP=local,SGW: Add DNAT rule towards 169.254.169.3

tssurya:
Host -> svc (ETP=local) backed by ovn pods
does not work in SGW because we add the DNAT
rule that converts the NP to CIP before it
hits the LB on the GR.

Current flow which is wrong:

1) traffic from host gets DNAT-ed to clusterIP svc using iptables
2) traffic sent to br-ex
3) hits the GR load balancer
4) gets DNAT-ed to the backend pods,
5) depending on if this is on the same node or a different one,
we'll have packet delivered to the pod if its on the same node,
or it passes via geneve tunnel to the destination node where the pod lives.

Technically if the backends are not local to the node, the
request should get rejected. This PR removes the iptable DNAT rule
towards CIP if the traffic is of ETP=local type. Instead it adds
the DNAT rule towards masqueradeIP which we already do for LGW
mode. With that, it will send the packet from host into OVN
via mp0 and hits the node-local-switch LB preserving sourceIP.

New flow:

1) NP/EIP/LIP traffic from host hits PRE-ROUTING chain where it
gets DNAT-ed to masqueradeIP:NP.
2) Then it hits OVN-KUBE-SNAT-MGMTPORT where its src-IP is
preserved.
3) enters OVN via mp0, hits the load balancer on the switch
4) gets DNAT-ed to the backend pods if they exist locally on the
node else get rejected.

(cherry picked from commit a258096271fb7c45ad4281a5a1239007bc30d330)

**- How to verify it**

### DIFF
--- a/docs/design/external_traffic_policy.md
+++ b/docs/design/external_traffic_policy.md
@@ -208,13 +208,7 @@ networked pods or cluster networked pods
 
 ### Host -> Service -> OVN Pod
 
-#### **Shared Gateway Mode**
-
-Documentation will be updated after https://bugzilla.redhat.com/show_bug.cgi?id=2025467 is fixed.
-
-#### **Local Gateway Mode**
-
-This case is similar to steps 3-6 on `External -> Service -> OVN Pod` traffic senario we saw above for local gateway.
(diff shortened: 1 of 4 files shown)


abhat:
Known perma-failing job tracked by https://bugzilla.redhat.com/show_bug.cgi?id=2040671

abhat:
Job is perma-failing. 

PR #909, opened 2022-01-14
Bug 2034577: Set l3GWConfig.mode correctly

tssurya:
The "mode" field in the l3gw annotation
actually denotes if there is a gateway
or not, and it means "have we created
an interface that we share with the node".
The value for this is always "shared" for
both local and shared modes and this is
confusing. Let's just update this with
the right mode.

We can also remove the check for modes
when checking vlan-id since both modes
support this.

(cherry picked from commit 7922e0687bae1d2c7ee869b88e08e520f36d11b8)

**- How to verify it**

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -216,7 +216,7 @@ func gatewayInitInternal(nodeName, gwIntf, egressGatewayIntf string, subnets []*
 	}
 
 	l3GwConfig := util.L3GatewayConfig{
-		Mode:           config.GatewayModeShared,
+		Mode:           config.Gateway.Mode,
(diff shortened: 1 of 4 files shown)


tssurya:
HO jobs are failing because we don't have https://github.com/ovn-org/ovn-kubernetes/pull/2720/files#diff-9811e6f2e1ecf87a3a4c6c1b6b8e9c0df5d921c8591985a1c09e93e95e9dae52R56 yet.

dcbw:
due to version skew between e2e tests and apiserver that will be fixed soon

dcbw:
due to version skew between e2e tests and apiserver that will be fixed soon

PR #910, opened 2022-01-17
Bug 2029742: egressip: fix usage of clientModel doAfter

flavio-fernandes:
When performing delete/mutation in ovn tables using modelClient.Delete(),
any rows returned by predicate will be removed.
Using doAfter is too late to decide the row's fate.

Also fixed issue in FindNatsUsingPredicate so that creating a slice of
NAT pointers uses the right address values.


flavio-fernandes:
@alexanderConstantinescu @trozet This is the back port for https://github.com/ovn-org/ovn-kubernetes/pull/2735

PR #928, opened 2022-01-25
Bug 2002868: metrics: enable host OVS process metrics collection

dcbw:
Gather process metrics from the host OVS process using the default
pidfile locations.

@mohit-sheth

### DIFF
--- a/go-controller/pkg/metrics/metrics.go
+++ b/go-controller/pkg/metrics/metrics.go
@@ -31,6 +31,7 @@ const (
 	MetricOvnSubsystemController = "controller"
 	MetricOvsNamespace           = "ovs"
 	MetricOvsSubsystemVswitchd   = "vswitchd"
+	MetricOvsSubsystemDB         = "db"
 
 	ovnNorthd     = "ovn-northd"
 	ovnController = "ovn-controller"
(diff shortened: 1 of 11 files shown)


dcbw:
/retest
aws platform issues
```
level=error msg=Error: Provider produced inconsistent result after apply
level=error
level=error msg=When applying changes to module.vpc.aws_route.igw_route[0], provider
level=error msg="openshift/local/aws" produced an unexpected new value: Root resource was
level=error msg=present, but now absent.
```

squeed:
What metrics does this expose? Is there any chance of a cardinality explosion?

dcbw:
@squeed it collects:

	ch <- c.cpuTotal
	ch <- c.openFDs
	ch <- c.maxFDs
	ch <- c.vsize
	ch <- c.maxVsize
	ch <- c.rss
	ch <- c.startTime

whenever prometheus scrapes the endpoints.

squeed:
Seems reasonable, doesn't seem like there any labels.

/lgtm

dcbw:
Upstream handles OVS metrics in a separate process (ovs-exporter) that we don't run (nor need to) so need to add an upstream config option to run them as part of the node metrics.

PR #934, opened 2022-01-27
Bug 2042001: Adds wait method for ovsdb operations that created named objects

trozet:
Due to a bug in OVSDB server where a single txn may become duplicated and cause multiple DB entries or txn failures, we add a wait method to every create operation where the resource has a unique name. This guards against OVSDB server from creating the duplicate db entry.

### DIFF
--- a/go-controller/go.mod
+++ b/go-controller/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/onsi/gomega v1.14.0
 	github.com/openshift/api v0.0.0-20211201215911-5a82bae32e46
 	github.com/openshift/client-go v0.0.0-20211202194848-d3f186f2d366
-	github.com/ovn-org/libovsdb v0.6.1-0.20220124205349-71bfee870326
+	github.com/ovn-org/libovsdb v0.6.1-0.20220127023511-a619f0fd93be
 	github.com/pkg/errors v0.9.1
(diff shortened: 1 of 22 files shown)


flavio-fernandes (commented):
/lgtm

dcbw:
Upstream CI completed successfully
/lgtm
from me too

jcaamano (commented):
One other idea would be to add a generic `func checkDuplicate(model Model) (Operation, error)` method where you could build the custom operation required for each model type, just like we have to generally get/set UUIDs for models:

https://github.com/ovn-org/ovn-kubernetes/blob/c0be0a3ede33e2798b06fefa099359fb13a6a180/go-controller/pkg/libovsdbops/model.go#L12-L19

It would return nil if no dup check is required. You could then easily generate the appropriate operation and condition condition for all model types, including those that don't have a name. They would all be in the same place and organized and we probably would not need to check for user error or corner cases in model client as model client would use it directly on creates. This also avoids changing the model client API with the new Name parameter and having to update all its uses to include it.

dcbw:
/override ci/prow/e2e-aws-ovn
https://bugzilla.redhat.com/show_bug.cgi?id=2047501
This is a kubelet static pod problem (race?) that doesn't appear related to this change at all...

PR #935, opened 2022-01-27
WIP Bug 2042999: restart pod on non-retriable failures when deleting stale objects

flavio-fernandes:


dcbw:
/hold
while Flavio reworks the upstream PR

flavio-fernandes:
Closing. fixed via https://github.com/ovn-org/ovn-kubernetes/pull/2773  .

PR #936, opened 2022-01-27
WIP [release-4.9] Bug 2047416: restart pod on non-retriable failures when deleting stale objects

flavio-fernandes:

Conflicts:
  go-controller/pkg/ovn/pods.go
  go-controller/pkg/ovn/policy.go

(cherry picked from commit 033cc76d0c1398b6e430d49e94b9a40efbc1f933)


dcbw:
/hold
while Flavio reworks the upstream PR

flavio-fernandes:
This change has been folded into https://github.com/openshift/ovn-kubernetes/pull/981 .
Closing.

PR #943, opened 2022-02-03
[DNM] metrics: fix ticker usage that was leaking

jcaamano:
Just a test PR
/hold


jcaamano:
This is not actually that much of an issue as long as the function is called once in the lifecycle of the program.

PR #945, opened 2022-02-08
[release-4.10]  Bug 2052017: restart pod on non-retriable failures when deleting stale objects

flavio-fernandes:

Signed-off-by: Flavio Fernandes [flaviof@redhat.com](mailto:flaviof@redhat.com)


flavio-fernandes:
Holding this PR to cherry-pick the commits from @tssurya :  https://github.com/ovn-org/ovn-kubernetes/pull/2787

PR #926, opened 2022-01-24
Bug 2043961: Fix pod-creation-retry

tssurya:
This commit fixes addRetryPods to correctly
construct the retryPods cache. Currently
since we assign pointer as value in
retryPods cache in a loop, the values
get overwritten and this causes a nasty
bug in the retry pod logic. Some pods
get retried while others don't.

This PR also fixes addLogicalPort to
fail hard if the node.l3gateway annotation
is not set in disableSNATGWMode.

Signed-off-by: Surya Seetharaman <suryaseetharaman.9@gmail.com>
(cherry picked from commit c63ddd62103314d8014ba2d68c596182e5b78f12)

**- What this PR does and why is it needed**

**- Special notes for reviewers**

**- How to verify it**

**- Description for the changelog**

PR #951, opened 2022-02-10
Bug 2053310: [release-4.9] Fix pod-creation-retry

dcbw:
This commit fixes addRetryPods to correctly
construct the retryPods cache. Currently
since we assign pointer as value in
retryPods cache in a loop, the values
get overwritten and this causes a nasty
bug in the retry pod logic. Some pods
get retried while others don't.

Cherry-pick of https://github.com/openshift/ovn-kubernetes/pull/926


tssurya:
/lgtm
thanks Dan!

PR #985, opened 2022-03-02
Bug 2060080: [release-4.9] Network Policy fixes

trozet:


trozet:
results look good. This will need careful review @dcbw PTAL

dcbw:
@trozet can't find anything obviously wrong...

Anyone else want to take a pass too?

PR #993, opened 2022-03-11
Bug 2063321: [4.11] After reboot egress node, lr-policy-list was not correct

flavio-fernandes:
commit 1:
Successful assignment of egress IP is really useful to log. This change makes it
happen on Info level by default.

commit 2:
In cases where OVN database for logical router policies or NATs
used by EgressIPs have stale nexthops or wrong external_ips, the
sync function should remove them, so the proper row/column gets set.


flavio-fernandes:
/hold

Going to use this bug to address issue found with sync, where stale nexthops are not properly cleaned up.

flavio-fernandes:
/hold
Will do this backport together with all other changes upstream

flavio-fernandes:
... Folded into  https://github.com/openshift/ovn-kubernetes/pull/1000

PR #1028, opened 2022-04-07
Fix gofmt for downstream files

abhat:
Adds go:build tag for OCP_HACKS files.


trozet (approved):
/lgtm

PR #1046, opened 2022-04-21
Bug 2077369: [release-4.9] fix ipv6 network policy 

JacobTanenbaum:
without adding the exemption for neighbor detection on network policies
the network policy will deny all attempts for ipv6 to send traffic

**- How to verify it**


hardys:
> /label cherry-pick-approved

@anuragthehatter please can you verify https://bugzilla.redhat.com/show_bug.cgi?id=2074839 - this won't merge until bugzilla/valid-bug is applied, which means the 4.10 bz has to be verified, thanks!

PR #1056, opened 2022-04-27
Bug 2072780: call clearInitialNodeNetworkUnavailableCondition for noHostSubnet nodes

JacobTanenbaum:
clean application of upstream commit: 28a23809114ea40bc5e7617809f575db5918fde3

a correctly configured noHostSubnet node will never become avalible on
GCP because the unavailable condition will never be cleared.

**- How to verify it**


JacobTanenbaum:
yes we can wait for that

PR #1069, opened 2022-05-02
Bug 2080882: Fix incorrectly logged error when searching for ovnkube master pod

martinkennelly:
Before registering OVN northd metrics, we check to see if ovnkube
master is co-located on the same node and repeatedly check if it
is present because ovnkube node may start before ovnkube master.
This repeated checking only causes one error mode to occur which
is a timeout error and this is incorrectly logged as an error
confusing the user.
The only failure path to occur within 'checkPodRunsOnGivenNode' can
be seen with increased log level.

(cherry picked from commit 1ef9add179ecf74cab8f819cb416fbe254c7745a)

To verify this:
1. Grep the logs of the ovnkube-node container that isn't on a control plane node for the Error message 'Timed out while checking if OVNKube Master Pod runs'. You should not get a result.
2. Grep the logs of the ovnkube-node container that isn't on a control plane node for the Info message 'Not registering OVN North Metrics because OVNKube Master Pod was not found running on this'. You should get a result.


martinkennelly:
/retest

Unrelated flake in metal job.

PR #1071, opened 2022-05-02
Bug 2080925: Don't spawn go routines for adding metric events

martinkennelly:
Use a queue instead of spawning go routines to add events.


martinkennelly:
/retest

ovn-upgrade failed due to unable to pull manifest - unrelated to this change.

PR #1073, opened 2022-05-02
[wip] debug ovn-controller

dcbw:


go-controller/pkg/ovn/pods.go
```diff
@@ -429,7 +429,7 @@ func (oc *Controller) addLogicalPort(pod *kapi.Pod) (err error) {
 
 	if needsIP {
 		// try to get the IP from existing port in OVN first
-		podMac, podIfAddrs, err = oc.getPortAddresses(logicalSwitch, portName)
+		podMac, podIfAddrs, err = oc.getPortAddresses(logicalSwitch, portName, existingLSP)
```
tssurya, line 432: oh wow this is a nice catch! so we were again looking up LSP inside `GetPortAddresses` !

PR #1084, opened 2022-05-09
[release-4.10] Bug 2083116: delete SNAT2NIP if pod.node == egressNodeServingPod

tssurya:
We should be deleting the snat towards nodeIP before
adding snat towards egressIP only from the node
where the pod lives if that same node is serving
as the egressIPnode for the pod.

A second problem that was fixed was in deleteEgressIPAssignments
we were re-adding the setup for SNAT towards nodeIP only if
pod had no egressIPs attached to it, which is wrong since depending
on the location of the pod, we'd need to re-add the SNAT towards
nodeIP even if the pod is being served by another egressIP.

(cherry picked from commit 044ead332e4e26cbc141afc0739b37b68bdb76c7)

Clean cherry-pick from master

**- How to verify it**

### DIFF
--- a/go-controller/pkg/kube/kube.go
+++ b/go-controller/pkg/kube/kube.go
@@ -241,7 +241,6 @@ func (k *Kube) UpdateEgressIP(eIP *egressipv1.EgressIP) error {
 }
 
 func (k *Kube) PatchEgressIP(name string, patchData []byte) error {
-	klog.Infof("Patching status on EgressIP %s", name)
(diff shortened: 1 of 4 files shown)


andreaskaris:
/lgtm

qq: Given that this is downstream, it makes no sense to cherry-pick that fix here right? I think it was related to the upstream ci failures and given that we're not running the tests, anyway .......

andreaskaris (approved):
/lgtm

tssurya:
@huiran0826 : could you PTAL? you have verified this on 4.11 already: https://bugzilla.redhat.com/show_bug.cgi?id=2070929

PR #1087, opened 2022-05-11
Retry schema migration

alvaroaleman:

In Hypershift, it is possible that when the `ovn-dbchecker` container runs the database (that runs in the same pod in a different container) is not yet fully up, especially when running it with multiple replicas. This causes the `ovn-dbchecker` to crash and the container to go into crashloop backoff, which in turn can slow down the rollout. An example error is the below one:
```
F0511 04:31:57.928725       1 ovndbmanager.go:44] NBDB Upgrade failed: %!w(*fmt.wrapError=&{failed to get schema version for NBDB, stderr: "ovsdb-client: transaction returned error: {\"details\":\"get_schema request specifies database OVN_Northbound which is not yet available because it has not completed joining its cluster\",\"error\":\"database not available\"}\n", error: OVN command '/usr/bin/ovsdb-client -t 10 get-schema-version unix:/var/run/ovn/ovnnb_db.sock OVN_Northbound' failed: exit status 1 0xc00042a160})
```

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovndbmanager/ovndbmanager.go
+++ b/go-controller/pkg/ovndbmanager/ovndbmanager.go
@@ -41,7 +41,7 @@ func RunDBChecker(kclient kube.Interface, stopCh <-chan struct{}) {
 	go func() {
 		defer wg.Done()
 		if err := upgradeNBDBSchema(); err != nil {
-			klog.Fatalf("NBDB Upgrade failed: %w", err)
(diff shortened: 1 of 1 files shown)


alvaroaleman:
Suggesting this upstream instead: https://github.com/ovn-org/ovn-kubernetes/pull/2983

PR #972, opened 2022-02-25
Bug 2058683: [4.9z] Multiple ExGW cache validation/improvements

trozet:
Cherry pick of: https://github.com/openshift/ovn-kubernetes/pull/895

minor conflicts were in 2nd commit:
        both modified:   go-controller/pkg/ovn/namespace.go
        both modified:   go-controller/pkg/ovn/pods.go

PR #1088, opened 2022-05-11
[release-4.9] Bug 2083239: [Downstream-only][4.9-only] Always add complete efw ACL rules

andreaskaris:
Recreate Egress Firewall ACL rules including their priority always.
This affects 4.9 only because:
In 4.8, we use individual transactions. In 4.10, we use libovsdb. In
4.9, we are in the odd situation that we use a single transaction for
the delete/add EgressFirewall, however during the add we readd a not
yet deleted rule (due to the pending transaction) and thus delete and
then immediately add the same rule with the same UUID, leading to rules
with duplicated priority.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2083239
For more details, see: https://bugzilla.redhat.com/show_bug.cgi?id=2083239#c8

Signed-off-by: Andreas Karis <ak.karis@gmail.com>

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/egressfirewall.go
+++ b/go-controller/pkg/ovn/egressfirewall.go
@@ -466,34 +466,17 @@ func (oc *Controller) createEgressFirewallRules(priority int, match, action, ext
 	} else {
 		logicalSwitches = append(logicalSwitches, types.OVNJoinSwitch)
 	}
-	uuids, stderr, err := util.RunOVNNbctl("--data=bare", "--no-heading",
-		"--columns=_uuid", "--format=table", "find", "ACL", match, "action="+action,
-		fmt.Sprintf("external-ids:egressFirewall=%s", externalID))
-	if err != nil {
-		return fmt.Errorf("error executing find ACL command, stderr: %q, %+v", stderr, err)
-	}
 	sort.Strings(logicalSwitches)
 	for _, logicalSwitch := range logicalSwitches {
-		if uuids == "" {
-			id := fmt.Sprintf("%s-%d", logicalSwitch, priority)
-			_, stderr, err := txn.AddOrCommit([]string{"--id=@" + id, "create", "acl",
-				fmt.Sprintf("priority=%d", priority),
-				fmt.Sprintf("direction=%s", types.DirectionToLPort), match, "action=" + action,
-				fmt.Sprintf("external-ids:egressFirewall=%s", externalID),
-				"--", "add", "logical_switch", logicalSwitch,
-				"acls", "@" + id})
-			if err != nil {
-				return fmt.Errorf("failed to commit db changes for egressFirewall  stderr: %q, err: %+v", stderr, err)
-
-			}
-
-		} else {
-			for _, uuid := range strings.Fields(uuids) {
-				_, stderr, err := txn.AddOrCommit([]string{"add", "logical_switch", logicalSwitch, "acls", uuid})
-				if err != nil {
-					return fmt.Errorf("failed to commit db changes for egressFirewall stderr: %q, err: %+v", stderr, err)
-				}
-			}
+		id := fmt.Sprintf("%s-%d", logicalSwitch, priority)
+		_, stderr, err := txn.AddOrCommit([]string{"--id=@" + id, "create", "acl",
+			fmt.Sprintf("priority=%d", priority),
+			fmt.Sprintf("direction=%s", types.DirectionToLPort), match, "action=" + action,
+			fmt.Sprintf("external-ids:egressFirewall=%s", externalID),
+			"--", "add", "logical_switch", logicalSwitch,
+			"acls", "@" + id})
+		if err != nil {
+			return fmt.Errorf("failed to commit db changes for egressFirewall  stderr: %q, err: %+v", stderr, err)
 		}
 	}
 	return nil


go-controller/pkg/ovn/egressfirewall.go
```diff
@@ -466,11 +466,15 @@ func (oc *Controller) createEgressFirewallRules(priority int, match, action, ext
 	} else {
 		logicalSwitches = append(logicalSwitches, types.OVNJoinSwitch)
 	}
-	uuids, stderr, err := util.RunOVNNbctl("--data=bare", "--no-heading",
-		"--columns=_uuid", "--format=table", "find", "ACL", match, "action="+action,
-		fmt.Sprintf("external-ids:egressFirewall=%s", externalID))
-	if err != nil {
-		return fmt.Errorf("error executing find ACL command, stderr: %q, %+v", stderr, err)
+	uuids := ""
+	if !ignoreUuidsOnCreation {
```
npinaeva, line 470: I don't quite understand the purpose of this block - on create there should be no acl rules with the same `external-ids:egressFirewall` value, and if they exist it looks like deletion error
But in general not using this part for update if definitely the right thing

andreaskaris: Thanks for looking into this. 

Dito here, I'm not super sure why we have that "add by uuid" part there at all. But it's in 4.8, too:
https://github.com/openshift/ovn-kubernetes/blob/release-4.8/go-controller/pkg/ovn/egressfirewall.go#L412
https://github.com/openshift/ovn-kubernetes/blob/release-4.8/go-controller/pkg/ovn/egressfirewall.go#L431
It works in 4.8 for updates because the update operation is executed in 2 different transactions, first a delete, then an add, so for updates, `if uuids == "" {` is always true and thus 4.8 doesn't run into the same problem as 4.9. 
And it has been that way ever since the original PR/commit and wasn' t discussed there, at all.
Commit: https://github.com/openshift/ovn-kubernetes/commit/f18a027078c2bd908197d6cc9281841a0873211d
PR: https://github.com/ovn-org/ovn-kubernetes/pull/1901
And was then broken by the subsequent addition of the transaction txn identifier and the associated logic.
Perhaps it's needed due to the Watch trying to readd rules on startup?
~~~
 771 // WatchEgressFirewall starts the watching of egressfirewall resource and calls
 772 // back the appropriate handler logic
 773 func (oc *Controller) WatchEgressFirewall() *factory.Handler {
 774         return oc.watchFactory.AddEgressFirewallHandler(cache.ResourceEventHandlerFuncs{
 775                 AddFunc: func(obj interface{}) {
 776                         egressFirewall := obj.(*egressfirewall.EgressFirewall).DeepCopy()
 777                         txn := util.NewNBTxn()
 778                         addErrors := oc.addEgressFirewall(egressFirewall, txn, false)
 (...)
 NORMAL  ᚠ 2083239  ~/development/ovn-kubernetes/go-controller/pkg/ovn/ovn.go 
~~~

I decided to use something that has the lowest risk, and limiting the code change to updates of firewall rules (as opposed to the add logic) seems to be the safest change with the lowest impact to me.

andreaskaris:
@trozet @JacobTanenbaum what do you think?

go-controller/pkg/ovn/egressfirewall.go
flavio-fernandes: That is bc the update blindly deletes and re-adds the ACLs.
And since the deletion is queued up in the txn, the looking up of the acls here would find the old ones which is not what we want.

flavio-fernandes (commented):
I added some comments, but do not consider any of them a must have.
So...
/lgtm

go-controller/pkg/ovn/egressfirewall.go
trozet: @JacobTanenbaum you probably wrote this code. Can you comment?

andreaskaris:
New PR pushed and tested against that manually:
~~~
[akaris@linux egress-firewall-acl]$ oc apply -f nginx.yaml 
sleep 5
oc apply -f egressfw-1.yaml 
sleep 5
oc apply -f egressfw-2.yaml



deployment.apps/nginx-deployment created
egressfirewall.k8s.ovn.org/default created
egressfirewall.k8s.ovn.org/default configured
~~~

Result:
~~~
[akaris@linux egress-firewall-acl]$ oc rsh -n openshift-ovn-kubernetes $(get_leader_pod)
Defaulted container "northd" out of: northd, nbdb, kube-rbac-proxy, sbdb, ovnkube-master, ovn-dbchecker
sh-4.4# ovn-nbctl acl-list join
  to-lport 10000 ((ip4.dst == $a15897474889412640632) && ip4.src == $a5154718082306775057 && ((tcp && ( tcp.dst == 25 ))) && inport == "jtor-ovn_cluster_router") allow
  to-lport  9999 ((ip4.dst == 0.0.0.0/0) && ip4.src == $a5154718082306775057 && inport == "jtor-ovn_cluster_router") drop
sh-4.4# 

~~~

andreaskaris:
Looking at the test failure for e2e-aws-ovn-local-gateway, https://github.com/openshift/ovn-kubernetes/pull/972 ran into exactly the same issue:
https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-ovn-kubernetes-release-4.9-e2e-aws-ovn-local-gateway

andreaskaris:
The failure is due to: https://bugzilla.redhat.com/show_bug.cgi?id=2089831 (hence I might need an override for ci/prow/e2e-aws-ovn-local-gateway )

PR #1098, opened 2022-05-18
Bug 2088627: EgressIP NATs are not being cleared correctly from the logical router 

flavio-fernandes:
While deleting EgressIP nats from a logical router find the router that has the nat instead of relying on the node name.

(cherry picked from commit 2f8ec3310ea608b7e0b3e905f8ad3fdcd5d057e3)


flavio-fernandes:
@kyrtapz this backport from upstream seems to be missing on 4.10 ... 4.8 !

flavio-fernandes:
hi @jechen0648 can you please add the `cherry-pick-approved` label here? or lmk if there is something else we should do before merging this.

PR #1103, opened 2022-05-20
[release-4.9] Bug 2088630: EgressIP NATs are not being cleared correctly from the logical router

flavio-fernandes:
When handling NAT table updates for EgressIPs, the previous implementation
relied on node from egress ip status field to locate logical router.

Additionally, update of NAT rows did not check logical_port field and
ended up with a SNAT row using the wrong logical router.


flavio-fernandes:
@kyrtapz @trozet PTAL

tssurya (approved):
/lgtm

PR #1125, opened 2022-06-03
Dockerfile: bump to Go 1.18 base image

dcbw:
Might as well get on the same train the rest of 4.11 is on.


danwinship:
dup of #937 isn't it?

dcbw:
Yep.

PR #1127, opened 2022-06-03
Bug 2093396: Remove node-tainting for too-small MTU

msherif1234:
The "MTU is too small" case never happens in normal operation, and
it required nodes patch access.
Get rid of tainting and just return an error which will prevent
node from becomig ready.

Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>
(cherry picked from commit 32d930197dbc65206d640a7b065889ad06e8f5e3)


dcbw:
/override ci/prow/4.11-upgrade-from-stable-4.10-e2e-aws-ovn-upgrade
https://bugzilla.redhat.com/show_bug.cgi?id=2081087
https://bugzilla.redhat.com/show_bug.cgi?id=2081084

PR #1128, opened 2022-06-03
Bug 2092889: update all egress ACLs' direction to "from-lport"

JacobTanenbaum:
clean cherry-pick of upstream commit bc941c289d6e8ddfd6fcee4e31fb85100ae97cf8

Direction of the ACLs built for multicast egress policy was always
"from-lport", so we cannot determine if we need to update egress ACL
direction based on the first egress ACL that found.

Signed-off-by: Yun Zhou <yunz@nvidia.com>

**- How to verify it**


JacobTanenbaum:
/retest

looks like something from the ci system messed up

PR #1137, opened 2022-06-14
[release-4.8] Bug 2097018: Duplicated IPs can be assigned to multiple Pods

martinkennelly:
This commit is is not a clean cherry-pick of
https://github.com/openshift/ovn-kubernetes/commit/12848e73062711e67d04bd676d2a7eb84fc7cad2.

This patch ensures that retry logic for pods is after syncPods
is executed. This is accomplished serially with the AddPodHandler
func's parameter 'processExisting' which we pass the syncPods func.
This ensures we exec 'syncPods' once before calling the retry funcs.
However, we need to fix the case where 'syncPods' fails in future
patches.

Also, when addNode() failed in addNodeAnnotations(), the node's IPAM can be
overwritten by subsequent addNode() retry attempts. As the result, the
same IP can be allocated to multiple pods.

All credit to Yun Zhou (yunz@nvidia.com).

### DIFF
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -1051,11 +1051,6 @@ func (oc *Controller) addNode(node *kapi.Node) ([]*net.IPNet, error) {
 			}
 		}
 	}()
-	// Ensure that the node's logical network has been created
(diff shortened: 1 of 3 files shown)


trozet (approved):
/lgtm

PR #1138, opened 2022-06-14
Bug 2099616:1.24 bump

ricky-rav:
Fixes #2099616


ricky-rav:
In order for CI jobs to succeed, we should first merge the necessary RBAC changes in CNO: https://github.com/openshift/cluster-network-operator/pull/1490

PR #1147, opened 2022-06-16
Bug 2082599: add upper bound to number of failed attempts

ricky-rav:
The retry logic should not attempt to add/update/delete an object indefinitely. Adding an upper bound to the maximum number of times we can attempt to add/update/delete a given object, as we do already in level-driven controllers.

(cherry picked from commit ca4559b06281cb12bcf675b39da3b99701e020f4)

fixes #2082599


trozet (approved):
/lgtm

PR #1148, opened 2022-06-16
[release-4.10] Bug 2092895: update all egress ACLs' direction to "from-lport"

JacobTanenbaum:
This is not a clean application of https://github.com/openshift/ovn-kubernetes/commit/bc941c289d6e8ddfd6fcee4e31fb85100ae97cf8 but it gets the functionality. There would have been many more patches to backport if we wanted a clean application of the upstream patch. 

Direction of the ACLs built for multicast egress policy was always
"from-lport", so we cannot determine if we need to update egress ACL
direction based on the first egress ACL that found.

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/policy.go
+++ b/go-controller/pkg/ovn/policy.go
@@ -1291,8 +1291,7 @@ func (oc *Controller) destroyNetworkPolicy(np *networkPolicy, lastPolicy bool) e
 		return fmt.Errorf("error cannot sync NetworkPolicy Egress obj: %v", err)
 	}
 	allEgressACLs = append(allEgressACLs, egressACLs...)
-	// if the first egress ACL is correct they should all be correct and not need to update
-	if allEgressACLs != nil && allEgressACLs[0].Direction != nbdb.ACLDirectionFromLport {
+	if len(allEgressACLs) != 0 {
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/policy.go
```diff
@@ -1291,8 +1291,7 @@ func (oc *Controller) destroyNetworkPolicy(np *networkPolicy, lastPolicy bool) e
 		return fmt.Errorf("error cannot sync NetworkPolicy Egress obj: %v", err)
 	}
 	allEgressACLs = append(allEgressACLs, egressACLs...)
-	// if the first egress ACL is correct they should all be correct and not need to update
-	if allEgressACLs != nil && allEgressACLs[0].Direction != nbdb.ACLDirectionFromLport {
+	if allEgressACLs != nil {
```
trozet, line 1294: I think using len(allEgressACls) > 0  is probably safer

trozet (commented):
one nit on the syntax, otherwise looks good

go-controller/pkg/ovn/policy.go
JacobTanenbaum: done

trozet (approved):
/lgtm

PR #1153, opened 2022-06-21
Bug 2099822: Fix pointer loop var

trozet:
Fixes loop variables and formatting so that now make lint and make gofmt pass on 4.10. make check still fails, there are problems with the network policy unit tests.

PR #1154, opened 2022-06-22
[release-4.10] Bug 2100092: Do not use a referance to a range variable.

kyrtapz:
While working on another unit test issue I discovered that we are wrongfully using an address of a range variable.

I think it was introduced during a backport: https://github.com/openshift/ovn-kubernetes/commit/a6956dcf0047f33dca9aa4a5ec6b8ad68a60eba1#diff-1f90f4648dc7d68c42854549c4257ddcfc9bd9a5b6fc480c74e3a23989914bc3R1479 and it does not affect 4.11/master

/assign @jcaamano @flavio-fernandes

### DIFF
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -1363,7 +1363,8 @@ func (oc *Controller) syncNodesPeriodic() {
 	}
 
 	staleChassis := []*sbdb.Chassis{}
-	for _, v := range chassisHostNameMap {
+	for i, _ := range chassisHostNameMap {
+		v := chassisHostNameMap[i]
 		staleChassis = append(staleChassis, &v)
 	}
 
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/master.go
```diff
@@ -1475,6 +1475,7 @@ func (oc *Controller) syncNodesRetriable(nodes []interface{}) error {
 	knownChassisNames := sets.NewString()
 	chassisDeleteList := []*sbdb.Chassis{}
 	for _, chassis := range chassisList {
+		chassis := chassis
```
flavio-fernandes, line 1478: ouch.
This is classic golang kung-fu! :)
afaik, the newer lint will detect issues like this, yes?
Good catch

kyrtapz:
/close
Handled by https://github.com/openshift/ovn-kubernetes/pull/1153

PR #1157, opened 2022-06-22
[release-4.10] Bug 2099843: In tests, expect that unreferenced logical router policies will not be garbage collected

kyrtapz:
libovsdb server [does not remove unreferenced non-root objects](https://github.com/ovn-org/libovsdb/issues/219).
It is expected that the entries will be there in tests and this is not something that could happen in real world.

This workaround is not needed in 4.11 and newer since we are explicitly removing logical router policies without next hops:
https://github.com/openshift/ovn-kubernetes/blob/27d489760ff7e62f025d3aa7452f44b52066a279/go-controller/pkg/libovsdbops/router.go#L352-L359

Since it is not something that will happen in real world deployments I opted for the workaround instead of changing the removal logic at the risk of introducing a regression.

/cc @flavio-fernandes @trozet

### DIFF
--- a/go-controller/pkg/ovn/egressip_test.go
+++ b/go-controller/pkg/ovn/egressip_test.go
@@ -4616,6 +4616,29 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations", func() {
 						Action:   nbdb.LogicalRouterPolicyActionAllow,
 						UUID:     "no-reroute-service-UUID",
 					},
+					// FIXME(kyrtapz): This will never happen during real operations because OVS
+					//   DB server will GC unreferenced non-root items. However
+					//   until https://github.com/ovn-org/libovsdb/issues/219 is
+					//   done, libovsdb test server won't
+					&nbdb.LogicalRouterPolicy{
+						Priority: types.EgressIPReroutePriority,
+						Match:    fmt.Sprintf("ip4.src == %s", egressPod2.Status.PodIP),
+						Action:   nbdb.LogicalRouterPolicyActionReroute,
+						ExternalIDs: map[string]string{
+							"name": eIP.Name,
+						},
+						UUID: "reroute-UUID1",
+					},
+					&nbdb.LogicalRouterPolicy{
+						Priority: types.EgressIPReroutePriority,
+						Match:    fmt.Sprintf("ip4.src == %s", egressPod1.Status.PodIP),
+						Action:   nbdb.LogicalRouterPolicyActionReroute,
+						ExternalIDs: map[string]string{
+							"name": eIP.Name,
+						},
+						UUID: "reroute-UUID2",
+					},
+
 					&nbdb.NAT{
 						UUID:       "egressip-nat-UUID1",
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/egressip_test.go
```diff
@@ -4616,6 +4616,28 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations", func() {
 						Action:   nbdb.LogicalRouterPolicyActionAllow,
 						UUID:     "no-reroute-service-UUID",
 					},
+					// This will never happen during real operations because OVS
```
flavio-fernandes, line 4619: I agree this is ugly but consistent with how we work around issue 219.
SUPER NIT: just like in https://github.com/ovn-org/ovn-kubernetes/pull/2826 ,
Tim added a TODO(trozet) in a similar comment. Maybe we should do it here as well?
Ref: https://github.com/ovn-org/ovn-kubernetes/blob/b1d8cdcaad3eb26016b52fe2510c47666d2d3eff/go-controller/pkg/ovn/policy_test.go#L1591-L1592

flavio-fernandes (commented):
/lgtm

flavio-fernandes:
libovsdb issue linked to this PR:  https://github.com/ovn-org/libovsdb/issues/219

flavio-fernandes (commented):
/lgtm

go-controller/pkg/ovn/egressip_test.go
kyrtapz: I think that not defining `Policies` serves the same purpose, right?

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -4616,6 +4616,29 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations", func() {
...
+					&nbdb.LogicalRouterPolicy{
+						Priority: types.EgressIPReroutePriority,
+						Match:    fmt.Sprintf("ip4.src == %s", egressPod2.Status.PodIP),
+						Action:   nbdb.LogicalRouterPolicyActionReroute,
+						ExternalIDs: map[string]string{
+							"name": eIP.Name,
+						},
+						UUID: "reroute-UUID1",
+					},
+					&nbdb.LogicalRouterPolicy{
```
flavio-fernandes, line 4632: .

go-controller/pkg/ovn/egressip_test.go
flavio-fernandes: oops. ack. please ignore that 🤦

kyrtapz:
/close

this is not needed with https://github.com/openshift/ovn-kubernetes/pull/1228 which brought the removal of empty lrps:

https://github.com/openshift/ovn-kubernetes/pull/1228/commits/159c52bb590ca1e9142cacbd4d8d3ef35527085a#diff-5d62dde13a2856ea3d4b324968dbaeed261fcb74cb3b81909edf40a6634fa1f4R505-R512

PR #1161, opened 2022-06-22
Bug 2100249: Revert "Bug 2082599: add upper bound to number of failed attempts"

trozet:
Reverts openshift/ovn-kubernetes#1147

This fix introduced an upper limit on the number of retries in ovnk for failed attempts. However, due to some compounding issues it causes pods to never come up. The scenario is like this:

1. a pod is added, fails to come up because the node is not ready
2. pod update happens, fails to come up because node is not ready (failed attempts incremented)
3. nodes 1..n come up, all immediately requesting retry of pods
4. backoff is ignored, cause immediate requests were made, the pod is retried n times, all fail, and failed attempts increment, causing the pod to be dropped

These steps happen over a period of 30 seconds, just before the node is finally ready. We need a few adjustments to the logic in ovnk before we can introduce an upper limit.


ricky-rav:
/lgtm

Surya just explained to me what happened with the pods being retried several times at each pod update event and getting all retried upon node updates at startup. Ugh! Sorry for not catching this!

flavio-fernandes:
Looking the failing job makes mt think we may need an `/override ci/prow/e2e-metal-ipi-ovn-ipv6`:

https://testgrid.k8s.io/redhat-openshift-ocp-release-4.11-blocking#periodic-ci-openshift-release-master-nightly-4.11-e2e-metal-ipi-ovn-ipv6

https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-ovn-kubernetes-master-e2e-metal-ipi-ovn-ipv6

@dcbw could you please help us out with that? This is a blocker for 4.11, so we should not hold this up for a known broken job.

PR #1185, opened 2022-07-11
[release-4.10] Bug 2105653: egressIP: node retrieval failure is not respected, causes panic

tssurya:
Fixes panic seen during checkEgressNodesReachability
E0606 11:57:34.063567       1 egressip.go:2096] Node:
huirwang-0606c-m5w24-worker-2-dkjmc reachability changed, but could not
retrieve node from cache, err: node
"huirwang-0606c-m5w24-worker-2-dkjmc" not found
W0606 11:57:34.063658       1 egressip.go:1657] Unable to remove GARP
configuration on external logical switch port for egress node:
huirwang-0606c-m5w24-worker-2-dkjmc, err: object not found
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x176626d]

goroutine 487 [running]:
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn.(*Controller).checkEgressNodesReachability(0xc000171c00)
	/home/surya/go/src/github.com/openshift/ovn-kubernetes/go-controller/pkg/ovn/egressip.go:2099
+0x3ed
created by
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn.(*Controller).initClusterEgressPolicies
	/home/surya/go/src/github.com/openshift/ovn-kubernetes/go-controller/pkg/ovn/egressip.go:1772
+0xdb

The problem here is that the node retrieval fails since the node has
already been deleted, instead of respecting that failure we just
log the error saying node object not found and actually continue
to use that object further in the code which causes this panic.

(cherry picked from commit 36176e5fcc4fb36ce175a02770f84c4556c6ef76)
(cherry picked from commit f436d25441dfbbf6dbd08c4d9535032426be923f)

Conflicts:
go-controller/pkg/ovn/egressip.go
go-controller/pkg/ovn/obj_retry.go

Since https://github.com/ovn-org/ovn-kubernetes/pull/2965 (EIP retry)
is missing in 4.10 and below

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -1444,8 +1444,8 @@ func (oc *Controller) setNodeEgressReachable(nodeName string, isReachable bool) 
 	}
 }
 
-func (oc *Controller) addEgressNode(egressNode *kapi.Node) error {
-	klog.V(5).Infof("Egress node: %s about to be initialized", egressNode.Name)
+func (oc *Controller) addEgressNode(nodeName string) error {
+	klog.V(5).Infof("Egress node: %s about to be initialized", nodeName)
 	// This option will program OVN to start sending GARPs for all external IPS
 	// that the logical switch port has been configured to use. This is
 	// necessary for egress IP because if an egress IP is moved between two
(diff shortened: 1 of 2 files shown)


flavio-fernandes (commented):
Looked at this change and what we did in pr https://github.com/ovn-org/ovn-kubernetes/pull/3019 . 
TY @tssurya !
/lgtm

tssurya:
we seem to be hitting https://bugzilla.redhat.com/show_bug.cgi?id=2106163 in the CI

go-controller/pkg/ovn/egressip.go
```diff
@@ -2017,14 +2017,14 @@ func (oc *Controller) checkEgressNodesReachability() {
 				klog.Errorf("Node: %s reachability changed, but could not retrieve node from cache, err: %v", node.Name, err)
```
trozet, line 2017: this will still cause a panic cause err !=nil and accessing nil node object Name field. We can just remove this entire GetNode since it doesnt look like its being used.

trozet (changes requested):

go-controller/pkg/ovn/egressip.go
tssurya: 100% yes
thanks for the catch, I ported the two commits we talked about here.

tssurya: was a bad rebase actually, seems like my original fix removed this completely and I lost it under conflicts...
https://github.com/ovn-org/ovn-kubernetes/pull/3019/files#r891363448

PR #1204, opened 2022-07-18
Bug 2108317: two hybrid overlay fixes

dcbw:
Cherry-pick of https://github.com/ovn-org/ovn-kubernetes/pull/3063 from upstream.

A missing comma before an "action=" meant that those OVS flows weren't getting added due to a parsing error.

Second, the `k8s.ovn.org/node-gateway-router-lrp-ifaddr` annotation added by https://github.com/ovn-org/ovn-kubernetes/pull/2862 was only evaluated on node add, but we can't always rely on that. It might be added slightly later in an Update event.


dcbw:
Closing in preference to https://github.com/openshift/ovn-kubernetes/pull/1205

PR #1208, opened 2022-07-21
[release-4.7] Bug 2073350: Make sure reject ACL is looked up when it is not in cache

ricky-rav:
The serviceLBMap cache stores, for each load balancer, a service VIP along with its endpoints and reject ACL (if any). The reject ACL is added to this cache:
1. if it was created in the current execution of ovnkube-master or
2. if it was created in a previous execution of ovnkube master and the service still needs a reject ACL (because it has no endpoints).

Now, if the reject ACL was created in the previous execution of ovnkube-master and after restart ovnkube-master first adds the backend pods for this service and only afterwards (re)creates the service along with its endpoints, **then the reject ACL will go unnoticed**. Endpoints will be added correctly, but traffic to this service will be dropped because of the reject ACL.

In order to fix this corner case, after adding endpoints, make sure to lookup reject ACLs in OVN every time the serviceLBMap cache contains no ACL name for any given (load_balancer, VIP) pair.

Fixes #2073350

### DIFF
--- a/go-controller/pkg/ovn/loadbalancer.go
+++ b/go-controller/pkg/ovn/loadbalancer.go
@@ -290,10 +290,9 @@ func (ovn *Controller) createLoadBalancerRejectACL(lb, sourceIP string, sourcePo
 }
 
 func (ovn *Controller) deleteLoadBalancerRejectACL(lb, vip string) {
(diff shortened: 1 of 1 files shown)


trozet (commented):
I don't think this is a good solution. There will be way too many ovn-nbctl calls and be a perf/scale regression.

PR #1211, opened 2022-07-26
[4.7] Bug 2073350: Lookup reject ACLs found at startup when removing reject ACLs for services [alternative]

ricky-rav:
**[alternative solution to https://github.com/openshift/ovn-kubernetes/pull/1208, avoiding extra ovn-nbctl calls]**

The serviceLBMap cache stores, for each load balancer, a service VIP along with its endpoints and reject ACL (if any). The reject ACL is added to this cache either (1) if it was created in the current execution of ovnkube-master or (2) if it was created in a previous execution of ovnkube master and the service still needs a reject ACL (because it has no endpoints).

Now, if the reject ACL was created in the previous execution of ovnkube-master and after restart ovnkube-master first adds the backend pods for this service and only afterwards (re)creates the service, then the existing reject ACL will go unnoticed. Endpoints will be added correctly, but traffic to this service will be dropped because of this reject ACL.

In order to fix this corner case, keep a list of valid reject ACLs found at startup and lookup this list too when trying to delete a reject ACL for a service with endpoints.

closes #2073350

### DIFF
--- a/go-controller/pkg/ovn/service.go
+++ b/go-controller/pkg/ovn/service.go
@@ -3,10 +3,11 @@ package ovn
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"net"
 	"reflect"
 	"strings"
+
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/gateway"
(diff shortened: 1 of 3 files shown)


go-controller/pkg/ovn/service.go
```diff
@@ -204,13 +208,17 @@ func (ovn *Controller) syncServices(services []interface{}) {
 									foundSwitches)
 								ovn.removeACLFromNodeSwitches(foundSwitches, uuid)
 							}
+						} else {
+							OVNRejectACLsAtStartup[name] = uuid
```
trozet, line 212: I don't understand how the sync code above will not catch a service that has endpoints with a stale ACL? What is different about your corner case?

ricky-rav: So the sync code is correct: it removes stale ACLs at startup. What happened with the customer's cluster is that after the sync code and before the service creation the endpoints were added, so the code - after running syncServices - never kept track of the existing reject acl. In particular:

1. the reject ACL was created in the previous execution of ovnkube-master;
2. after restart, the sync code correctly keeps this reject ACL, since the service still has no endpoints
3. the backend pods for this service are added;
4.  the service is (re)created along with its endpoints:
- the service has endpoints, so we do not hit https://github.com/openshift/ovn-kubernetes/blob/b1abcda7c2593c8de9d005ebfd14efe524dcfaf6/go-controller/pkg/ovn/loadbalancer.go#L196, where we create a reject ACL or keep track of an existing one if there are no endpoints;
- we try instead to find an existing reject ACL to remove, but `serviceLBMap` contains no reference to the ACL that's actually in OVN, so the ACL is not removed:
https://github.com/openshift/ovn-kubernetes/blob/release-4.7/go-controller/pkg/ovn/loadbalancer.go#L313-L317

PR #1220, opened 2022-07-29
[release-4.9] Bug 2090315: Remove conntrack entries after rules

tssurya:
When service is deleted, we were removing the conntrack
entries before removing the flows/iptable rules. It's
safer to do the reverse so that in the nano second interim
the conntrack entries don't get recreated before the rules/flows
get deleted.

(cherry picked from commit 58b23ea)
(cherry picked from commit 525e7f1)
(cherry picked from commit bedef57)

Conflicts from 4.10 since ETP isn't present in 4.9


jcaamano:
@tssurya not a cherrypick of https://github.com/openshift/ovn-kubernetes/commit/bedef574d014c46cbd1252f3ed66c7b5a6791c7c as stated on the commit message?

PR #1230, opened 2022-08-02
Egress IP health monitoring over GRPC

flavio-fernandes:
This PR includes a new functionality to egress ip, so it can
use gRPC in order to determine the health of its egress nodes.

Info on the protobuf definition used for the probing is here:
https://github.com/grpc/grpc/blob/master/doc/health-checking.md

The requirement in order to use this method is that both
ovnkube-master and ovnkube-node pods are started with a
new flag: --egressip-node-healthcheck-port <TCP_PORT>.

The original behavior remains as a viable method, should
ovnkube-master pod get started without the new flag.
For sake of completeness, the original method is to
have the ovnkube-master dial to discard (TCP port 9) and
declare node up upon receiving a TCP reject answer.

**NOTE:** this PR will likely not merged; it is expected to be merged as part of the downstream merge that contains the PR:  https://github.com/ovn-org/ovn-kubernetes/pull/3100

Signed-off-by: Flavio Fernandes [flaviof@redhat.com](mailto:flaviof@redhat.com)
Reported-at: https://issues.redhat.com/browse/SDN-3156


flavio-fernandes:
Closing.
Use: https://github.com/openshift/ovn-kubernetes/pull/1263  instead.

PR #1233, opened 2022-08-05
Bug 2115845: [release-4.9] OCP CARRY: Add the rules to EXTERNALIPs only for SGW mode

tssurya:
We call getGatewayIPTRules from both LGW and SGW. This adds
iptable rules for services to EXTERNALIPs chain. We already
add iptable rules for services of type LB.ingress.ip to
NODEPORT chain for LGW. So we need to add the LB ingress rules
to EXTERNALIPs chain only for SGW mode.

/assign @flavio-fernandes @trozet

### DIFF
--- a/go-controller/pkg/node/gateway_iptables.go
+++ b/go-controller/pkg/node/gateway_iptables.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package node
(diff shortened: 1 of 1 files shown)


trozet (approved):
/lgtm

tssurya:
@asood-rh PTAL

PR #1213, opened 2022-07-27
Introduce ovn-kubernetes-{base|singlenode} images

zshi-redhat:
ovn-kubernetes-base image is built from Dockerfile.base
ovn-kubernetes-singlenode image is built from Dockerfile.microshift
ovn-kubernetes-singlenode uses ovn-kubernetes-base as its base image.

PR #1236, opened 2022-08-08
Fix ovn version in Dockerfile.base

zshi-redhat:


zshi-redhat:
New ovn version was not updated while this [PR](https://github.com/openshift/ovn-kubernetes/pull/1213) went in

PR #1241, opened 2022-08-09
Bug 2117078: [release-4.7] Multiple ExGW cache validation/improvements

trozet:
Minor conflicts in the 2nd and 3rd commit

### DIFF
--- a/go-controller/pkg/ovn/egressgw.go
+++ b/go-controller/pkg/ovn/egressgw.go
@@ -166,6 +166,12 @@ func (oc *Controller) addPodExternalGW(pod *kapi.Pod) error {
 
 // addPodExternalGWForNamespace handles adding routes to all pods in that namespace for a pod GW
 func (oc *Controller) addPodExternalGWForNamespace(namespace string, pod *kapi.Pod, egress gatewayInfo) error {
+	tmpPodGWs := oc.getRoutingPodGWs(namespace)
+	tmpPodGWs[pod.Name] = egress
+	if err := validateRoutingPodGWs(tmpPodGWs); err != nil {
+		return fmt.Errorf("unable to add pod: %s/%s as external gateway for namespace: %s, error: %v",
+			pod.Namespace, pod.Name, namespace, err)
+	}
 	nsInfo, nsUnlock, err := oc.ensureNamespaceLocked(namespace, false)
 	if err != nil {
 		return fmt.Errorf("failed to ensure namespace locked: %v", err)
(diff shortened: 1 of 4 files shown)


go-controller/pkg/ovn/egressgw.go
```diff
@@ -166,6 +166,19 @@ func (oc *Controller) addPodExternalGW(pod *kapi.Pod) error {
 
 // addPodExternalGWForNamespace handles adding routes to all pods in that namespace for a pod GW
 func (oc *Controller) addPodExternalGWForNamespace(namespace string, pod *kapi.Pod, egress gatewayInfo) error {
+	var gws string
```
tssurya, line 169: hmmm why are you bringing in L169 to L175 ? we no longer need it since we aren't using `gws` anywhere...

go-controller/pkg/ovn/namespace.go
```diff
@@ -39,6 +39,22 @@ func (oc *Controller) syncNamespaces(namespaces []interface{}) {
 	}
 }
 
+// wrapper function to log if there are duplicate gateway IPs present in the cache
+func validateRoutingPodGWs(podGWs map[string]gatewayInfo) error {
+	// map to hold IP/podName
+	ipTracker := make(map[string]string)
+	for podName, gwInfo := range podGWs {
+		for gwIP := range gwInfo.gws {
+			if foundPod, ok := ipTracker[gwIP]; ok {
```
tssurya, line 48: self note: line different from 4.8 pick since we don't need this to be gwIP.String() since we converted these variables into strings as part of that conntrack backport.

tssurya (changes requested):
COMMIT1: CLEAN PICK LGTM
COMMIT2: PTAL my comment AND should we be bringing in https://github.com/openshift/ovn-kubernetes/commit/988fbb134e6e65f0f354b512b82b7ce496b3aa6f#diff-3696280f7dbfece2e1136d5fa06a3339a05a8a3067710f7510ecdefba51bcb29R194 ? I moved the `parseRoutingExternalGWAnnotation` into utils I see.. but do we need to bring it down?

go-controller/pkg/ovn/egressgw.go
```diff
@@ -257,19 +257,20 @@ func (oc *Controller) deletePodExternalGW(pod *kapi.Pod) {
...
 	nsInfo, nsUnlock := oc.getNamespaceLocked(namespace, false)
 	if nsInfo == nil {
 		return
 	}
+	podGWKey := makePodGWKey(pod)
 	// check if any gateways were stored for this pod
-	foundGws, ok := nsInfo.routingExternalPodGWs[pod]
-	delete(nsInfo.routingExternalPodGWs, pod)
+	foundGws, ok := nsInfo.routingExternalPodGWs[podGWKey]
+	delete(nsInfo.routingExternalPodGWs, podGWKey)
```
tssurya, line 273: assuming conflict was here somewhere? But code change lgtm based on comparision with https://github.com/openshift/ovn-kubernetes/commit/18c8afe2d4cf6adfc6d7a3396b980d2050246a7b.

tssurya (commented):
COMMIT 3 & 4 LGTM

go-controller/pkg/ovn/egressgw.go
trozet: oops removing...

PR #1261, opened 2022-09-01
OCPBUGS-811: [Release 4.9] correct creation and update of egressFirewall 

JacobTanenbaum:
commit 1: 

revert https://github.com/openshift/ovn-kubernetes/commit/02d349323c254c4c106ff65520835946d75f29ce
which causes new egressfirewall ACLs to be created everytime a new master is elected 

commit 2:
correct the way that egressFirewalls are updated. When updating an
egressfirewall rule we need to check the previous rules priority to
figure out if we actually need to delete it.

**- How to verify it**


npinaeva:
/lgtm
tested restart and egress firewall update - works fine

dcbw:
/retest

aws deprovisioning failed, not the actual tests

dcbw:
/override ci/prow/e2e-gcp-ovn

Neither of these failures is due to this PR.

PR #1262, opened 2022-09-06
Use iptables-restore to add MCS/metadata blocking in the pod

dcbw:
Optimize our downstream iptables setup by (1) consolidating all the individual iptables commands into a single iptables-restore, and then (2) running it in parallel with the OVS operations to save sandbox setup time.

@tssurya @trozet @flavio-fernandes @jcaamano @npinaeva


dcbw:
/hold
gotta fix something up

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -14,6 +15,38 @@ import (
 	utilnet "k8s.io/utils/net"
 )
 
+const baseRestoreData string = `# Generated by ovnkube-node
+*filter
+:INPUT ACCEPT [0:0]
+:FORWARD ACCEPT [0:0]
+:OUTPUT ACCEPT [0:0]
+# Block Machine Config Service
+-A FORWARD -p tcp -m tcp --dport 22623 --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable
```
flavio-fernandes, line 24: I'm a little confused because the `--reject-with icmp-port-unreachable` was not part of the original code, but I get it now!

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -26,47 +59,23 @@ func setupIPTablesBlocks(netns ns.NetNS, ifInfo *PodInterfaceInfo) error {
...
-				if err != nil {
-					return fmt.Errorf("could not set up pod iptables rules: %s", string(out))
-				}
-			}
-			if hasIPv6 {
-				out, err := exec.Command("ip6tables", args...).CombinedOutput()
-				if err != nil {
-					return fmt.Errorf("could not set up pod iptables rules: %s", string(out))
-				}
+		args := []string{"--noflush", "--counters", "-w", "5"}
```
flavio-fernandes, line 62: `--noflush` plays an important factor here, so this behaves as additional changes like original iptables command would do it.

go-controller/pkg/cni/OCP_HACKS.go
```diff
@@ -14,6 +15,38 @@ import (
...
+:OUTPUT ACCEPT [0:0]
+# Block Machine Config Service
+-A FORWARD -p tcp -m tcp --dport 22623 --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable
+-A FORWARD -p tcp -m tcp --dport 22624 --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable
+-A OUTPUT -p tcp -m tcp --dport 22623 --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable
+-A OUTPUT -p tcp -m tcp --dport 22624 --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable
+`
+
+func generateRestoreData(isIPv4 bool) string {
+	metadataServiceIP := "169.254.169.254"
```
flavio-fernandes, line 31: SUPER NIT: metadataServiceIP could be moved into the `isIPv4` block, since ipv6 has no use for it.

flavio-fernandes (commented):
/lgtm
cool beans!

go-controller/pkg/cni/OCP_HACKS.go
dcbw: @flavio-fernandes I think it's the default actually. I ran a cluster, then did `iptables-save` in a pod namespace to get the input data here.

go-controller/pkg/cni/OCP_HACKS.go
dcbw: @flavio-fernandes yeah, I'll do that. I was thinking that eventually we get v6 here but that's not for a while.

go-controller/pkg/cni/OCP_HACKS.go
dcbw: Yeah; also it only runs once at sandbox setup and there won't be any existing rules anyway. So I guess `--noflush` doesn't really matter anyway.

flavio-fernandes (commented):
/lgtm

PR #1263, opened 2022-09-06
[DownstreamMerge] 9-2-22 merge

trozet:
Merge conflicts around ICNIv1 hybrid overlay code. To resolve:
1. Moved add/update pod ICNIv1 logic into the core controller generic pod handler
2. Moved add namespace ICNIv1 logic into the namespace handler
3. Eliminated the waitForNamespaces function that the pod handler path in ICNIv1 used to use. There is no reason for this because we now have retry to so if pod handler fails to get the namespace it will be retried later.
4. Updated the test cases specific to ICNIv1

Added an extra commit to avoid removing the pod default route to DR (.3) IP for ICNIv1 pods.


trozet:
@JacobTanenbaum @dcbw PTAL

@flavio-fernandes

trozet:
@JacobTanenbaum I added  https://github.com/openshift/ovn-kubernetes/pull/1263/commits/6bb044bc41cd187880070f02502fe3f7bb4118dd

Let me know if it looks right to you.

flavio-fernandes:
@flavio-fernandes [selfishly] looked at the Egress IP gRPC changes.

/lgtm

PR #1259, opened 2022-08-31
[reelase-4.11] OCPBUGS-772: Fix NP upgrade bug: unexpectedly found multiple equivalent ACLs (arp v/s arp||nd)

tssurya:
Minor conflict in commit 1 on the e2e file since https://github.com/openshift/ovn-kubernetes/commit/e1f7da0fe5072ca8574f0aa98fa0c80ff2c95a92 is missing in 4.11

PR #1281, opened 2022-09-27
OCPBUGS-1705: Trim ACL names according to RFC1123

tssurya:
We didn't consider this tiny hack that we do
https://github.com/openshift/ovn-kubernetes/blob/44ad75466e486cce605e39513a3ecd9e0b306e7d/go-controller/pkg/libovsdbops/acl.go#L60 there when we wrote https://github.com/openshift/ovn-kubernetes/pull/1259 and unit tests don't actually scream loudly for longer names.

Without this PR we break users who have namespace names longer than 45 characters.

Co-authored-by: Patryk Diak <pdiak@redhat.com>
(cherry picked from commit b10ed7bcc2fe486e6cb8b1a06024227b06abaa71)

/cc @trozet and @jcaamano

### DIFF
--- a/go-controller/pkg/ovn/policy.go
+++ b/go-controller/pkg/ovn/policy.go
@@ -115,7 +115,7 @@ func (oc *Controller) updateStaleDefaultDenyACLNames(npType knet.PolicyType, gre
 		return item.ExternalIDs[defaultDenyPolicyTypeACLExtIdKey] == string(npType) && // default-deny-policy-type:Egress or default-deny-policy-type:Ingress
 			strings.Contains(item.Match, gressSuffix) && // Match:inport ==	@ablah80448_egressDefaultDeny or Match:inport == @ablah80448_ingressDefaultDeny
 			!strings.Contains(*item.Name, arpAllowPolicySuffix) && // != name: namespace_ARPallowPolicy
-			!strings.Contains(*item.Name, gressSuffix) // filter out already converted ACLs
+			!strings.Contains(*item.Name, gressSuffix) // filter out already converted ACLs (we still continue to pick the cases where names were >45 chars)
 	}
 	gressACLs, err := libovsdbops.FindACLsWithPredicate(oc.nbClient, p)
 	if err != nil {
@@ -138,8 +138,17 @@ func (oc *Controller) updateStaleDefaultDenyACLNames(npType knet.PolicyType, gre
 				return err
 			}
 		}
-		aclList[0].Name = &newName
-		err := libovsdbops.CreateOrUpdateACLs(oc.nbClient, aclList[0])
+		newACL := BuildACL(
+			newName, // this is the only thing we need to change, keep the rest same
+			aclList[0].Direction,
+			aclList[0].Priority,
+			aclList[0].Match,
+			aclList[0].Action,
+			oc.GetNamespaceACLLogging(namespace),
+			aclList[0].ExternalIDs,
+			aclList[0].Options,
+		)
+		err := libovsdbops.CreateOrUpdateACLs(oc.nbClient, newACL)
 		if err != nil {
 			return fmt.Errorf("cannot update old NetworkPolicy ACLs for namespace %s: %v", namespace, err)
 		}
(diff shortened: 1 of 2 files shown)


trozet (approved):
/lgtm

jcaamano:
Well, it looks like the problem we were having is actually saving us from bigger problems.

The whole logic of `updateStaleDefaultDenyACLNames` assumes that we have full un-cropped ACLs names:

```
	p := func(item *nbdb.ACL) bool {
		return item.ExternalIDs[defaultDenyPolicyTypeACLExtIdKey] == string(npType) && // default-deny-policy-type:Egress or default-deny-policy-type:Ingress
			strings.Contains(item.Match, gressSuffix) && // Match:inport ==	@ablah80448_egressDefaultDeny or Match:inport == @ablah80448_ingressDefaultDeny
			!strings.Contains(*item.Name, arpAllowPolicySuffix) && // != name: namespace_ARPallowPolicy
			!strings.Contains(*item.Name, gressSuffix) // filter out already converted ACLs (we still continue to pick the cases where names were >45 chars)
	}
```

We need to assume that the name might be cropped.
A more proper predicate would check if the name ends (portion after the `_`) with a substring of the policy name.

Additionally (not directly relevant to this PR), I think that we should short-term (and to be back-portable):
- stop using the name (and external ids) in `isEquivalentACL`
- stop using DeleteACLs (ACLs should always be removed via garbage collection after removing them from PGs and switches, and never directly)
cc @npinaeva I don't know if your ACL work is scoped in 4.12 but we should get this done for 4.12 at least and probably backport it. Let me know your plans!

The units tests don't actually account for a long namespace name :/

npinaeva:
Oh I didn't know we are testing long namespaces, there should be more that 1 bug for that case.
My plan was to get rid of `isEquivalentACL` which is a big change, I don't know yet if I need to backport it
I can try to make some immediate changes to prevent cropped name bugs, need to investigate some aspects
About deleting ACLs, I think in general it's better to be explicit that ACLs need to be deleted, and not just garbage-collected. I need to check if it will throw an error if ACL I am trying to delete still has a reference? If not, there is not much difference

jcaamano:
> Oh I didn't know we are testing long namespaces, there should be more that 1 bug for that case. My plan was to get rid of `isEquivalentACL` which is a big change, I don't know yet if I need to backport it I can try to make some immediate changes to prevent cropped name bugs, need to investigate some aspects About deleting ACLs, I think in general it's better to be explicit that ACLs need to be deleted, and not just garbage-collected. I need to check if it will throw an error if ACL I am trying to delete still has a reference? If not, there is not much difference

They will throw an error since they are strong references. And we are currently not able to verify if an ACL is referenced in more than one place. And an efficient implementation would probably reuse ACLs in different places if possible but keeping track of where would be complex from a client perspective, making this garbage collection a meaningful server feature to have and rely on.
Anyway, if we are able to verify ACLs are only referenced once, I would still not export any DeleteACLs method from libovsdbops and handle it internally, even if explicit.

jcaamano:
> My plan was to get rid of `isEquivalentACL` 

For 4.12?

jcaamano:
> ovn is taking care of that, it will create only 1 logical_flow for equivalent ACLs

This is not the same thing. Less NB DB space consumed and less northd processing time.

> What kind of interface do you mean here? I still would expect to have DeleteACL if I have CreateACL function?

You can expect it, but you don't necessarily need to have it. Or we can have it and panic with the appropriate explanation so that no one is tempted to add it.

> I don't have a timeline yet, it wasn't a high priority issue :( We may need to discuss if it is safe and makes sense to merge that to 4.12

No, in that case I think we should have a separate bug that specifically targets this issue and can be assigned.

tssurya:
> 
> > I don't have a timeline yet, it wasn't a high priority issue :( We may need to discuss if it is safe and makes sense to merge that to 4.12
> 
> No, in that case I think we should have a separate bug that specifically targets this issue and can be assigned.

I have opened https://issues.redhat.com/browse/OCPBUGS-1958 -> for the bug fix we need to take the approach we decided which is either remove the name part or fix the externalID part to be unique.

The long term fix which is a bit different needs to be discussed in the team meeting.

npinaeva:
I am working on a long term fix for at least a month already: https://issues.redhat.com/browse/SDN-3447
If we need a faster bug fix, I would prefer this bug to be a part of the same effort ^, so we don't develop 2 different ways to fix the same problem.

tssurya:
> I am working on a long term fix for at least a month already: https://issues.redhat.com/browse/SDN-3447 If we need a faster bug fix, I would prefer this bug to be a part of the same effort ^, so we don't develop 2 different ways to fix the same problem.

We can't backport the externalIDs fix and this current bug needs to be backport, the externalIDs effort around telling who is the CMS is different from this bug. Thanks for creating the JIRA card! The bug needs to be fixed nevertheless.

PR #1282, opened 2022-09-27
OCPBUGS-1750: Trim ACL names according to RFC1123

tssurya:
Cherry picks https://github.com/openshift/ovn-kubernetes/pull/1281 and https://github.com/openshift/ovn-kubernetes/commit/c20451998bcc62ef5281d1316110ed44219386b9
First cherry-pick had to be modified to use `libovsdbops.BuildACL` instead of `ovn.BuildACL` since we are missing https://github.com/openshift/ovn-kubernetes/commit/44ad75466e486cce605e39513a3ecd9e0b306e7d in 4.11
Second cherry-pick had a very minor conflict outlined in the commit message.

### DIFF
--- a/go-controller/pkg/libovsdbops/acl.go
+++ b/go-controller/pkg/libovsdbops/acl.go
@@ -2,6 +2,7 @@ package libovsdbops
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 
@@ -159,8 +160,24 @@ func UpdateACLsDirection(nbClient libovsdbclient.Client, acls ...*nbdb.ACL) erro
 	return err
 }
 
-// DeleteACLs deletes the provided ACLs
-func DeleteACLs(nbClient libovsdbclient.Client, acls ...*nbdb.ACL) error {
+// DeleteACLsOps deletes the provided ACLs and returns the corresponding ops
+// portGroupNames and switchPred reminds to delete ACL references for port groups or switches
+// in case port group or switch is not completely deleted
+func DeleteACLsOps(nbClient libovsdbclient.Client, ops []libovsdb.Operation,
+	portGroupNames []string, switchPred switchPredicate, acls ...*nbdb.ACL) ([]libovsdb.Operation, error) {
+	var err error
+	for _, portGroupName := range portGroupNames {
+		ops, err = DeleteACLsFromPortGroupOps(nbClient, ops, portGroupName, acls...)
+		if err != nil && !errors.Is(err, libovsdbclient.ErrNotFound) {
+			return ops, fmt.Errorf("deleting ACLs from port group %s failed: %v", portGroupName, err)
+		}
+	}
+	if switchPred != nil {
+		ops, err = RemoveACLsFromLogicalSwitchesWithPredicateOps(nbClient, ops, switchPred, acls...)
+		if err != nil && !errors.Is(err, libovsdbclient.ErrNotFound) {
+			return ops, fmt.Errorf("deleting ACLs from switch with predicate failed: %v", err)
+		}
+	}
 	opModels := make([]operationModel, 0, len(acls))
 	for i := range acls {
 		// can't use i in the predicate, for loop replaces it in-memory
@@ -175,5 +192,18 @@ func DeleteACLs(nbClient libovsdbclient.Client, acls ...*nbdb.ACL) error {
 	}
 
 	modelClient := newModelClient(nbClient)
-	return modelClient.Delete(opModels...)
+	return modelClient.DeleteOps(ops, opModels...)
+}
+
+// DeleteACLs deletes the provided ACLs
+// portGroupNames and switchPred reminds to delete ACL references for port groups or switches
+// in case port group or switch is not completely deleted
+func DeleteACLs(nbClient libovsdbclient.Client, portGroupNames []string, switchPred switchPredicate, acls ...*nbdb.ACL) error {
+	ops, err := DeleteACLsOps(nbClient, nil, portGroupNames, switchPred, acls...)
+	if err != nil {
+		return err
+	}
+
+	_, err = TransactAndCheck(nbClient, ops)
+	return err
 }
--- a/go-controller/pkg/libovsdbops/portgroup.go
+++ b/go-controller/pkg/libovsdbops/portgroup.go
@@ -1,10 +1,25 @@
 package libovsdbops
 
 import (
+	"context"
+
 	libovsdbclient "github.com/ovn-org/libovsdb/client"
 	libovsdb "github.com/ovn-org/libovsdb/ovsdb"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 )
+
+type portGroupPredicate func(group *nbdb.PortGroup) bool
+
+// FindPortGroupsWithPredicate looks up port groups from the cache based on a
+// given predicate
+func FindPortGroupsWithPredicate(nbClient libovsdbclient.Client, p portGroupPredicate) ([]*nbdb.PortGroup, error) {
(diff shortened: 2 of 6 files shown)


jcaamano:
@tssurya Why do we need `https://github.com/openshift/ovn-kubernetes/commit/c20451998bcc62ef5281d1316110ed44219386b9`? 
It doesn't really do much other than accommodating for the limitation of the libovsdb test server that does not handle referential integrity. So it should only be relevant for unit tests.

tssurya:
> @tssurya Why do we need `https://github.com/openshift/ovn-kubernetes/commit/c20451998bcc62ef5281d1316110ed44219386b9`? It doesn't really do much other than accommodating for the limitation of the libovsdb test server that does not handle referential integrity. So it should only be relevant for unit tests.

There is a place in the code: https://github.com/openshift/ovn-kubernetes/pull/1282/files#diff-cc83e19af1c257d5a09b711d5977d8f8c20beb34b7b5d3eb37b2f2c53ded1bf7L143 where we were calling `DeleteACLs` directly which would complain against the referential integrity violation. Nadia's PR will help with that and keep things consistent. SInce 4.11 will be around for some time, I don't want people stepping on the DeleteACLs function and forgetting to individually remove things from PGs and that PR adds that part: https://github.com/openshift/ovn-kubernetes/pull/1282/files#diff-ee54e95b45c5a079454546fed94fcef68b13d9dc6cd14e192585fe2465bcaefcR170 whereby calling `DeleteACLs` will auto-delete the the acls from PGs and switches.

tssurya:
> So it should only be relevant for unit tests.

Not sure if I follow this part, any place in code like efw or policies that call `DeleteACLs` - this will be relevant for all that right?

jcaamano:
> > @tssurya Why do we need `https://github.com/openshift/ovn-kubernetes/commit/c20451998bcc62ef5281d1316110ed44219386b9`? It doesn't really do much other than accommodating for the limitation of the libovsdb test server that does not handle referential integrity. So it should only be relevant for unit tests.
> 
> There is a place in the code: https://github.com/openshift/ovn-kubernetes/pull/1282/files#diff-cc83e19af1c257d5a09b711d5977d8f8c20beb34b7b5d3eb37b2f2c53ded1bf7L143 where we were calling `DeleteACLs` directly which would complain against the referential integrity violation. Nadia's PR will help with that and keep things consistent. SInce 4.11 will be around for some time, I don't want people stepping on the DeleteACLs function and forgetting to individually remove things from PGs and that PR adds that part: https://github.com/openshift/ovn-kubernetes/pull/1282/files#diff-ee54e95b45c5a079454546fed94fcef68b13d9dc6cd14e192585fe2465bcaefcR170 whereby calling `DeleteACLs` will auto-delete the the acls from PGs and switches.

Ok. So comments about this:

- That was because using `DeleteACLs` there is not the best thing to do. The call to use should be `DeleteACLsFromPortGroup` directly. The server will garbage-collect the ACLs once they are removed form the port groups. On some places we would call `DeleteACLs` after `DeleteACLsFromPortGroup` to counter the fact that the test server does not do the garbage collection; on other places we would account for the leftover ACLs in the tests. Arguably the other best alternative would have been to remove the ACLs inside `DeleteACLsFromPortGroup` instead of the other way around; and not having a `DeleteACLs` at all to avoid the confusion. Anyway we will live with it for now.
- This seems like separate fix. Even if we use this PR, should we track it with a different bug so that people that encounters the issue is aware that this is now fixed?

jcaamano:
How difficult would be to cherry pick `ovn.BuildACL` as well or how probable it is that is cherry-picked in the future?
The reason I ask is because if `ovn.BuildACL` is back-ported in the future, most likely the call to `libovsdbops.BuildACL` that is done here will be missed and not changed to `ovn.BuildACL`

tssurya:
> > > @tssurya Why do we need `https://github.com/openshift/ovn-kubernetes/commit/c20451998bcc62ef5281d1316110ed44219386b9`? It doesn't really do much other than accommodating for the limitation of the libovsdb test server that does not handle referential integrity. So it should only be relevant for unit tests.
> > 
> > 
> > There is a place in the code: https://github.com/openshift/ovn-kubernetes/pull/1282/files#diff-cc83e19af1c257d5a09b711d5977d8f8c20beb34b7b5d3eb37b2f2c53ded1bf7L143 where we were calling `DeleteACLs` directly which would complain against the referential integrity violation. Nadia's PR will help with that and keep things consistent. SInce 4.11 will be around for some time, I don't want people stepping on the DeleteACLs function and forgetting to individually remove things from PGs and that PR adds that part: https://github.com/openshift/ovn-kubernetes/pull/1282/files#diff-ee54e95b45c5a079454546fed94fcef68b13d9dc6cd14e192585fe2465bcaefcR170 whereby calling `DeleteACLs` will auto-delete the the acls from PGs and switches.
> 
> Ok. So comments about this:
> 
> * That was because using `DeleteACLs` there is not the best thing to do. The call to use should be `DeleteACLsFromPortGroup` directly. The server will garbage-collect the ACLs once they are removed form the port groups. On some places we would call `DeleteACLs` after `DeleteACLsFromPortGroup` to counter the fact that the test server does not do the garbage collection; on other places we would account for the leftover ACLs in the tests. Arguably the other best alternative would have been to remove the ACLs inside `DeleteACLsFromPortGroup` instead of the other way around; and not having a `DeleteACLs` at all to avoid the confusion. Anyway we will live with it for now.

Exactly, like I have brought it up in the team meetings, this was an implicit thing only libovsdb refactoring folks would know about. At the very least there should be a developer docs that outline things like always use `ovn.BuildACLs` and not `libovsdbops.BuildACLs` OR always use`DeleteACLsFromPortGroup` and not `DeleteACLs` which is only for testing and garbage collection, another option is to add them to the function explanation but I doubt people will see that. 

> * This seems like separate fix. Even if we use this PR, should we track it with a different bug so that people that encounters the issue is aware that this is now fixed?

The same bug is tracking both the issues. It's already on the watch list so doesn't matter at this point. If you want a new bug I can do that as well it will just trigger a new series of backports downwards till 4.9 that's doubling the work I need to get all the labels for. before Tuesday :(

tssurya:
> How difficult would be to cherry pick `ovn.BuildACL` as well or how probable it is that is cherry-picked in the future? The reason I ask is because if `ovn.BuildACL` is back-ported in the future, most likely the call to `libovsdbops.BuildACL` that is done here will be missed and not changed to `ovn.BuildACL`

I tried to do this actually to bring in the new `ovn/acl.go` file but I didn't need everything in it, and I saw Nadia is again refactoring those parts in a another PR that you are reviewing where the position of ovn.BuildACL jumps up in the ovn/acl.go. I didn't want to make things hard so it can be done in another PR when those bits are stable in master first.

jcaamano:
> I tried to do this actually to bring in the new `ovn/acl.go` file but I didn't need everything in it, and I saw Nadia is again refactoring those parts in a another PR that you are reviewing where the position of ovn.BuildACL jumps up in the ovn/acl.go. I didn't want to make things hard so it can be done in another PR when those bits are stable in master first.

OK, I think we should just remove `libovsdbops.BuildACL` actually. 
cc @npinaeva

abhat:
cc @anuragthehatter, @rbbratta for the cherry-pick labels and getting the dependent bug verified.

anuragthehatter:
> cc @anuragthehatter, @rbbratta for the cherry-pick labels and getting the dependent bug verified.

I am QE'ing this right now. Will update soon

anuragthehatter:
This fix seems to be introducing a regression as updated in the bug

jcaamano:
> This fix seems to be introducing a regression as updated in the bug

Added comment about  it ion the original PR https://github.com/openshift/ovn-kubernetes/pull/1281#issuecomment-1265272739

anuragthehatter:
QE testing passed on this PR covering all corner cases

/label qe-approved

PR #1302, opened 2022-10-11
OCPBUGS-2176: add endpointSlice informer in master process

zshi-redhat:
In MicroShift, ovnkube-master and ovnkube-node run in a single process/container
which acts as both master and node roles. The NewMasterWatchFactory needs to
initialize a new watch factory for both, otherwise it results in panic watching
endpointSlice which only existed in node watch factory.

This PR creates endpointSlices in master watch factory even it is not used when
master and node are deployed separately, this will be fixed by upstream PR:
https://github.com/ovn-org/ovn-kubernetes/pull/3193.

(cherry picked from commit b6c1b081fb670423a53febc1e62ea9db89abdb63)

### DIFF
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -152,6 +152,15 @@ func NewMasterWatchFactory(ovnClientset *util.OVNClientset) (*WatchFactory, erro
 			noAlternateProxySelector())
 	})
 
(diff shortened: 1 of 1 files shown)


zshi-redhat:
@jcaamano @trozet PTAL.

jcaamano:
@zshi-redhat can you explain a bit why we need this in the commit message or PR message or the bug? There is no info anywhere. Might be obvious for some but not all.

zshi-redhat:
> @zshi-redhat can you explain a bit why we need this in the commit message or PR message or the bug? There is no info anywhere. Might be obvious for some but not all.

@jcaamano sure, I updated the PR message.

PR #1295, opened 2022-10-06
OCPBUGS-1705: [Downstream Merge] - 06-10-2022

tssurya:
Conflicts:
    1) go-controller/pkg/ovn/hybrid_test.go
    Conflict in the import library since we need k8stypes, fmt and runtime
    for the downstream only unit tests.
Screenshot from 2022-10-06 15-09-19

    2) go-controller/pkg/ovn/ovn.go
    Conflict in type namespaceInfo struct positioning - had to keep the hybridoverlay variables in nsInfo

Screenshot from 2022-10-06 15-02-37

    3) go-controller/pkg/ovn/pods.go
    Confict in addLogicalPort around
    https://github.com/ovn-org/ovn-kubernetes/pull/3048/commits/6f6d134f04ac2a9ff9b44ccebcbb0bceb579c2c5#diff-9446f38c60c510407e435ea73659254783cc86dcffd3b05debf078144d430f13L577
    because we seem to have different positions where we call
    oc.addPodToNamespace(pod.Namespace, podIfAddrs) upstream versus
    downstream. This is not related to this PR; seems the difference
    has existed for a long time.

Screenshot from 2022-10-06 14-46-18

PR #1304, opened 2022-10-11
OCPBUGS-2004: egress IP: fix log when gRPC connection fails

flavio-fernandes:
Connecting logic iterates through all the management addresses that egress IP can use to reach its egress nodes. The log, however, was only providing the last one it tried.

Reported-at: https://issues.redhat.com/browse/OCPBUGS-2004


flavio-fernandes:
/assign @jcaamano 

This was part of  https://github.com/openshift/ovn-kubernetes/pull/1295 but that PR got closed and we will need to cherry pick individual fixes like this.

PR #1330, opened 2022-10-21
OCPBUGS-1427: Ignore non-ready endpoints when processing endpointslices

ricky-rav:
Ignore non-ready endpoints when processing endpointslices

In ovn-k node we introduced a regression when replacing endpoints with endpointslices, causing 10x higher disruption times in ingress-to-console/oauth/image-registry new connections in CI upgrade jobs, as reported in #OCPBUGS-1427

Previously we were processing Endpoints from `k8s.io/api/core/v1`, where we get the current list of _ready_ endpoint addresses from  `EndpointSubset.Addresses` and the list of non-ready addresses from `EndpointSubset.NotReadyAddresses`:  https://github.com/openshift/ovn-kubernetes/blob/da6ac5b78ff34d10201edbadb37e7b658b3ea991/go-controller/vendor/k8s.io/api/core/v1/types.go#L4711-L4724

However, endpointslices are defined differently and offer a more fine-grained view of each endpoint, reporting its status with respect to the endpoint being `ready`, `serving` and `terminating`: https://github.com/openshift/ovn-kubernetes/blob/ac9909dadf452a27153ee7f983f935a1992f365b/go-controller/vendor/k8s.io/api/discovery/v1/types.go#L117-L141

 All endpoints inside an endpointslice, regardless of their current status, are in the same slice:
 https://github.com/openshift/ovn-kubernetes/blob/ac9909dadf452a27153ee7f983f935a1992f365b/go-controller/vendor/k8s.io/api/discovery/v1/types.go#L43-L46

Our code didn't filter out non-ready endpoints, which explains why we were getting 10x the disruption times we previously had: throughout the upgrade, we were always considering all endpoints as being ready, while they were not.

Signed-off-by: Riccardo Ravaioli [rravaiol@redhat.com](mailto:rravaiol@redhat.com)

Fixes #OCPBUGS-1427

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -708,7 +708,11 @@ func (npw *nodePortWatcher) AddEndpointSlice(epSlice *discovery.EndpointSlice) {
 	// Here we make sure the correct rules are programmed whenever an AddEndpointSlice
 	// event is received, only alter flows if we need to, i.e if cache wasn't
 	// set or if it was and hasLocalHostNetworkEp state changed, to prevent flow churn
-	namespacedName := namespacedNameFromEPSlice(epSlice)
+	namespacedName, err := namespacedNameFromEPSlice(epSlice)
+	if err != nil {
+		klog.Errorf("Skipping %s/%s: %v", epSlice.Namespace, epSlice.Name, err)
+		return
+	}
 	out, exists := npw.getAndSetServiceInfo(namespacedName, svc, hasLocalHostNetworkEp)
 	if !exists {
 		klog.V(5).Infof("Endpointslice %s ADD event in namespace %s is creating rules", epSlice.Name, epSlice.Namespace)
@@ -738,7 +742,11 @@ func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice
 
 	klog.V(5).Infof("Deleting endpointslice %s in namespace %s", epSlice.Name, epSlice.Namespace)
 	// remove rules for endpoints and add back normal ones
-	namespacedName := namespacedNameFromEPSlice(epSlice)
+	namespacedName, err := namespacedNameFromEPSlice(epSlice)
+	if err != nil {
+		klog.Errorf("Skipping %s/%s: %v", epSlice.Namespace, epSlice.Name, err)
+		return
+	}
 	if svcConfig, exists := npw.updateServiceInfo(namespacedName, nil, &hasLocalHostNetworkEp); exists {
 		// Lock the cache mutex here so we don't miss a service delete during an endpoint delete
 		// we have to do this because deleting and adding iptables rules is slow.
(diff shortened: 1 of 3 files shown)


go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -657,6 +657,7 @@ func (npw *nodePortWatcher) SyncServices(services []interface{}) error {
 				}
 
 				for _, ep := range epSlice.Endpoints {
+					// TODO ask Ori: should we only select ready endpoints?
```
ricky-rav, line 660: @oribon In this PR I'm fixing our usage of endpoint slices. I'm not familiar with the part on egress services that you wrote. Do you think that here (and in the following lines where I'm going to tag you) we need to filter out endpoints that are not ready or is it the way it is? Thanks a lot!

go-controller/pkg/node/gateway_shared_intf_linux.go
```diff
@@ -108,6 +108,7 @@ func updateEgressSVCIptRules(svc *kapi.Service, npw *nodePortWatcher) {
 		}
 
 		for _, ep := range epSlice.Endpoints {
+			// TODO ask Ori: should we only select ready endpoints?
```
ricky-rav, line 111: @oribon Here too, could you have a look? Thanks!

go-controller/pkg/ovn/controller/egress_services/egress_services_service.go
```diff
@@ -407,6 +407,7 @@ func (c *Controller) allEndpointsFor(svc *corev1.Service) (sets.String, sets.Str
 		}
 
 		for _, ep := range eps.Endpoints {
+			// TODO Ask Ori if non-ready endpoints should be skipped
```
ricky-rav, line 410: @oribon and finally here. Thanks!

ricky-rav:
With @trozet and @oribon we decided not to change the part about egress services that I brought up in a previous commit, since it wouldn't bring any considerable benefit, as it affects traffic _originating from_ endpoints, while this PR fixes the endpoint selection for traffic _destined to_ endpoints.

trozet (approved):
@arkadeepsen fyi @ricky-rav omitted using ParseIPSloppy because: https://github.com/ovn-org/ovn-kubernetes/pull/3138/commits/dbe0a9137698088b0516acdafba9e38a9853d699 
is not downstream, just a heads up if you backport that to make sure you call that function on the places affected in this PR:
node.go -> WatchEndpointSlices -> UpdateFunc
gateway_shared_intf.go -> getEndpointAddresses

/lgtm

dcbw:

metal jobs currently do not bootstrap

dcbw:
/hold

To stop CI retesting so we can get a fresh namespace and solve the A-record error that the metal jobs are running into.

dcbw:
/hold cancel
CI namespaces long since deleted by now

dcbw:
/override ci/prow/4.12-upgrade-from-stable-4.11-local-gateway-e2e-aws-ovn-upgrade
https://coreos.slack.com/archives/C01CQA76KMX/p1665427374304029

PR #1333, opened 2022-10-21
OCPBUGS-1520: Prioritize adding events to handlers for shared resources

tssurya:
Prioritize adding events to handlers for shared resources
    
    When handlers share the informer for the same type of objects - example
    egressIP feature also watches for Pods and creates a handler to receive
    events for any pod-related activities - it is better to prioritize the
    order in which a specific event for a pod will be processed. Pod
    creation should happen first, so addLogicalPort should go first before
    egressIP Pod handler. Similarly for network policies and other
    controllers.
    
    Today OVNK already handles the events serially, the only problem is we
    don't control which handler gets the event first, so this PR adds a
    priority system for the handlers where by we say all critical handlers
    get priority 0 - this is the highest priority and it applies to
    pods/namespaces/nodes and all the CRD objects. Then we start handing out
    priority to rest of the handlers, example egressIP pod handler will have
    priority 1 which means it will always get the event after the main pod
    handler has got it. Same for egressIP namespace handler which will
    always get the event after the main namespace handler has got it.
    
    (cherry picked from commit c0f32f952c20bc4c9b90b50624b7303bbf65a8f2)

### DIFF
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -78,6 +78,11 @@ const (
 
 	// namespace, node, and pod handlers
 	defaultNumEventQueues uint32 = 15
+
+	// default priorities for various handlers (also the highest priority)
+	defaultHandlerPriority uint32 = 0
+	// lowest priority among various handlers (See GetHandlerPriority for more information)
+	minHandlerPriority uint32 = 4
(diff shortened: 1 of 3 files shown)


trozet (approved):
/lgtm

PR #1341, opened 2022-10-25
[release-4.11] OCPBUGS-2859: ovn node, gw router: set node masquerade IP on br-ex

kyrtapz:
This PR is required in 4.11 to allow for rollbacks from 4.12
Backport of https://github.com/ovn-org/ovn-kubernetes/pull/3136

Commits with conflicts:
- https://github.com/openshift/ovn-kubernetes/pull/1341/commits/4e2baa4422c648787541d409e0d53e4af0c811d3
- https://github.com/openshift/ovn-kubernetes/pull/1341/commits/d974c9d38e03fc1478527912e1e895635ac2901d


andreaskaris:
I have a feeling that https://github.com/openshift/ovn-kubernetes/commit/f98a5c071d897bcf4adf45ed1211ddc9a4a65b15 is causing a few issues:
https://issues.redhat.com/browse/OCPBUGS-2512 (fixed in latest)

I also have the following 2 which might be the same issue:
* https://issues.redhat.com/browse/OCPBUGS-2817 IPv6 masquerade route with new “src” addition causing issues on IP address change
* https://issues.redhat.com/browse/OCPBUGS-3394  “could not find node IPv6 address” --> might be an issue when SLAAC adds an IPv6 address in dual stack after a few minutes

cybertron:
Based on my latest investigation I think there's a possibility that this is impacting the metal ipv6 upgrade jobs too: https://issues.redhat.com/browse/OCPBUGS-2319

There seems to be a routing issue with traffic to the API VIP during the upgrade and while I'm not positive it's caused by this I've seen some suspicious stuff with the traffic appearing to come from fd69 instead of the regular host address.

kyrtapz:
It seems that https://github.com/ovn-org/ovn-kubernetes/pull/3246 is causing some issues in the latest release.

Closing this PR in favor of https://github.com/openshift/ovn-kubernetes/pull/1366 that only brings the required commit without introducing any flow changes.
If we decide to backport the rest of https://github.com/ovn-org/ovn-kubernetes/pull/3246 it should be fairly straight forward.

PR #1351, opened 2022-11-02
[release-4.12] OCPBUGS-3348: Validate node subnet annotations against cluster networks

dcbw:
If a node is added to the cluster after the master is running and the node already has a subnet annotation, ensure we validate that:

(a) the node's annotated subnets are within the cluster networks
(b) the node's annotated subnets are not already claimed by another node

Also post a node Event when something fails.

Backport of upstream https://github.com/ovn-org/ovn-kubernetes/pull/2825


dcbw:
@pperiyasamy want to LGTM?

PR #1352, opened 2022-11-02
OCPBUGS-3291: Disable packet MTU check when OVS HW Offload is enabled

wizhaoredhat:
This is a OVS dependent work around fix. If we detect that OVS HW Offload is enabled, we are aware that the offloading of packet MTU check support in OVS is not supported. This support will not be available until the offload support for sFlow is added.

In order to have the most optimal performance, the packet MTU check support should be disabled when OVS HW Offload is enabled on the node.

Backport of upstream https://github.com/ovn-org/ovn-kubernetes/pull/3227
(cherry picked from commit 70e0e13bb1bf15a6c5d2b2ca36988ddf74bd5d13)

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -239,10 +239,30 @@ func gatewayInitInternal(nodeName, gwIntf, egressGatewayIntf string, gwNextHops 
 			return nil, nil, err
 		}
 
+		ovsHardwareOffloadEnabled, err := util.IsOvsHwOffloadEnabled()
+		if err != nil {
+			return nil, nil, err
+		}
+
+		reason := ""
 		if !chkPktLengthSupported {
-			klog.Warningf("OVS on this node does not support check packet length action in kernel datapath. This "+
-				"will cause incoming packets destined to OVN and larger than pod MTU: %d to the node, being dropped "+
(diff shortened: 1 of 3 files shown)


zshi-redhat:
@wizhaoredhat Could you use `cherry-pick -x` to preserve the upstream commit info and create a OCPBUGS in jira for this change (it will be required when backporting to 4.12 branch).

wizhaoredhat:
> @wizhaoredhat Could you use `cherry-pick -x` to preserve the upstream commit info and create a OCPBUGS in jira for this change (it will be required when backporting to 4.12 branch).

@zshi-redhat Good point. I created OCPBUGS-3291 targeting 4.13.0. We will need another bug to target 4.12.z. Not sure if we need to backport all the way to 4.11.z. Also updated the commit with the commit hash of the cherry-pick.

wizhaoredhat:
@trozet Hi Tim, could this get an approval from you? Thanks

wizhaoredhat:
@dcbw Hi Dan could we get an approval for this?

wizhaoredhat:
Noticed that master synced and pulled in my changes.

wizhaoredhat:
Also this commit is part of master, release-4.13 and release-4.14

PR #1354, opened 2022-11-03
OCPBUGS-1486: Avoid re-metricing pods

bpickard22:
Issue: When a new ovnkube master starts, or gains leadership through leaderelection, addLogicalPort is called for all existing pods to ensure they are set up correctly, which means we recollect the podCreated metrics on pods that may not need it, and we will set an incorrect creation latency for pods.

Solution: Do not record pod creation latency metric for pods with annotations or pods with existing logical switch ports when addLogicalPort is ran

Signed-off-by: Ben Pickard <bpickard@redhat.com>

### DIFF
--- a/go-controller/pkg/ovn/pods.go
+++ b/go-controller/pkg/ovn/pods.go
@@ -727,7 +727,10 @@ func (oc *Controller) addLogicalPort(pod *kapi.Pod) (err error) {
 		}
 	}
 	// observe the pod creation latency metric.
-	metrics.RecordPodCreated(pod)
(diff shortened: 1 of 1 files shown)


dcbw:
@bpickard22 did this get cherry-picked from an upstream commit? If so, can you reference that commit SHA or PR? If not, can you resubmit upstream to ovn-org/ovn-kubernetes, and then we can do the downstream part when that merges?

bpickard22:
> 

@dcbw will resubmit upstream, thought this was a downstream issue given that it was an openshift bug

bpickard22:
@dcbw https://github.com/ovn-org/ovn-kubernetes/pull/3259

PR #1355, opened 2022-11-03
SDN-3589: downstream merge 11-02-2022

msherif1234:
Merge conflicts:
go-controller/pkg/node/gateway_shared_intf.go
go-controller/pkg/node/node.go 
go-controller/pkg/ovn/acl.go 
go-controller/pkg/ovn/egressgw.go 
go-controller/pkg/ovn/egressgw_test.go 
go-controller/pkg/ovn/egressip.go 
go-controller/pkg/ovn/egressip_test.go 
go-controller/pkg/ovn/hybrid_test.go 
go-controller/pkg/ovn/master.go 
go-controller/pkg/ovn/master_test.go 
go-controller/pkg/ovn/namespace.go 
go-controller/pkg/ovn/ovn.go 
go-controller/pkg/ovn/pods.go 
go-controller/pkg/ovn/pods_test.go 
go-controller/pkg/ovn/policy.go 
go-controller/pkg/ovn/policy_test.go 
go-controller/pkg/util/pod_annotation.go 

delete
go-controller/pkg/ovn/obj_retry.go

PR #1358, opened 2022-11-05
EIP: remove downstream's duplicate node delete test

msherif1234:
during review #1355 dcbw caught duplicate test downstream
removing this test to keep upstream and downsteam in sync
Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>


tssurya:
/lgtm 
but for every new commit we are adding here and in the other downstream merge - the last few commits that weren't squashed into the conflict commits.. we are actually deviating from upstream ovnk right? Usually when looking at previous merges the pattern has been to keep the same history and outline the conflicts in each commit...? Is this ok?

tssurya:
seems like 4.12 has this duplicate test issue, but 4.11 doesn't.. so this might need back port? not sure if that needs a bugzilla :D

msherif1234:
> /lgtm thanks Mohamed! not related to this PR: but for every new commit we are adding here and in the other downstream merge - the last few commits that weren't squashed into the conflict commits.. we are actually deviating from upstream ovnk right? Usually when looking at previous merges the pattern has been to keep the same history and outline the conflicts in each commit...? Is this ok? - on the other hand the bridge has already been crossed so I guess we need to do this one anyways since the problem is only in downstream? I wonder for how long this duplicate test has existed?

this is downstream and it was introduced by this commit 
https://github.com/openshift/ovn-kubernetes//commit/2100d49a169b0df503b4052da36b8aab72b53ec9

trozet (approved):

PR #1360, opened 2022-11-07
OCPBUGS-166: Handle k8s watcher restart scenario

pperiyasamy:
Sometimes k8s watcher getting restarted and for this condition update events are sent for all updated resources. If the object was recreated with same namespace and name while the watcher was off, ovnk gets update event where old and new objects have different UIDs.

Hence this commit handles such scenario by comparing object UIDs, deleting resources for old object followed by adding resources for new object.

Signed-off-by: Periyasamy Palanisamy <pepalani@redhat.com>
(cherry picked from commit bcd6124549e2e9ced9a397dc21c02f26f651bedd)


npinaeva:
/close
merged here https://github.com/openshift/ovn-kubernetes/pull/1364

PR #1366, opened 2022-11-09
[release-4.11] OCPBUGS-2859: node, route update: generic way to filter routes

kyrtapz:
This PR is required in 4.11 to allow for rollbacks from 4.12: https://issues.redhat.com/browse/OCPBUGS-2517?focusedCommentId=21154727&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21154727

Currently the rollback fails because 4.11 ovnkube-node is trying to add a route with a different destination which fails with the following error:
```
F1106 04:48:21.166135  193616 ovnkube.go:133] unable to add/update route for service via br-ex, error: file exists
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.12-upgrade-from-stable-4.11-e2e-aws-ovn-upgrade-rollback/1589046338517995520/artifacts/e2e-aws-ovn-upgrade-rollback/gather-extra/artifacts/pods/openshift-ovn-kubernetes_ovnkube-node-z8ljm_ovnkube-node.log

With 69d5d7d3cf434dfa7e864b7a9f19b1c81f4333d8 backported the route will get updated.

(cherry picked from commit 69d5d7d3cf434dfa7e864b7a9f19b1c81f4333d8)

### DIFF
--- a/go-controller/pkg/util/net_linux.go
+++ b/go-controller/pkg/util/net_linux.go
@@ -299,7 +299,7 @@ func LinkRoutesAdd(link netlink.Link, gwIP net.IP, subnets []*net.IPNet, mtu int
 
 func LinkRoutesAddOrUpdateMTU(link netlink.Link, gwIP net.IP, subnets []*net.IPNet, mtu int) error {
 	for _, subnet := range subnets {
-		route, err := LinkRouteGet(link, gwIP, subnet)
+		route, err := LinkRouteGetFilteredRoute(filterRouteByDst(link, subnet))
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


jcaamano:
@kyrtapz I am a bit confused on this one. Is it the gateway what is being updated, not the destination? Also, wouldn't it be updated only if the MTU is different? Is that actually happening?

kyrtapz:
> @kyrtapz I am a bit confused on this one. Is it the gateway what is being updated, not the destination? Also, wouldn't it be updated only if the MTU is different? Is that actually happening?

the changes introduced allow us to change the route gw from the newly introduced `169.254.169.4` to default gw:
`172.30.0.0/16 via 169.254.169.4 dev br-ex mtu 8901`
`172.30.0.0/16 via 10.0.192.1 dev br-ex mtu 8901`

So we actually update the gw, which means that `LinkRoutesAddOrUpdateMTU` name is no longer correct, I thought of updating it to its current form(`LinkRoutesAddOrUpdateSourceOrMTU`) but I don't think it is correct either(we can also update the gw with it), especially in the downgrade case.

@jcaamano do you have any naming preferences?

kyrtapz:
/hold
With https://github.com/openshift/ovn-kubernetes/commit/69d5d7d3cf434dfa7e864b7a9f19b1c81f4333d8 backported the 4.12 route will be [found](https://github.com/openshift/ovn-kubernetes/pull/1366/files#diff-462c1c75c92761bec5429ecd1a006c4228ecfb2f1a5694fd98a69ac9efc6dd27R325) but it won't be changed since the MTU hasn't changed. the 4.11 route will get added after the node reboots.

On downgrade/upgrade the ovnkube-node image is changed when a previous route still exists. In the current implementation(master and this PR) we find the route but don't update it.
This means that the route changes only after the node is rebooted.

4.11 -> 4.12 example:
1. Rollout new onvkube-node, node routes:
```
# The service route was not replaced
172.30.0.0/16 via 10.0.192.1 dev br-ex mtu 8901
```
2. Reboot the node, node routes:
```
# New service route was added
172.30.0.0/16 via 169.254.169.4 dev br-ex mtu 8901 
```

We need to decide whether this behavior is acceptable before back-porting it to 4.11.

kyrtapz:
/close
We will address https://github.com/openshift/ovn-kubernetes/pull/1366#issuecomment-1319887191 upstream and back-port it to 4.11 accordingly

PR #1368, opened 2022-11-09
OCPBUGS-3468: Disable packet MTU check when OVS HW Offload is enabled

wizhaoredhat:
This is a OVS dependent work around fix. If we detect that OVS HW Offload is enabled, we are aware that the offloading of packet MTU check support in OVS is not supported. This support will not be available until the offload support for sFlow is added.

In order to have the most optimal performance, the packet MTU check support should be disabled when OVS HW Offload is enabled on the node.

Backport of upstream https://github.com/ovn-org/ovn-kubernetes/pull/3227
(cherry picked from commit 70e0e13bb1bf15a6c5d2b2ca36988ddf74bd5d13)

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -239,10 +239,30 @@ func gatewayInitInternal(nodeName, gwIntf, egressGatewayIntf string, gwNextHops 
 			return nil, nil, err
 		}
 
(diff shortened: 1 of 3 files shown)


wizhaoredhat:
/cc @trozet
Hi Tim, could you please take a look at this?

PR #1369, opened 2022-11-10
OCPBUGS-3262: Don't warn on failure to create pod when it isn't scheduled

oribon:
If the pod isn't yet scheduled, we should just ignore the event and try again later when we get the update event as we do in 4.11+.

Signed-off-by: Ori Braunshtein <obraunsh@redhat.com>

This is similar to 28fc798a72f469c2fa4deec3236bfb41f56402c9 but could not be cherry-picked since obj_retry is since 4.11


oribon:
/hold cancel
checked on a cluster-bot, seems fine

oribon:
closing, fixed by https://github.com/openshift/ovn-kubernetes/pull/1434

PR #1348, opened 2022-11-02
[release-4.12] OCPBUGS-3346: vendor: bump libovsdb to 8f21d188c3a50d0ce378bd66ec68215967aaad77

dcbw:
Pulls in the following commits, for which we really just care about the client logging one for performance.

```
64a1543b16 client: don't construct transaction log string when it's not needed
14b9f5b67b Make table name a const above generated model
```

(cherry picked from commit 3b56eb09a6a0fca6ad7e9d2361adf77f2b468120)

PR #1374, opened 2022-11-14
OCPBUGS-3614: [4.11] CARRY: client: don't construct transaction log string when it's not needed

dcbw:
Backport of https://github.com/ovn-org/libovsdb/pull/342 (see also 4.12 version https://github.com/openshift/ovn-kubernetes/pull/1348)

fmt.Sprintf() will be evaluated before the logging function is called, which means that even if the logging level is less than V(4) we'll still construct the whole string for the operation, just to discard it. That can be pretty expensive.

### DIFF
--- a/go-controller/go.mod
+++ b/go-controller/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/onsi/gomega v1.14.0
 	github.com/openshift/api v0.0.0-20211201215911-5a82bae32e46
 	github.com/openshift/client-go v0.0.0-20211202194848-d3f186f2d366
-	github.com/ovn-org/libovsdb v0.6.1-0.20220826133223-e4d2d8648be7
+	github.com/ovn-org/libovsdb v0.6.1-0.20221101143603-8f21d188c3a5
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.12.1
 	github.com/prometheus/client_model v0.2.0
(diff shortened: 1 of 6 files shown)


jcaamano:
@dcbw why not bump it it like in 4.12?

dcbw:
> @dcbw why not bump it it like in 4.12?

@jcaamano sure, looking at the diff we could do that. Delta is pretty small and well-contained.

dcbw:
@jcaamano did the revendor, PTAL thanks!

dcbw:
/hold cancel
https://coreos.slack.com/archives/CFP6ST0A3/p1670860628776879

dcbw:
/override ci/prow/e2e-aws-ovn-upgrade

not a network thing
```
ns/openshift-etcd pod/etcd-ip-10-0-220-163.ec2.internal node/ip-10-0-220-163.ec2.internal - 3478.73 seconds after deletion - reason/FailedCreatePodSandBox Failed to create pod sandbox: rpc error: code = Unknown desc = create dropped infra 3740280ad77f34e75cd12c85cd12f00ebc76daaa83e2a1d92e297cb656ab30bc cgroup: error creating cgroup path for devices: mkdir /sys/fs/cgroup/devices/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod871de77c2dc84506b1b76d91beb065d1.slice: no such file or directory}
```

PR #1375, opened 2022-11-14
OCPBUGS-3615: [4.10] CARRY: client: don't construct transaction log string when it's not needed

dcbw:
https://github.com/ovn-org/libovsdb/pull/342 (see also 4.12 version https://github.com/openshift/ovn-kubernetes/pull/1348 and 4.11 version https://github.com/openshift/ovn-kubernetes/pull/1374)

fmt.Sprintf() will be evaluated before the logging function is called, which means that even if the logging level is less than V(4) we'll still construct the whole string for the operation, just to discard it. That can be pretty expensive.


maiqueb (approved):
Nice.

tssurya (approved):
/lgtm

PR #1380, opened 2022-11-15
[WIP] make sure fix doesn't break windows 

JacobTanenbaum:
before this fix the hybrid overlay tried to resuse the allocated hybrid overlay distributed router ip address between restarts. This caused a race condition when we waited for all pods to finish processing, a pod add event could allocate the address before we allocated it for the hybrid overlay distributed router ip address.

Change the allocation to allocate a new IP address every time, there is no specific benifit to using the same ip address between restarts, there is a negligable amount of node side action that needs to take place when the address changes

**- How to verify it**


JacobTanenbaum:
downstream version of: https://github.com/ovn-org/ovn-kubernetes/pull/3276

PR #1395, opened 2022-11-23
[WIP] Netpol bundle db transactions

npinaeva:
Apply all db changes caused by the same event within 1 transaction as a part of `(i *informer) forEachQueuedHandler`.
Error handling with delayed transaction is safe, since the same key for pod can't be handled at the same time by another thread.
unit tests don't work on 4.10, but I still added some changes to make if fail less.

That is a follow up from https://github.com/openshift/ovn-kubernetes/pull/1394


npinaeva:
Testing results show performance improvements, see linked bug for more details.

PR #1401, opened 2022-11-29
Dockerfiles: consolidate common bits into base image

dcbw:
Consolidate RPM installation and some other common bits into the base image, and use that image from the main image.


dcbw:
Depends on https://github.com/openshift/release/pull/33002

dcbw:
Dupe of https://github.com/openshift/ovn-kubernetes/pull/1401

PR #1412, opened 2022-12-01
[release-4.12] Bug OCPBUGS-4380: Hold lock when deleting completed pod during update event

ricky-rav:
Hold lock when deleting completed pod during update event

Original upstream PR: https://github.com/ovn-org/ovn-kubernetes/pull/3274

Closes #OCPBUGS-4380


trozet (approved):
/lgtm

PR #1414, opened 2022-12-01
egress ip: Skip mgmt ports that cannot have assignable IP addresses

wizhaoredhat:
There can be multiple management ports. Currently in the case of Full mode with hardware backed mgmt port, the virtual function representor and virtual function netdev exists separately. However the management port config belongs to the virtual function representor in this case. This caused the Egress IP health check to fail to initialize since there won't be an IP address on the virtual function representor interface. Instead the code now iterates through all management port interfaces, skips the virtual function representor and finds the virtual function netdev management port with a valid IP address configuration.

In the case of DPU, there will be a DPU-Host mode node and DPU mode node. Also it will always use hardware backed mgmt ports. The DPU mode node will only have a virtual function representor, thus the Egress IP Health Check won't run on this node. The DPU-Host mode node will only have a virtual function netdev; thus the Egress IP Health Check will run on this node.

Also see discussion in: https://github.com/ovn-org/ovn-kubernetes/pull/3251

Signed-off-by: William Zhao <wizhao@redhat.com>

### DIFF
--- a/go-controller/pkg/node/management-port-dpu.go
+++ b/go-controller/pkg/node/management-port-dpu.go
@@ -151,6 +151,11 @@ func (mp *managementPortRepresentor) CheckManagementPortHealth(cfg *managementPo
 		stopChan)
 }
 
+// Port representors should not have any IP address assignable to them, thus always return false.
+func (mp *managementPortRepresentor) HasIpAddr() bool {
+	return false
+}
+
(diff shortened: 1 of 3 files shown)


wizhaoredhat:
Mistakenly posted this patch downstream. Sorry about the spam.

PR #1416, opened 2022-12-02
[release-4.12] OCPBUGS-4453: Set NODAD flag on masquerade address

dougsland:
When there are multiple addresses on the primary interface for a node, conntrack uses the first one by default. During upgrade that ends up being the masquerade address because it was added last. This causes connection problems because the masquerade address ends up in a dadfailed state because we use the same address on all nodes. Since we don't care about DAD on that address, let's disable it so the address doesn't get into a bad state.

Conflicts:
go-controller/go.mod

Signed-off-by: Ben Nemec <bnemec@redhat.com>
Signed-off-by: Douglas Schilling Landgraf <dougsland@redhat.com>
(cherry picked from commit 2508881b1efce6c2f4ef9d977b4f5eeb8d99b024)

**- How to verify it**


dcbw:
IMage jobs were all failing due to https://coreos.slack.com/archives/CBN38N3MW/p1669999531511899

flavio-fernandes:
@dougsland let's add this line change to go.mod
```
golang.org/x/sys v0.2.0
```

PR #1422, opened 2022-12-06
[release-4.11] OCPBUGS-4564: pods: deleteLogicalPort should not fail when node is gone

flavio-fernandes:
deleteLogicalPort should not fail when node is already deleted from lsManager's cache.

Cherry pick from https://github.com/openshift/ovn-kubernetes/pull/1419

Also back porting fixes from the following jiras here, which are all related to similar issues:

[OCPBUGS-4617](https://issues.redhat.com/browse/OCPBUGS-4617)
[OCPBUGS-4618](https://issues.redhat.com/browse/OCPBUGS-4618)

PR #1423, opened 2022-12-06
[release-4.10] OCPBUGS-1568: pods: deleteLogicalPort should not fail when node is gone

flavio-fernandes:
deleteLogicalPort should not fail when node is already deleted from lsManager's cache.

Cherry pick from https://github.com/openshift/ovn-kubernetes/pull/1422

Also back porting fixes from the following jiras here, which are all related to similar issues:

[OCPBUGS-4622](https://issues.redhat.com/browse/OCPBUGS-4622)
[OCPBUGS-4623](https://issues.redhat.com/browse/OCPBUGS-4623)

### DIFF
--- a/go-controller/pkg/ovn/pods.go
+++ b/go-controller/pkg/ovn/pods.go
@@ -15,6 +15,7 @@ import (
 	util "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 	"github.com/pkg/errors"
 	kapi "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	ktypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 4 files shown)


go-controller/pkg/ovn/pods.go
```diff
@@ -114,10 +114,15 @@ func (oc *Controller) syncPodsRetriable(pods []interface{}) error {
 		// find the logical switch for the node
 		ls := &nbdb.LogicalSwitch{}
 		if lsUUID, ok := oc.lsManager.GetUUID(n.Name); !ok {
-			klog.Errorf("Error getting logical switch for node %s: %s", n.Name, "Switch not in logical switch cache")
+			klog.Warningf("Error getting logical switch for node %s: %s", n.Name, "Switch not in logical switch cache")
 
 			// Not in cache: Try getting the logical switch from ovn database (slower method)
+			// It is possible that logical switch is removed and we can safely skip it, since there
+			// are no stale ports to worry about in that case.
 			if ls, err = libovsdbops.FindSwitchByName(oc.nbClient, n.Name); err != nil {
```
martinkennelly, line 122: Does `FindSwitchByName` return the error wrapped so we can detect `ErrNotFound`?

flavio-fernandes: it should! I will dbl check

flavio-fernandes: phew... I think I'm not wrong this time @martinkennelly :
https://github.com/openshift/ovn-kubernetes/blob/201ea08fde1a2fbdf1951064829cba7c5d8580f7/go-controller/pkg/libovsdbops/switch.go#L84

trozet:
ironic is failing on dualstack jobs for the past week

PR #1433, opened 2022-12-09
DNM/TEST: Use ovsdb client to detect if schema needs changing

martinkennelly:
Convert the db schema when there is a difference not just when there is a difference in version.
This should support upgrade and downgrade and also if we consume a schema where the author doesn't
change the schema version when making changes to a schema.

Reported-at: https://issues.redhat.com/browse/OCPBUGS-4557

/cc


martinkennelly:
/close
test successful

PR #1434, opened 2022-12-12
OCPBUGS-3778: [release-4.10] Netpol races partial fixes

npinaeva:
The problems that are fixed, and perf improvement:
1. don't retry unscheduled pod, wait for update event instead. 
2.  Cleanup pod handler for namespaceAndPod handler on namespace delete
event.
3. Only update localPods after successful db transaction, 
return fast from localPod handlers based on `np.localPods`
4. Don't retry fetching lsp from the lspCache, that "stops" the handler for 1 second
5.  Use stored portUUID to delete local pods instead of getting that info
from lspCache

### DIFF
--- a/go-controller/pkg/ovn/pods_test.go
+++ b/go-controller/pkg/ovn/pods_test.go
@@ -1272,14 +1272,6 @@ var _ = ginkgo.Describe("OVN Pod Operations", func() {
 
 				initialDB = libovsdbtest.TestSetup{
 					NBData: []libovsdbtest.TestData{
-						&nbdb.LogicalSwitch{
-							Name:  "node1",
-							Ports: []string{t1.portUUID},
-						},
-						&nbdb.LogicalSwitch{
-							Name:  "node2",
-							Ports: []string{t2.portUUID},
-						},
 						&nbdb.LogicalSwitchPort{
 							UUID:      t1.portUUID,
 							Name:      util.GetLogicalPortName(t1.namespace, t1.podName),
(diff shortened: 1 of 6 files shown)


go-controller/pkg/ovn/policy.go
```diff
@@ -894,73 +895,26 @@ func (oc *Controller) processLocalPodSelectorSetPods(policy *knet.NetworkPolicy,
...
-			continue
+		// port was deleted and not re-added to the lspCache yet, will be retried on the next update event
+		if !portInfo.expires.IsZero() {
+			klog.Warningf("Stale LSP %s for network policy %s found in cache", portInfo.name, policy.Name)
+			return
 		}
 
-		getPolicyPortsWg.Add(1)
-		go getPortInfo(pod)
+		// LSP get succeeded and LSP is up to fresh, exit and continue
```
trozet, line 910: nit if you push again: remove "exit and continue"

go-controller/pkg/ovn/pods_test.go
```diff
@@ -1310,6 +1302,14 @@ var _ = ginkgo.Describe("OVN Pod Operations", func() {
 							},
 							PortSecurity: []string{fmt.Sprintf("%s %s", t2.podMAC, t2.podIP)},
 						},
+						&nbdb.LogicalSwitch{
```
trozet, line 1305: why did you need to move this?

npinaeva: it sets the referencing UUIDs correctly , and makes "reconciles an existing pod with an existing logical switch port" test pass

go-controller/pkg/ovn/policy.go
```diff
@@ -874,14 +874,15 @@ func (oc *Controller) processLocalPodSelectorSetPods(policy *knet.NetworkPolicy,
 	// paranoid.
 	policyPorts = make([]string, 0, len(objs))
 	policyPortsInfo := make([]*lpInfo, 0, len(objs))
+	localPodsInfo = map[string]*lpInfo{}
```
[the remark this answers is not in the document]
npinaeva: Checked, no regression

trozet (approved):
/lgtm

dcbw:
https://coreos.slack.com/archives/CFP6ST0A3/p1670860628776879

PR #1445, opened 2022-12-13
[release-4.11] OCPBUGS-4829: Set the node as reachable only when it is being added as an egressip node

flavio-fernandes:
egressip node update: set the node as reachable only when it is being added as an egressip node

there are three places where ovnkube marks an egress node as reachable:
- node add event handler
- node update even handler
- periodic connectivity checker

In a scenario where the egressip node becomes reachable at the same time as the node object is updated
the node is set as reachable but because there was no label/readiness change it is never added as an egressip node.
The periodic check will then ignore the node because it considers it as already reachable.

To address the issue only set the node as reachable if it is actually getting added as an egressip node.

Conflicts:
go-controller/pkg/ovn/egressip.go
go-controller/pkg/ovn/obj_retry_master.go --> go-controller/pkg/ovn/obj_retry.go
go-controller/pkg/ovn/ovn.go
go-controller/pkg/ovn/egressip_test.go --> depends on changes that do not exist in release-4.11

(cherry picked from commit 65bd3ae8c7d8ccc9a2814a221c989dd8ab8155d0)

### DIFF
--- a/go-controller/pkg/ovn/egressip_test.go
+++ b/go-controller/pkg/ovn/egressip_test.go
@@ -26,10 +26,12 @@ import (
 	utilpointer "k8s.io/utils/pointer"
 )
 
-type fakeEgressIPDialer struct{}
+type fakeEgressIPDialer struct {
+	dialResult bool
+}
 
-func (f fakeEgressIPDialer) dial(ip net.IP, timeout time.Duration) bool {
-	return true
+func (f *fakeEgressIPDialer) dial(ip net.IP, timeout time.Duration) bool {
+	return f.dialResult
 }
 
 var (
@@ -134,7 +136,7 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations", func() {
 		},
 	}
 
-	dialer = fakeEgressIPDialer{}
+	dialer = &fakeEgressIPDialer{true}
 
 	getEgressIPAllocatorSizeSafely := func() int {
 		fakeOvn.controller.eIPC.allocator.Lock()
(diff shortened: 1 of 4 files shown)


go-controller/pkg/ovn/egressip_test.go
```diff
@@ -7091,6 +7091,115 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations", func() {
...
+
+				err := fakeOvn.controller.WatchEgressNodes()
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+				gomega.Eventually(getEgressIPStatusLen(eIP1.Name)).Should(gomega.Equal(1))
+				egressIPs, _ := getEgressIPStatus(eIP1.Name)
+				gomega.Expect(egressIPs[0]).To(gomega.Equal(egressIP))
+
+				// TODO (FF): Backport healthClient and fakeEgressIPHealthClient to release-4.11 in order to use this test.
+				//            Maybe not add this test at all until then?!?
```
flavio-fernandes, line 7177: @kyrtapz Maybe not add this test at all until then?!?

kyrtapz: The test doesn't do much as it is right now(the second half commented out).
It would be great if we could adapt it to be compatible with what is currently in 4.11.

flavio-fernandes: > 

yeah, that is my dilemma. I feel it will be a substantial effort to do that w/out backporting the changes in the main code, just so we can run the test. @kyrtapz  May we simply remove it from this PR and deal with that -- or not -- separately?

kyrtapz: I don't think it is too difficult.
I tried this out here: https://github.com/openshift/ovn-kubernetes/pull/1456 had to import one more commit you created

flavio-fernandes: Ah, nice! I forgot I had https://github.com/openshift/ovn-kubernetes/pull/1456 in a separate commit. 😅
I updated the PR with the passing new test.
TY @kyrtapz !

flavio-fernandes:
@trozet @dcbw I think this PR will need to override e2e-metal-ipi-ovn-dualstack as well.
😣

trozet (approved):
/label backport-risk-assessed

PR #1464, opened 2023-01-03
[release-4.10] OCPBUGS-5299: Set the node as reachable only when it is being added as an egressip node

flavio-fernandes:
egressip node update: set the node as reachable only when it is being added as an egressip node

there are three places where ovnkube marks an egress node as reachable:

- node add event handler
- node update even handler
- periodic connectivity checker

In a scenario where the egressip node becomes reachable at the same time as the node object is updated
the node is set as reachable but because there was no label/readiness change it is never added as an egressip node.
The periodic check will then ignore the node because it considers it as already reachable.

To address the issue only set the node as reachable if it is actually getting added as an egressip node.

### DIFF
--- a/go-controller/pkg/ovn/egressip_test.go
+++ b/go-controller/pkg/ovn/egressip_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
@@ -24,10 +25,12 @@ import (
 	utilpointer "k8s.io/utils/pointer"
 )
 
-type fakeEgressIPDialer struct{}
+type fakeEgressIPDialer struct {
+	dialResult bool
+}
 
-func (f fakeEgressIPDialer) dial(ip net.IP) bool {
-	return true
+func (f *fakeEgressIPDialer) dial(ip net.IP) bool {
+	return f.dialResult
 }
 
 var (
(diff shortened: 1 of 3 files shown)


go-controller/pkg/ovn/egressip_test.go
```diff
@@ -6489,6 +6492,114 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations", func() {
...
+
+				node.Annotations["test"] = "dummy"
+				_, err := fakeOvn.fakeClient.KubeClient.CoreV1().Nodes().Update(context.TODO(), &node, metav1.UpdateOptions{})
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				// the node should not be marked as reachable in the update handler as it is not getting added
+				gomega.Consistently(func() bool { return fakeOvn.controller.eIPC.allocator.cache[node.Name].isReachable }).Should(gomega.Equal(false))
+
+				// egress IP should get assigned on the next checkEgressNodesReachabilityIterate call
+				// explicitly call check reachability, periodic checker is not active
+				fakeDialer.dialResult = true
```
flavio-fernandes, line 6594: @kyrtapz I needed to tweak the ordering in this test, because node would be reachable when setting the dialer to return `true` after a node update. Can you please help me understand how this is not an issue on 4.11 and newer releases?

kyrtapz: I think there is a mistake here: https://github.com/openshift/ovn-kubernetes/pull/1464/commits/1e4ca0af7d1bd645f06c3a3c1a6270910cf61952#diff-796adb9d8a9b5b5636dc81e7350a00ef143e6e5bcb52ec30dc350c866cd9208cL1011-R1032
We should move [oc.setNodeEgressReachable](https://github.com/openshift/ovn-kubernetes/pull/1464/commits/1e4ca0af7d1bd645f06c3a3c1a6270910cf61952#diff-796adb9d8a9b5b5636dc81e7350a00ef143e6e5bcb52ec30dc350c866cd9208cR1008) and not `oc.setNodeEgressAssignable`

flavio-fernandes: whoa!!! 🤦

PR #1474, opened 2023-01-13
Bug 2047299, OCPBUGS-2337: [DownstreamMerge] 13 Jan 2023

kyrtapz:
Revert the previous merge https://github.com/openshift/ovn-kubernetes/commit/c76929fe8d9f59aa4456a97703b802ef1a5c5bf3 since it was done wrong :disappointed: 

Merge current upstream:

Conflicts:
go-controller/pkg/ovn/base_network_controller_pods.go
go-controller/pkg/ovn/pods.go

Moved addLogicalPortToNetwork back to BaseNetworkController since BaseSecondaryNetworkController uses it.
HACK: Downstream addLogicalPortToNetwork needs to call addPodToNamespace(DefaultNetworkController), to fix that provide a DefaultNetworkController pointer as an argument.

PR #1476, opened 2023-01-13
OCPBUGS-5840: ovnkube-node: Existing management port check

wizhaoredhat:
When ovnkube-node configures management port interface it may fail if host already have it configured but it is different interface. To handle that do the following:
- verify that port exists on the integrational bridge and it represents provided management port netdevice or OVS internal port, if no netdevice were provided;
- verify that no interface exists with the name of the management port;

If any of the checks had failed then need to remove port from the bridge, unless it is OVS internal port and we expect it; flush IP addresses, routes, tables from management netdevice interface. Both representor and netdevice interfaces renamed to original names, if latter stored in OVS database, or to generic names: {net|rep} + ddmmyyHHMMSS

Because ovnkube doesn't have unload sequence, port check done during new management port creation.

Signed-off-by: Dmytro Linkin <dlinkin@nvidia.com>
(cherry picked from upstream commit 9da055f26906d2dcf40af2584df9196878116503)

### DIFF
--- a/go-controller/pkg/node/management-port-dpu.go
+++ b/go-controller/pkg/node/management-port-dpu.go
@@ -42,18 +42,27 @@ func (mp *managementPortRepresentor) Create(nodeAnnotator kube.Annotator, waiter
 	if config.OvnKubeNode.MgmtPortRepresentor != "" {
 		k8sMgmtIntfName += "_0"
 	}
+
+	klog.Infof("Lookup representor link and existing management port")
 	// Get management port representor netdevice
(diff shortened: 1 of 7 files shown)


wizhaoredhat:
Might be taken care of in this sync here: https://github.com/openshift/ovn-kubernetes/pull/1474

wizhaoredhat:
Already part of master:
https://github.com/openshift/ovn-kubernetes/commit/9da055f26906d2dcf40af2584df9196878116503
This PR is no longer needed.

PR #1481, opened 2023-01-18
iptables: use container iptables, not the host's

dcbw:
Using the host's iptables was necessary when the container image might have been running on RHEL7 BYOH, but given that's no longer possible, we can clean this up and just use the container's iptables (since both it and the host will be iptables-nft).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,6 @@ FROM registry.ci.openshift.org/ocp/4.13:cli AS cli
 # - creating directories required by ovn-kubernetes
 # - git commit number
 # - ovnkube.sh script
-# - iptables wrappers
 FROM registry.ci.openshift.org/ocp/4.13:ovn-kubernetes-base
 
 USER root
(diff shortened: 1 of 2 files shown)


msherif1234:
do u want to wipe `./dist/images/iptables-scripts/` since its no longer in use ?

dcbw:
> do u want to wipe `./dist/images/iptables-scripts/` since its no longer in use ?

No, the Dockerfile.fedora stuff still uses it; we can't control what mode Fedora users have their iptables/iptables-nft in so I think we still need it there.

dcbw:
/retest-required

vsphere RPM mirrors are working again

dcbw:
/test e2e-aws-ovn-local-to-shared-gateway-mode-migration

```
level=error msg=Error: creating EC2 Instance: InvalidParameterValue: Value (ci-op-wnjmh14v-9f5e1-m5w7x-bootstrap-profile) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name
level=error msg=	status code: 400, request id: c26ac0ce-1080-4280-98e3-d32edce05cd8
level=error
level=error msg=  with aws_instance.bootstrap, 
```

PR #1484, opened 2023-01-19
[release-4.12] OCPBUGS-6040: addMasqueradeRoute: fallback to gateway interface IPs

jcaamano:
There might be circumstances where the node status internal IPs are not current (in our case, a dual-stack conversion procedure in which kubelet is not configured with the newly acquired IPv6 address and restarted, which in itself is a problem that would need to be fixed).

While node status IPs are preferred to avoid issues with IPs that the user or other components might arbitrarily add to the gateway interface, fallback to an interface IP if we are missing it from the node status.

Also fix DPU. Masquerading should not be needed in the DPU (when newSharedGateway is called with mode NodeModeDPU), only in the host (where initGatewayDPUHost is use instead for NodeModeDPUHost). Ref: https://github.com/openshift/ovn-kubernetes/blob/master/docs/design/dpu_support.md

(cherry picked from commit f105108f2cbc6f026b214d99ba198eaf2c5d3b3f)

Conflicts:
* [go-controller/pkg/node/gateway_shared_intf.go](https://github.com/openshift/ovn-kubernetes/compare/release-4.12...jcaamano:ovn-kubernetes-downstream:release-4.12?expand=1#diff-d3aa58d9b58a0a09264f072df46ab01d0501eb508c4656411ae2dc1ac68fb3c4)
  Replace 4.13 LinkRoutesApply with 4.12 LinkRoutesAddOrUpdateSourceOrMTU
* [go-controller/pkg/node/gateway_init_linux_test.go](https://github.com/openshift/ovn-kubernetes/compare/release-4.12...jcaamano:ovn-kubernetes-downstream:release-4.12?expand=1#diff-baef649410d0c75a89ece69e5a452be4a68d9fb8f8febed16657ff4fc2e6b2a4)
  For DPU host test, sharedGw.Start() call was added in 4.13, but the version of that in 4.12 is sharedGw.Start(stop, wg)

### DIFF
--- a/go-controller/pkg/node/gateway_init.go
+++ b/go-controller/pkg/node/gateway_init.go
@@ -402,11 +402,16 @@ func (n *OvnNode) initGatewayDPUHost(kubeNodeIP net.IP) error {
 		return err
 	}
 
+	ifAddrs, err := getNetworkInterfaceIPAddresses(gatewayIntf)
+	if err != nil {
+		return err
+	}
+
 	if err := setNodeMasqueradeIPOnExtBridge(gwIntf); err != nil {
 		return fmt.Errorf("failed to set the node masquerade IP on the ext bridge %s: %v", gwIntf, err)
 	}
 
-	if err := addMasqueradeRoute(gwIntf, n.name, n.watchFactory); err != nil {
+	if err := addMasqueradeRoute(gwIntf, n.name, ifAddrs, n.watchFactory); err != nil {
 		return fmt.Errorf("failed to set the node masquerade route to OVN: %v", err)
 	}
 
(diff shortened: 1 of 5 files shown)


jcaamano:
/retitle [release-4.12] OCPBUGS-6040: addMasqueradeRoute: fallback to gateway interface IPs

jcaamano:
e2e-metal-ipi-ovn-dualstack failure is unrelated: https://issues.redhat.com/browse/OCPBUGS-4425
e2e-aws-ovn-shared-to-local-gateway-mode-migration is failing in different ways simarly to other PRs.

/retest-required

trozet (approved):
/lgtm

jcaamano:
Checking the logs of e2e-gcp-ovn this change is working as expected and the failure seems unrelated; but the job has already failed 5 times.

/retest-required

PR #1514, opened 2023-02-08
[DNM] [release-4.10] Always send packets from OVN to Host

tssurya:
In LGW packets coming from OVN are sent straight outside into the wire from br-ex if the destination is outside the node. However we want to steer such traffic into the node so that the custom routes users have in lgw mode can take effect.

(cherry picked from commit 02626d5b435bc4dd3d4c45edb9bd4a9c1b875374)

**- How to verify it**


dcbw:
@tssurya ./pkg/node/gateway_shared_intf.go needs a gofmt; might be that 4.10 uses an older golang that has different formatting rules?

tssurya:
sorry this is not up for review, its for a customer test. apologies. Added a DNM flag.

PR #1525, opened 2023-02-10
cleanup: drop redundant selinux-policy install in onvkube dockerfiles

dcbw:
They're already installed by the base Dockerfiles.


dcbw:
if the image built, the PR works...

PR #1527, opened 2023-02-13
[WIP] OCPBUGS-6947: Check the "serving" field for endpoints

ricky-rav:

**- How to verify it**

### DIFF
--- a/go-controller/pkg/util/kube.go
+++ b/go-controller/pkg/util/kube.go
@@ -351,13 +351,13 @@ type LbEndpoints struct {
 	Port  int32
 }
 
-// GetLbEndpoints return the endpoints that belong to the IPFamily as a slice of IPs
-func GetLbEndpoints(slices []*discovery.EndpointSlice, svcPort kapi.ServicePort) LbEndpoints {
+// GetLbEndpoints returns the IPv4 and IPv6 addresses of valid endpoints as slices inside a struct
+func GetLbEndpoints(slices []*discovery.EndpointSlice, svcPort kapi.ServicePort, includeTerminating bool) LbEndpoints {
 	v4ips := sets.NewString()
 	v6ips := sets.NewString()
 
 	out := LbEndpoints{}
-	// return an empty object so the caller don't have to check for nil and can use it as an iterator
+	// return an empty object so the caller doesn't have to check for nil and can use it as an iterator
 	if len(slices) == 0 {
 		return out
 	}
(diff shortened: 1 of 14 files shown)


go-controller/pkg/util/kube.go
```diff
@@ -352,7 +352,7 @@ func GetLbEndpoints(slices []*discovery.EndpointSlice, svcPort kapi.ServicePort)
 			out.Port = *port.Port
 			for _, endpoint := range slice.Endpoints {
 				// Skip endpoints that are not ready
-				if endpoint.Conditions.Ready != nil && !*endpoint.Conditions.Ready {
+				if IsEndpointServing(endpoint) {
```
msherif1234, line 355: do u mean `!IsEndpointServing()` ?

go-controller/pkg/util/util.go
```diff
@@ -349,3 +350,18 @@ func UpdateNodeSwitchExcludeIPs(nbClient libovsdbclient.Client, nodeName string,
...
+
+// discovery.EndpointSlice reports in the same slice all endpoints along with their status.
+// isEndpointReady takes an endpoint from an endpoint slice and returns true if the endpoint is
+// to be considered ready.
+func IsEndpointReady(endpoint discovery.Endpoint) bool {
+	return endpoint.Conditions.Ready == nil || *endpoint.Conditions.Ready
+}
+
+func IsEndpointServing(endpoint discovery.Endpoint) bool {
+	if endpoint.Conditions.Serving != nil {
```
msherif1234, line 362: do u need to check once its ~= nil to make sure its also `true` ?
so we are aligned to its definition here https://github.com/kubernetes/api/blob/0478a3e95231398d8b380dc2a1905972be8ae1d5/discovery/v1/types.go#L137 ?

go-controller/pkg/util/kube.go
ricky-rav: ops, yes.  I shouldn't code during meetings!

go-controller/pkg/util/util.go
ricky-rav: I discussed this with Mohamed on slack and we agreed that the documentation for this field might be a bit misleading. It says:

>  serving is identical to ready except that it is set regardless of the
>  terminating state of endpoints. This condition should be set to true for
>  a ready endpoint that is terminating. If nil, consumers should defer to
>  the ready condition.

The second sentence doesn't mean that `serving` is true only when the endpoint is terminating. It means that `serving` is always equal to ready, except for when the endpoint is terminating, in which case ready=false and serving=true. So we're good here,

go-controller/pkg/util/util.go
```diff
@@ -349,3 +350,18 @@ func UpdateNodeSwitchExcludeIPs(nbClient libovsdbclient.Client, nodeName string,
 
 	return nil
 }
+
+// discovery.EndpointSlice reports in the same slice all endpoints along with their status.
+// isEndpointReady takes an endpoint from an endpoint slice and returns true if the endpoint is
+// to be considered ready.
+func IsEndpointReady(endpoint discovery.Endpoint) bool {
+	return endpoint.Conditions.Ready == nil || *endpoint.Conditions.Ready
```
abhat, line 358: just curious why do we consider endpoint.Conditions.Ready==nil to imply endpoint is Ready instead of saying 
`return endpoint.Conditions.Ready != nil && *endpoint.Conditions.Ready`

ricky-rav: Only because of this comment in the documentation:
https://github.com/kubernetes/api/blob/0478a3e95231398d8b380dc2a1905972be8ae1d5/discovery/v1/types.go#L129-L131
`A nil value indicates an unknown state. In most cases consumers should interpret this unknown state as ready.
`
We can be stricter of course and only consider the field as true if it is set as such.

deads2k:
first attempt had a bug.  trying again

/payload 4.13 nightly blocking

go-controller/pkg/util/util.go
abhat: That makes sense. Thanks.

PR #1530, opened 2023-02-14
iptables package is missing in microshift image

zshi-redhat:


dcbw:
@zshi-redhat do we need the same change in the rhel8 Dockerfile.base ?

zshi-redhat:
> @zshi-redhat do we need the same change in the rhel8 Dockerfile.base ?

Yes, let me update the PR.

zshi-redhat:
> > @zshi-redhat do we need the same change in the rhel8 Dockerfile.base ?
> 
> Yes, let me update the PR.

Done.

PR #1531, opened 2023-02-14
OCPBUGS-7296: Remove ICNIv1 from ovn-kubernetes

JacobTanenbaum:
we no longer need to support ICNIv1. that allows us to remove the referances to the hybridOverlayVTEP and hybridOvertlayExternalGW and associated annotations and the relevant OCP_HACKS

**- How to verify it**


JacobTanenbaum:
/retest-required 

the job 4.13-upgrade-from-stable-4.12-e2e-aws-ovn-upgrade seems to fail often and was an issue with a sig-instrumentation test

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -640,7 +640,7 @@ func (bnc *BaseNetworkController) addLogicalPortToNetwork(oc *DefaultNetworkCont
 	var hybridOverlayExternalGW net.IP
 	if oc != nil {
 		// Ensure the namespace/nsInfo exists
-		routingExternalGWs, routingPodGWs, hybridOverlayExternalGW, ops, err = oc.addPodToNamespace(pod.Namespace, podIfAddrs)
```
dcbw, line 643: @JacobTanenbaum can't we now remove `hybridOverlayExternalGW` entirely from this function, and remove the argument from `addRoutesGatewayIP()` too?

dcbw:
@JacobTanenbaum @trozet do we need to clean up any left-over ICNIv1 OVN database objects on upgrade?

go-controller/pkg/ovn/base_network_controller_pods.go
trozet: we can defer this to another PR because we are going to have to change a bunch of files around to match upstream after this. @zeeke FYI

trozet:
> @JacobTanenbaum @trozet do we need to clean up any left-over ICNIv1 OVN database objects on upgrade?

Looks like the only thing that gets created in OVN (besides the hybrid port) are logical routes:
https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/ovn/hybrid.go#L129

These are not ICNI specific so I think we are good.

/lgtm

PR #1534, opened 2023-02-15
[WIP][DNM] [release-4.13] testing health check changes for cloud load balancers

ricky-rav:


go-controller/pkg/node/healthcheck_node.go
```diff
@@ -39,16 +43,62 @@ func newNodeProxyHealthzServer(nodeName, address string, eventRecorder record.Ev
...
+		klog.Errorf("[rira] couldn't read namespace file %s", namespace_filepath)
+		return false
+	}
+	namespace := string(namespaceBytes)
+	if len(namespace) == 0 {
+		klog.Errorf("[rira] found empty namespace")
+		return false
+	}
+	// spec.nodeName==$HOSTNAME
+	nodeName := os.Getenv("K8S_NODE")
```
dcbw, line 64: @ricky-rav we already pass the node name into `newNodeProxyHealthzServer()` so can we just use that?

go-controller/pkg/node/healthcheck_node.go
```diff
@@ -39,16 +43,62 @@ func newNodeProxyHealthzServer(nodeName, address string, eventRecorder record.Ev
...
+	podList, err := phu.kube.GetPods(
+		namespace,
+		metav1.LabelSelector{MatchLabels: ovnkubeNodeLabel})
+	if err != nil {
+		klog.Errorf("[rira] could not retrieve matching pods: %v", err)
+		return false
+	}
+	klog.Infof("[rira] we got %d matching pods", len(podList.Items))
+	var isTerminating bool
+	for _, pod := range podList.Items {
```
dcbw, line 81: Instead of getting all pods, can we just use the Kube Downward API (see https://kubernetes.io/docs/concepts/workloads/pods/downward-api/ ) and either set an environment variable for pod name and populate that with `metadata.name` rather than listing all pods? Then we can just get our one pod directly.

Kinda like bindata/network-diagnostics/network-check-source.yaml does with POD_NAME

ricky-rav: absolutely! I have local version of these changes that uses POD_NAME and rate-limits the requests to the API server to 1 every half a second. I'll post that in a couple of hours at most.

go-controller/pkg/node/healthcheck_node.go
ricky-rav: Ah true. I am not using the node name in the latest changes I haven't posted yet.

go-controller/pkg/node/healthcheck_node.go
```diff
@@ -39,16 +43,62 @@ func newNodeProxyHealthzServer(nodeName, address string, eventRecorder record.Ev
 			UID:       ktypes.UID(nodeName),
 			Namespace: "",
 		},
+		kube: kube,
 	}
 }
 
+func (phu *proxierHealthUpdater) isOvnkNodePodTerminating() bool {
+	// retrieve namespace name
+	namespace_filepath := "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
```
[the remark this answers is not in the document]
ricky-rav: Awesome! I wasn't aware :) Thanks!

PR #1546, opened 2023-02-28
[DownstreamMerge] 2-28-23 

trozet:
Includes carry commit to fully remove ICNIv1 and remove downstream only hacks around it.


trozet:
After CARRY applied, there were 0 merge conflicts.

/assign @ricky-rav

go-controller/hybrid-overlay/pkg/controller/node_linux.go
```diff
@@ -765,13 +733,6 @@ func (n *NodeController) syncFlows() {
 	_, stderr, err = util.ReplaceOFFlows(extBridgeName, flows)
 	if err != nil {
 		klog.Errorf("Failed to add flows, error: %v, stderr: %s, flows: %s", err, stderr, flows)
-		return
-	}
-
-	// handle resetting ignore learn as we may not have encountered some pods while
```
tssurya, line 771: this logic was only for ICNIV1? I don't know what `ignoreLearn` does but I hope removing it from this place was indeed intended.

tssurya (approved):
LGTM. waiting for the upgrade jobs to pass.

go-controller/hybrid-overlay/pkg/controller/node_linux.go
trozet: yeah this was for whether or not to store flows in our ovnk cache that we found in br-ex. For ICNIv1 it sometimes programmed dynamic learned flows (flows which are created by packets hitting another flows). Those were meant to be transient flows, so this ignoreLearn flag stopped us from accidentally including those in our ovnk side cache. This flag is not used outside of ICNIv1. The code now matches upstream.

PR #1549, opened 2023-03-01
node: small downstream CARRY patch cleanup

dcbw:
Consolidate some MCS blocking code that's the same between local and shared gateway modes.


dcbw:
Tide be super slow

PR #1560, opened 2023-03-07
Don't recreate clusterPGs and clusterRtrPGs unless needed

tssurya:
In SetupMaster, we always call CreateOrUpdatePortGroupsOps with empty ACLs and PGs for the cluster-wide port group and cluster-wide-router-PG. This is disruptive during upgrades since momentarily all efw ACLs and multicast ACLs will be wiped out.

This commit changes this to first check if the PG already exists, if then no need to do anything.
Each of those features are responsible for ensuring ACLs, Ports are good on those PGs they own.

NOTE: This bug was an issue for multicast and started being an issue for egf from https://github.com/ovn-org/ovn-kubernetes/pull/3338/commits/bd29f417b43c257897f27badc64af38df78b99bd Before that we didn't have ACLs on cluster wide PG.

(cherry picked from commit 935bc5570cb6661beb58e98288cae33c13bd6a8e) (cherry picked from commit 7c153b3071bf674fce53de429531d0ffd5c40dd4) (cherry picked from commit eb312deb02407671fff91ab1e9192a6b004a6997)

Conflicts:
	go-controller/pkg/ovn/master.go
tiny conflict in the imports since we have hocontroller import in 4.11 only

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 
+	libovsdbclient "github.com/ovn-org/libovsdb/client"
 	hocontroller "github.com/ovn-org/ovn-kubernetes/go-controller/hybrid-overlay/pkg/controller"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/informer"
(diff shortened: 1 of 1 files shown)


tssurya:
> @tssurya I took a look at this by chance when troubleshooting a different issue.
> 
> I do have a comment which we will need to consider going forward not meant for this PR.
> 
> The problem here is that we are always specifying to update the attributes:
> 
> ```
> // getAllUpdatableFields returns a list of all of the columns/fields that can be updated for a model
> func getAllUpdatableFields(model model.Model) []interface{} {
> 	switch t := model.(type) {
> 	case *nbdb.LogicalSwitchPort:
> 		return []interface{}{&t.Addresses, &t.Type, &t.TagRequest, &t.Options, &t.PortSecurity}
> 	case *nbdb.PortGroup:
> 		return []interface{}{&t.ACLs, &t.Ports, &t.ExternalIDs}
> 	default:
> 		panic(fmt.Sprintf("getAllUpdatableFields: unknown model %T", t))
> 	}
> }
> ```
> 
> That departs from the otherwise default behavior. The default behavior is only to update attributes that have a non-zero value and is 99% of the time the correct behavior.

hmm true, I forgot we did that behavior change a while ago!

> 
> The behavior that should be treated as a corner case is the one that we are trying handle with that piece of code above: resetting attributes to their zero value. We only want to do that in very specific cases.

true, I wanna admit that I found that piece of logic a bit weird that we always wanted to update those fields on startup but didn't know if that was intentional or not.

> 
> The approach in this PR is to have specific code in place outside of libovsdbops to handle a situation that is 99% of the time the correct behavior. And within libovsdbops we have generic code in case to handle a situation that is exceptional.
> 
> With time, similar specific handling will proliferate outside of libovsdbops for other similar situations while keeping within libovsdbops a generic behavior for corner cases. IMHO it should be the other way around: keep within libovsdbops a generic behavior that applies for most cases, and handle the corner case specifically where needed.
> 
> cc @trozet

true.

yeah thanks for bringing this up @jcaamano ! We should talk more about this in next week's team meeting when folks are back from PTO and try to educate (myself included) the team on an agreed approach for such situations.

PR #1569, opened 2023-03-09
[release-4.12] Check the "Serving" field for endpoints

ricky-rav:
4.12 backport of https://github.com/openshift/ovn-kubernetes/commit/f6ef43368cc79b85bf0de535ff3b79a5856ab481 
Not a clean backport, because I had to account for the changes from https://github.com/openshift/ovn-kubernetes/commit/0111e1faec20d16505a110449966273b430b7ad1#diff-d3aa58d9b58a0a09264f072df46ab01d0501eb508c4656411ae2dc1ac68fb3c4, which are in 4.13 but not in 4.12.

Master, node and healthcheck code looked up the Ready field in endpoints inside endpointslices, in line with the previous implementation that dealt with the older and less fine-grained Endpoints resource, that only distinguished between ready and not ready. Healthchecks should indeed rely on the Ready field, while node and master could look up the "Serving" field, so that we won't prematurely refuse incoming connections to an endpoint that is scheduled to be terminating but is still up and running.

Making a special case for services with PublishNotReadyAddresses set, as their endpoints should always be considered ready, even if terminating. A couple of upstream kubernetes tests use this feature in 1.26.


trozet (approved):
/lgtm

PR #1577, opened 2023-03-13
OCPBUGS-7346: [release-4.12] Fully remove dependency on default gateway for services

trozet:
There remained a dependency for return service traffic from OVN->Host where OVN would need to think it was routing towards the default gw. Once this traffic made it into the shared gw bridge, it would be hijacked and redirected back to the Host. This only applied to clusters where a default gateway exists. In a cluster with no default gateway, the default route for OVN pointed to a fake gw (169.254.169.4) with a static MAC binding for this address.

However, in a cluster where there is a default gateway, but we are unable to ARP and create a MAC binding, return service traffic will fail to be routed. This commit changes the behavior so that:

1. Dummy GW MAC bindings are always created
2. Always add a route destined to the dummy masquerade subnet via dummy GW. Like "169.254.169.0/29 via 169.254.169.4"

This removes depending on an external default gateway for routing return service traffic from OVN.

(cherry picked from commit 32490da97a6118eefc88cea52b5db25c7e59ad12)

Backport was clean other than conflicts in master_test.go and namespace_test.go.

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/master_test.go
+++ b/go-controller/pkg/ovn/master_test.go
@@ -877,8 +877,9 @@ var _ = ginkgo.Describe("Gateway Init Operations", func() {
 		eventsLock        sync.Mutex
 		recorder          *record.FakeRecorder
 
-		l3GatewayConfig *util.L3GatewayConfig
-		nodeHostAddrs   sets.String
+		l3GatewayConfig         *util.L3GatewayConfig
+		nodeHostAddrs           sets.String
+		expectedSBDatabaseState []libovsdbtest.TestData
 	)
 
 	const (
@@ -926,6 +927,12 @@ var _ = ginkgo.Describe("Gateway Init Operations", func() {
 		expectedClusterLBGroup := newLoadBalancerGroup()
 		expectedNodeSwitch := node1.logicalSwitch(expectedClusterLBGroup.UUID)
 
+		gr := types.GWRouterPrefix + node1.Name
+		datapath := &sbdb.DatapathBinding{
+			UUID:        gr + "-UUID",
+			ExternalIDs: map[string]string{"logical-router": gr + "-UUID", "name": gr},
+		}
+		expectedSBDatabaseState = []libovsdbtest.TestData{datapath}
 		dbSetup = libovsdbtest.TestSetup{
 			NBData: []libovsdbtest.TestData{
 				newClusterJoinSwitch(),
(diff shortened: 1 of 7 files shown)


go-controller/pkg/ovn/master_test.go
```diff
@@ -877,8 +877,9 @@ var _ = ginkgo.Describe("Gateway Init Operations", func() {
 		eventsLock        sync.Mutex
 		recorder          *record.FakeRecorder
 
-		l3GatewayConfig *util.L3GatewayConfig
-		nodeHostAddrs   sets.String
+		l3GatewayConfig         *util.L3GatewayConfig
+		nodeHostAddrs           sets.String
+		expectedSBDatabaseState []libovsdbtest.TestData
```
tssurya, line 882: this is the slight conflict part I guess, wondering why you didn't add the `expectedNBDatabaseState` as well if you added the SB variable anyways.. (https://github.com/openshift/ovn-kubernetes/commit/32490da97a6118eefc88cea52b5db25c7e59ad12#diff-da0dab63258c018f29107af5cc19fdfce50a8b86d24d9a2e285fff4334187845R894)

go-controller/pkg/ovn/namespace_test.go
```diff
@@ -254,7 +255,11 @@ var _ = ginkgo.Describe("OVN Namespace Operations", func() {
 				Name: ovntypes.ClusterLBGroupName,
 				UUID: ovntypes.ClusterLBGroupName + "-UUID",
 			}
-
+			gr := ovntypes.GWRouterPrefix + node1.Name
```
tssurya, line 258: second file with conflict also lgtm

tssurya (approved):
/lgtm
LGTM compared line-by-line to its predecessor: https://github.com/openshift/ovn-kubernetes/commit/32490da97a6118eefc88cea52b5db25c7e59ad12

self-note: had a tiny comment which is just a nit since it has no functional changes to the PR so I won't hold for that specially since its tests.
unit tests are passing, we are good here.

go-controller/pkg/ovn/master_test.go
trozet: oh cause dcbw had added a commit which cleaned up duplicate per test expectedNBDatabaseState and moved it to the top. That commit is not in 4.12, so thats why it is defined in each test case and not at the top.

PR #1567, opened 2023-03-08
OCPBUGS-8741: [release-4.13] Handle Completed pods deletion

zeeke:
**- What this PR does and why is it needed**

**- Special notes for reviewers**

**- How to verify it**

**- Description for the changelog**

Backport of:
- https://github.com/ovn-org/ovn-kubernetes/pull/3433

PR #1581, opened 2023-03-15
OCPBUGS-10314: [release-4.12] Handle Completed pods deletion

zeeke:

Backport of:
- https://github.com/openshift/ovn-kubernetes/pull/1567

Optimization about skipping pods on different nodes while search colliding IP has been removed, as nodeName is not present in `handlePeerPodSelectorDelete`

See [pkg/ovn/pods.go#L256](https://github.com/openshift/ovn-kubernetes/blob/f65d02b84f7acf787f53f11710916d871e8f2d5b/go-controller/pkg/ovn/pods.go#L256)

```
if util.PodCompleted(p) || !util.PodWantsNetwork(p) || !util.PodScheduled(p) || p.Spec.NodeName != nodeName {
	continue
}
```

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/pods.go
+++ b/go-controller/pkg/ovn/pods.go
@@ -246,32 +246,23 @@ func (oc *Controller) deleteLogicalPort(pod *kapi.Pod, portInfo *lpInfo) (err er
 	// check to make sure no other pods are using this IP before we try to release it if this is a completed pod.
 	if util.PodCompleted(pod) {
 		if shouldRelease, err = oc.lsManager.ConditionalIPRelease(nodeName, podIfAddrs, func() (bool, error) {
-			pods, err := oc.watchFactory.GetAllPods()
-			if err != nil {
-				return false, fmt.Errorf("unable to get pods to determine if completed pod IP is in use by another pod. "+
-					"Will not release pod %s/%s IP: %#v from allocator", pod.Namespace, pod.Name, podIfAddrs)
+			var needleIPs []net.IP
+			for _, podIPNet := range podIfAddrs {
+				needleIPs = append(needleIPs, podIPNet.IP)
 			}
-			// iterate through all pods, ignore pods on other nodes
-			for _, p := range pods {
-				if util.PodCompleted(p) || !util.PodWantsNetwork(p) || !util.PodScheduled(p) || p.Spec.NodeName != nodeName {
-					continue
-				}
-				// check if the pod addresses match in the OVN annotation
-				pAddrs, err := util.GetAllPodIPs(p)
-				if err != nil {
-					continue
-				}
 
-				for _, pAddr := range pAddrs {
-					for _, podAddr := range podIfAddrs {
-						if pAddr.Equal(podAddr.IP) {
-							klog.Infof("Will not release IP address: %s for pod %s/%s. Detected another pod"+
-								" using this IP: %s/%s", pAddr.String(), pod.Namespace, pod.Name, p.Namespace, p.Name)
-							return false, nil
-						}
-					}
-				}
+			collidingPod, err := oc.findPodWithIPAddresses(needleIPs)
+			if err != nil {
+				return false, fmt.Errorf("unable to determine if completed pod IP is in use by another pod. "+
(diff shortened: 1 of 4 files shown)


zeeke:
/retest

failing test:
```
 [sig-arch][Late] operators should not create watch channels very often [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel] 
```

go-controller/pkg/ovn/pods.go
```diff
@@ -340,6 +331,35 @@ func (oc *Controller) deleteLogicalPort(pod *kapi.Pod, portInfo *lpInfo) (err er
...
 
+func (bnc *Controller) findPodWithIPAddresses(needleIPs []net.IP) (*kapi.Pod, error) {
+	allPods, err := bnc.watchFactory.GetAllPods()
+	if err != nil {
+		return nil, fmt.Errorf("unable to get pods: %w", err)
+	}
+
+	// iterate through all pods
+	for _, p := range allPods {
+		if util.PodCompleted(p) || !util.PodWantsNetwork(p) || !util.PodScheduled(p) {
```
tssurya, line 342: Should we be adding back the `|| p.Spec.NodeName != nodeName` here? 
Example in https://github.com/openshift/ovn-kubernetes/pull/1567/files?diff=unified&w=0#diff-941b484ab7f41b84a6ba9c2dee91f3454401447db0ca70cb401a8cc582cc072eR299 we moved `|| expectedSwitchName != switchName` check to before`findPodWithIPAddresses`. Wondering why we are dropping this check here.

tssurya (changes requested):
I am almost lgtm, just one question: please see my comment on adding back the nodeName match?

go-controller/pkg/ovn/pods.go
tssurya: note that I have read your PR description where you say you didn't do this because `handlePeerPodSelectorDelete` doesn't have nodeName... but we could just pass empty string from there but do it at least in the main logic? It just seems like we are kinda dropping that thing out of the blue compared to what we used to do.

go-controller/pkg/ovn/pods.go
```diff
@@ -246,32 +246,23 @@ func (oc *Controller) deleteLogicalPort(pod *kapi.Pod, portInfo *lpInfo) (err er
 	// check to make sure no other pods are using this IP before we try to release it if this is a completed pod.
 	if util.PodCompleted(pod) {
 		if shouldRelease, err = oc.lsManager.ConditionalIPRelease(nodeName, podIfAddrs, func() (bool, error) {
-			pods, err := oc.watchFactory.GetAllPods()
+			var needleIPs []net.IP
```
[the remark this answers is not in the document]
zeeke: Yes: "haystack and needle" are sometimes used as [search function parameters](https://www.php.net/manual/en/function.array-search.php). I didn't find any better names :smiley:

go-controller/pkg/ovn/pods.go
zeeke: Good point! The fact is that that check has already been dropped in 4.13+ by [1], as neither `expectedSwitchName` 
 nor `switchName` are calculated from `p.Spec.NodeName`.

The changes you are proposing make release-4.12 branch diverge a little from master.

Saying that, if you are still in favor of restoring the check, I can go for it.

[1] https://github.com/ovn-org/ovn-kubernetes/commit/f620c6c6adbb2a2367eabb5c86b6fd2b0e831272#diff-941b484ab7f41b84a6ba9c2dee91f3454401447db0ca70cb401a8cc582cc072eR195

tssurya:
@trozet for backport-risk-assess

go-controller/pkg/ovn/pods.go
zeeke: Thanks @tssurya!

zeeke:
@zhaozhanqi @anuragthehatter
can you have a look at this and add cherry-pick-approved ?

dcbw:

Known flake https://redhat-internal.slack.com/archives/CM4ERHBJS/p1680032625731999

PR #1589, opened 2023-03-16
[release-4.13] rhel9: rollback to ovn22.12-22.12.0-25.el9fdp

jcaamano:
CT flush behavior introduced with ovn23.03 bump is not working well for us.

https://bugzilla.redhat.com/show_bug.cgi?id=2178962


dcbw:
We bumped OVN to disable the ct_flush stuff instead.

PR #1553, opened 2023-03-03
Dockerfiles: copy RHEL-9 bits over top of unused RHEL8 bits

dcbw:
Step 2 to get back to using regular names (eg without "-rhel-9" suffix). This is posssible without breaking CI because
https://github.com/openshift/release/pull/36772 merged which stops using these Dockerfiles.

Next steps:
- flip the dockerfile_paths in openshift/release back to the non-.rhel9 names since this PR makes the normal names RHEL9 now.
- finally 'git rm' the old .rhel9 Dockefiles in ovnkube.

PR #1605, opened 2023-03-24
OCPBUGS-10592: Remove pyOpenSSL dependency for ovn-detrace command

pperiyasamy:
The pyOpenSSL package is no longer needed for ovn-detrace with RHEL9 image and removing it from dependency check. There is a possiblity to use pip3 to install ovs python package on a running ovnk pod to make ovn-detrace command to work. hence adding python3-pip into install pkgs.

### DIFF
--- a/go-controller/cmd/ovnkube-trace/ovnkube-trace.go
+++ b/go-controller/cmd/ovnkube-trace/ovnkube-trace.go
@@ -887,7 +887,7 @@ func getOfprotoIPFamilyArgs(protocol string, ip net.IP) (string, string, string)
 func installOvnDetraceDependencies(coreclient *corev1client.CoreV1Client, restconfig *rest.Config, podName, ovnNamespace string) error {
 	dependencies := map[string]string{
 		"ovs":       "if type -p ovn-detrace >/dev/null 2>&1; then echo 'true' ; fi",
-		"pyOpenSSL": "if rpm -qa | egrep -q python3-pyOpenSSL; then echo 'true'; fi",
+		"pyOpenSSL": "if python -c 'import ssl; print(ssl.OPENSSL_VERSION)'; then echo 'true'; fi",
 	}
 	for dependency, dependencyCmd := range dependencies {
 		depVerifyOut, depVerifyErr, err := execInPod(coreclient, restconfig, ovnNamespace, podName, "ovnkube-node", dependencyCmd, "")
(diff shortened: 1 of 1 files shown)


pperiyasamy:
> `Error: Unable to find a match: python3-pyOpenSSL`
> 
> This package doesn't exist in RHEL9... ovn-detrace doesn't actually use any SSL directly, it uses the OVS python bits that got ported to the Python internal SSL last year.

Yes @dcbw , ovn-detrace is working fine now without a need for dependency on `python3-pyOpenSSL`.
```
# ./ovnkube-trace --src-namespace openshift-dns -src dns-default-fvftq -dst-namespace openshift-dns -dst dns-default-p2msg -udp -dst-port 53 -loglevel 0
I0331 13:23:12.238444  566820 ovs.go:90] Maximum command line arguments set to: 191102
ovn-trace source pod to destination pod indicates success from dns-default-fvftq to dns-default-p2msg
ovn-trace destination pod to source pod indicates success from dns-default-p2msg to dns-default-fvftq
ovs-appctl ofproto/trace source pod to destination pod indicates success from dns-default-fvftq to dns-default-p2msg
ovs-appctl ofproto/trace destination pod to source pod indicates success from dns-default-p2msg to dns-default-fvftq
ovn-detrace source pod to destination pod indicates success from dns-default-fvftq to dns-default-p2msg
ovn-detrace destination pod to source pod indicates success from dns-default-p2msg to dns-default-fvftq
```
Updated the PR with changing ovn-detrace dependency check accordingly.

Dockerfile
```diff
@@ -33,7 +33,7 @@ ENV PYTHONDONTWRITEBYTECODE yes
 # - openvswitch-ipsec
 # - ovn-vtep
 RUN INSTALL_PKGS=" \
-	openssl firewalld-filesystem \
+	openssl python3-pip firewalld-filesystem \
```
tssurya, line 36: I was just telling about the `pythonOpenSSL` stuff here: https://github.com/openshift/ovn-kubernetes/pull/1553#discussion_r1129434693
I am going to leave the same comment I left here: https://github.com/openshift/ovn-kubernetes/pull/1553#discussion_r1133798426

I think we need to start adding some comments on why packages are being added else its difficult to keep track

I would make this:
RUN INSTALL_PKGS=" \
	openssl firewalld-filesystem \
	python3-pip  \ # add comment here why you are adding this so that in future people know not to remove this.

and its not totally your fault since nobody did this in the past, but we need to start somewhere...

tssurya (changes requested):

Dockerfile
tssurya: yes lgtm thanks.

tssurya (approved):
/lgtm

jcaamano:
@pperiyasamy `Dockerfile.base` already installs `python3-openvswitch3.1` and pip is used to install the same library. Do we need the pip version for a specific reason? Otherwise, it is a possibility to remove the need to install ovs with pip and remove the need for pip altogether. Not a strong opinion. maybe it is useful to keep pip around, but I am surprised that we use it to install a library that should be already in place.

pperiyasamy:
> @pperiyasamy `Dockerfile.base` already installs `python3-openvswitch3.1` and pip is used to install the same library. Do we need the pip version for a specific reason? Otherwise, it is a possibility to remove the need to install ovs with pip and remove the need for pip altogether. Not a strong opinion. maybe it is useful to keep pip around, but I am surprised that we use it to install a library that should be already in place.

@jcaamano the pip package is needed by ovn-detrace command ([here](https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/cmd/ovnkube-trace/ovnkube-trace.go#L901)) as its used for installing its dependency ovs package. hence adding python3-pip into `Dockerfile.base` . But I'm not sure at what scenario this happens because ovs package is already there in the ovnk pods.

jcaamano:
So I think that we don't need to add python-pip if we know we satisfy ovnkube-trace dependencies in our base image.

For people using ovnkube-trace in other ways, the pyOpenSSL check could be changed to 
```
"pyOpenSSL": "if python -c 'import ssl; print(ssl.OPENSSL_VERSION)'; then echo 'true'; fi"
```

But I guess that needs to be an upstream change first?

tssurya:
@pperiyasamy explained the fix that we are doing upstream: https://github.com/ovn-org/ovn-kubernetes/pull/3549 and so we don't need this anymore.. SGTM.

PR #1622, opened 2023-03-31
OCPBUGS-6947: CARRY: use "prefer local" for annotated services

ricky-rav:
Openshift-only hack:

The router service is annotated with "traffic-policy.network.alpha.openshift.io/local-with-fallback" because, while it is defined with externalTrafficPolicy=Local, it expects the CNI to switch to externalTrafficPolicy=Cluster as soon as there are no local endpoints on any given node. 

SDN does this already, let's do it in OVNK as well to further reduce disruption time to openshift ingress during upgrades.

Closes #OCPBUGS-6947

### DIFF
--- a/go-controller/pkg/ovn/controller/services/lb_config.go
+++ b/go-controller/pkg/ovn/controller/services/lb_config.go
@@ -19,6 +19,7 @@ import (
 // magic string used in vips to indicate that the node's physical
 // ips should be substituted in
 const placeholderNodeIPs = "node"
+const localWithFallbackAnnotation = "traffic-policy.network.alpha.openshift.io/local-with-fallback"
 
 // lbConfig is the abstract desired load balancer configuration.
 // vips and endpoints are mixed families.
@@ -307,6 +308,28 @@ func buildPerNodeLBs(service *v1.Service, configs []lbConfig, nodes []nodeInfo) 
 					switchV4targetips = util.FilterIPsSlice(switchV4targetips, node.nodeSubnets(), true)
 					switchV6targetips = util.FilterIPsSlice(switchV6targetips, node.nodeSubnets(), true)
 				}
+				// OCP HACK BEGIN
+				zeroRouterV4LocalEndpoints := false
+				zeroRouterV6LocalEndpoints := false
+				if _, set := service.Annotations[localWithFallbackAnnotation]; set && config.externalTrafficLocal {
+					// if service is annotated and is ETP=local, fallback to ETP=cluster on nodes with no local endpoints:
+					// include endpoints from other nodes
+					if len(routerV4targetips) == 0 {
+						zeroRouterV4LocalEndpoints = true
+						routerV4targetips = config.eps.V4IPs
+					}
+					if len(routerV6targetips) == 0 {
+						zeroRouterV6LocalEndpoints = true
+						routerV6targetips = config.eps.V6IPs
+					}
+					if len(switchV4targetips) == 0 {
+						switchV4targetips = config.eps.V4IPs
+					}
+					if len(switchV6targetips) == 0 {
+						switchV6targetips = config.eps.V6IPs
+					}
+				}
+				// OCP HACK END
 				// at this point, the targets may be empty
 
 				// any targets local to the node need to have a special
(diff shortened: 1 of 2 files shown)


go-controller/pkg/ovn/controller/services/lb_config.go
```diff
@@ -302,6 +303,26 @@ func buildPerNodeLBs(service *v1.Service, configs []lbConfig, nodes []nodeInfo)
...
+					}
+					if len(routerV6targetips) == 0 {
+						zeroRouterV6LocalEndpoints = true
+						routerV6targetips = config.eps.V6IPs
+					}
+					if len(switchV4targetips) == 0 {
+						switchV4targetips = config.eps.V4IPs
+					}
+					if len(switchV6targetips) == 0 {
+						switchV6targetips = config.eps.V6IPs
```
trozet, line 322: so you set switchV6targetips to all addresses, but then if the service has ITP = local (few lines down) it will remove all of the endpoints except the local one. Seems wrong?

go-controller/pkg/ovn/controller/services/lb_config.go
```diff
@@ -397,10 +418,12 @@ func buildPerNodeLBs(service *v1.Service, configs []lbConfig, nodes []nodeInfo)
 						Targets: targets,
 					}
 
+					localWithFallback := (isv6 && zeroRouterV6LocalEndpoints) || zeroRouterV4LocalEndpoints
```
trozet, line 421: should this be:

```
localWithFallback := (isv6 && zeroRouterV6LocalEndpoints) || (!isv6 && zeroRouterV4LocalEndpoints)
```

Otherwise if isv6 is true, zeroRouterV6LocalEndpoints is false, zeroRouterV4LocalEndpoints is true, then localWithFallback will be true, even though this vip is ipv6, and the rule will an SNAT rule when  it should have really been an ETP local, no SNAT rule.

trozet:
@ricky-rav should we reference the jira OCPBUGS for disruption?

go-controller/pkg/ovn/controller/services/lb_config.go
tssurya: ITP=local and ETP=local will not be set on the same LB config. Will doubke check this in the coming week.

ricky-rav: in the end the router-default service is configured with `internalTrafficPolicy: Cluster`, so this downstream hack won't get executed with `internalTrafficPolicy:Local`

ricky-rav:
/hold cancel

@trozet @tssurya I've run upgrade jobs on the 4.13 PR and the disruption time for ingress on all cloud providers is down to 0:
https://github.com/openshift/ovn-kubernetes/pull/1623#issuecomment-1493893347

ricky-rav:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.13-upgrade-from-stable-4.12-e2e-gcp-ovn-rt-upgrade 10
/payload-aggregate periodic-ci-openshift-release-master-ci-4.13-upgrade-from-stable-4.12-e2e-azure-ovn-upgrade 10

ricky-rav:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-gcp-ovn-rt-upgrade 10
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-azure-ovn-upgrade 30

ricky-rav:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade 30

go-controller/pkg/ovn/controller/services/lb_config.go
tssurya: @ricky-rav : but its not just router default service who uses ITP=cluster and for them we have the special downstream HACK in some other part of the code :D 
In general users can create ITP=local as well.

tssurya: as spoken offline, the reason why I was saying ITP=local and ETP=local won't be set together is because of the way we do things in `buildServiceLBConfigs`, but I can see that @ricky-rav already moved the code to after that. So consider this comment. resolved.

go-controller/pkg/ovn/controller/services/lb_config.go
```diff
@@ -397,10 +420,12 @@ func buildPerNodeLBs(service *v1.Service, configs []lbConfig, nodes []nodeInfo)
 						Targets: targets,
 					}
 
+					localWithFallback := (isv6 && zeroRouterV6LocalEndpoints) || (!isv6 && zeroRouterV4LocalEndpoints)
```
tssurya, line 423: I'd have liked to see v6 unit test coverage that checks this logic but guess nobody should really be changing any of these hacks around...so we can skip this for now.

tssurya (approved):
/lgtm

tssurya:
> @trozet @tssurya I've run upgrade jobs on the 4.13 PR and the disruption time for ingress on all cloud providers is down to 0: [#1623 (comment)](https://github.com/openshift/ovn-kubernetes/pull/1623#issuecomment-1493893347)

THANK YOU @ricky-rav for ALL the hard work on this one and for picking up the commit and adding tests +++

PR #1627, opened 2023-04-04
OCPBUGS-11511: [release-4.13] static pod creation

bpickard22:
Clean backport of https://github.com/ovn-org/ovn-kubernetes/pull/3347

### DIFF
--- a/go-controller/pkg/cni/cni.go
+++ b/go-controller/pkg/cni/cni.go
@@ -2,10 +2,12 @@ package cni
 
 import (
 	"fmt"
+	"net"
+
+	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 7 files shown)


jcaamano:
@bpickard22 was the last commit required? Would it make lint fail? How was that not already happening? Out of curiosity.

bpickard22:
> @bpickard22 was the last commit required? Would it make lint fail? How was that not already happening? Out of curiosity.

yes the last commit is required to satisfy the lint. I think there was some mixup with the timing of merges and this commit went in after the 1.26 bump https://github.com/openshift/ovn-kubernetes/blob/8513b38e5a74c0c5f303dca30a27e6cc24bdf89c/go-controller/pkg/ovn/policy.go

trozet (approved):
/lgtm

PR #1631, opened 2023-04-06
OCPBUGS-11447: [release-4.13] Delete equivalent ACLs when searching by predicate.

npinaeva:
Equivalent acls just happen


trozet (approved):
/lgtm

npinaeva:
aws-windows failed 2 times with the same reason, filed a bug to check if the timeout is right https://issues.redhat.com/browse/OCPBUGS-12178

PR #1633, opened 2023-04-07
[release-4.12] OCPBUGS-8226: fix hybridOverlay DRIP in ICNIv1 pods

JacobTanenbaum:
starting with release-4.12 it was possible for nodes to use any address in there subnet for the hybridOverlay distributed router IP. However the ICNIv1 pods are still using only the .3 address for the hybrid Overlay router IP.

This fix queries the node for the correct Hybrid Overlay Distributed Router IP on pod creation and startup

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/pods.go
+++ b/go-controller/pkg/ovn/pods.go
@@ -6,6 +6,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	hotypes "github.com/ovn-org/ovn-kubernetes/go-controller/hybrid-overlay/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/metrics"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/ipallocator"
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/pods.go
```diff
@@ -82,14 +83,34 @@ func (oc *Controller) syncPods(pods []interface{}) error {
...
+			} else {
+				for _, subnet := range oc.lsManager.GetSwitchSubnets(pod.Spec.NodeName) {
+					hybridOverlayDRIPs = append(hybridOverlayDRIPs, util.GetNodeHybridOverlayIfAddr(subnet).IP)
+				}
+
+			}
+			for _, drip := range hybridOverlayDRIPs {
 				for _, route := range annotations.Routes {
-					if !route.NextHop.Equal(hybridOverlayIFAddr) {
+					if !route.NextHop.Equal(drip) {
```
trozet, line 108: why are you changing this part of the code? I thought this was to detect if a non ICNIv1 pod had an old route to the .3 and remove it? ICNIv1 pods wont be affected in this path because of !hasHybridAnnotation above

JacobTanenbaum: you are right, whoops and there should be no routes to other addresses on pods.

removed

trozet (approved):
/lgtm

PR #1641, opened 2023-04-14
OCPBUGS-11453: [release-4.10] Batch potentially big transaction on egress firewall ACLs migration

npinaeva:
Backport of https://github.com/openshift/ovn-kubernetes/pull/1629
Had to make some changes, this one requires a proper review

### DIFF
--- a/go-controller/pkg/libovsdbops/switch.go
+++ b/go-controller/pkg/libovsdbops/switch.go
@@ -12,6 +12,8 @@ import (
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
+
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // LOGICAL_SWITCH OPs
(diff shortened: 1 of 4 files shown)


npinaeva:
> @npinaeva any reason to skip the batch test?

because the tests were written for int, and now Batch only works for ACLs, so it would require re-writing the whole test. So I though I can cheat and say we know the function works based on the tests in the previous versions :P

jcaamano:
The batching & tests could be []interface{} based instead and retain some of its generic nature?
That might be interesting as we might end up backporting other batching.

npinaeva:
> The batching & tests could be []interface{} based instead and retain some of its generic nature? That might be interesting as we might end up backporting other batching.

Yeah I considered this option, but that would require copying the slice twice to convert types back and forth, so I thought we would just copy the Batch function for other  types in the future if needed.
We can go the []interface{} way, or try to rewrite tests for acls (which is not that difficult),  both approaches have their pros and cons, wdyt @jcaamano ?

jcaamano:
I see. The other only thing I can think of is if a batch function that gives you the indexes of the array you need to work with:

```
// the meat goes here
func Batch(batchSize int, dataSize int, eachFn func(from, to int) error) error

// adapter function per type that uses the previous function
func BatchACLs(batchSize int, data []nbdb.ACL, eachFn func([]nbdb.ACL) error) error {
```

Then we can just test Batch.

npinaeva:
> I see. The other only thing I can think of is a batch function that gives you the indexes of the array you need to work with:
> 
> ```
> // the meat goes here
> func batch(batchSize int, dataSize int, eachFn func(from, to int)) error
> 
> // adapter function per type that uses the previous function
> func BatchACLs(batchSize int, data []nbdb.ACL, eachFn func([]nbdb.ACL) error) error {
> ```
> 
> Then we can just test batch.

I didn't want to introduce any new logic (which switching to indexes will require), I prefer to stay as close to the initial implementation as possible when it comes to backporting. So the most straightforward way for now I would say it to just make Batch type pre-defined. I updated the tests, since it is not that many changes I believe, lmk what you think

PR #1663, opened 2023-05-01
Check if the status of a pod before trying to get its ip

bpickard22:
Clean cherry-pick
Fixes the ips not found issue found here -> https://issues.redhat.com/browse/OCPBUGS-11567
We should not try to get the ip of a pod during nat cleanup if the pod is in one of two states, completed, or not scheduled, because the pod will not have an ip allocated to it during that time. This is a transient issue that will eventually be resolved by the retry loop, but we should not rely on that, hence this fix

Signed-off-by: Ben Pickard <bpickard@redhat.com>
(cherry picked from commit 432340548ffa1ecb44c8ee6fa6a29ba835caff12)


bpickard22:
@tssurya PTAL

PR #1664, opened 2023-05-02
OCPBUGS-13716: Use ovsver and ovnver to infer the short version numbers for ovs and ovn

aleskandro:
This PR proposes to add a build argument to drive the proper openvswitch package to install from the RPM repositories.

This will ease the versions bump and will be helpful for some OKD-related builds when the RPM repositories are not in sync, and some packages version might miss.

cc @vrutkovs @LorbusChris

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -16,12 +16,16 @@ ARG ovsver=3.1.0-10.el9fdp
 ARG ovnver=23.03.0-7.el9fdp
 
 RUN INSTALL_PKGS="iptables" && \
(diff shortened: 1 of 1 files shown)


trozet (approved):
/lgtm

dcbw:
Gotta be really careful here and check with ART... their tooling does some magic with the Dockerfiles and substitutes commands; we need to make sure the sed stuff here will work.

@jupierce @joepvd  ?

jupierce:
lgtm. The only limitation I'm aware of is that our lexer doesn't like `[[ ]]` bash test syntax.

PR #1670, opened 2023-05-08
Bug OCPBUGS-12792: [release-4.10] Delete equivalent ACLs when searching by predicate.

npinaeva:
Backport of https://github.com/openshift/ovn-kubernetes/pull/1666
I had to rewrite almost everything in this PR, needs proper review

### DIFF
--- a/go-controller/pkg/libovsdbops/acl.go
+++ b/go-controller/pkg/libovsdbops/acl.go
@@ -10,6 +10,8 @@ import (
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
+
+	"k8s.io/klog/v2"
 )
 
 // getACLName returns the ACL name if it has one otherwise returns
@@ -69,12 +71,15 @@ func findACL(nbClient libovsdbclient.Client, acl *nbdb.ACL) error {
 		return fmt.Errorf("can't find ACL by equivalence %+v: %v", *acl, err)
 	}
 
-	if len(acls) > 1 {
-		return fmt.Errorf("unexpectedly found multiple equivalent ACLs: %+v", acls)
-	}
-
 	if len(acls) == 0 {
 		return libovsdbclient.ErrNotFound
+	}
+
+	if len(acls) > 1 {
+		err = deleteDuplicateACLs(nbClient, acls)
+		if err != nil {
+			return fmt.Errorf("unexpectedly found multiple equivalent ACLs, failed to delete: %w", err)
+		}
 	}
 
 	acl.UUID = acls[0].UUID
@@ -249,6 +254,11 @@ func DeleteACLs(nbClient libovsdbclient.Client, acls []nbdb.ACL) error {
 		opModels = append(opModels, OperationModel{
 			Model:          &acl,
 			ModelPredicate: func(item *nbdb.ACL) bool { return IsEquivalentACL(item, &acl) },
+			// don't error on equivalent acls
+			// DeleteACLs will always be a noop, because deleted ACLs should be de-referenced in another transaction,
+			// and then they are already garbage-collected, or deleting ACL without dereferencing will return error.
+			// DeleteACLs has only effect for unit tests.
+			BulkOp: true,
 		})
 	}
 
@@ -269,4 +279,19 @@ func UpdateACLLogging(nbClient libovsdbclient.Client, acl *nbdb.ACL) error {
 
 	_, err = TransactAndCheck(nbClient, ops)
 	return err
+}
+
+func deleteDuplicateACLs(nbClient libovsdbclient.Client, duplicateACLs []nbdb.ACL) error {
+	if len(duplicateACLs) > 1 {
+		klog.Warningf("Found multiple acls equivalent to %+v, cleanup", duplicateACLs[0])
+		err := DeleteACLsFromAllPortGroups(nbClient, duplicateACLs[1:]...)
+		if err != nil {
+			return fmt.Errorf("failed to delete duplicate acls: delete from port groups failed: %w", err)
+		}
+		err = RemoveACLsFromAllSwitches(nbClient, duplicateACLs[1:])
+		if err != nil {
+			return fmt.Errorf("failed to delete duplicate acls: delete from switches failed: %w", err)
+		}
+	}
+	return nil
 }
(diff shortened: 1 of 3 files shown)


go-controller/pkg/libovsdbops/acl.go
```diff
@@ -255,6 +261,13 @@ func DeleteACLs(nbClient libovsdbclient.Client, acls []nbdb.ACL) error {
 	m := NewModelClient(nbClient)
 	err := m.Delete(opModels...)
 	if err != nil {
+		if strings.Contains(err.Error(), "unexpectedly found multiple results for provided predicate") {
+			// run findACL for every acl in the list, all the duplicate acls will be deleted as a side effect
+			for _, acl := range acls {
+				acl := acl
+				_ = findACL(nbClient, &acl)
+			}
+		}
```
jcaamano, line 270: Why not use BulkOp set to true instead of this?

Wouldn't this leave 1 ACL without being deleted?

go-controller/pkg/libovsdbops/acl.go
```diff
@@ -270,3 +283,18 @@ func UpdateACLLogging(nbClient libovsdbclient.Client, acl *nbdb.ACL) error {
...
 }
+
+func deleteDuplicateACLs(nbClient libovsdbclient.Client, duplicateACLs []nbdb.ACL) error {
+	if len(duplicateACLs) > 1 {
+		klog.Warningf("Found multiple acls equivalent to %+v, cleanup", duplicateACLs[0])
+		err := RemoveACLsFromAllPortGroups(nbClient, duplicateACLs[1:])
+		if err != nil {
+			return fmt.Errorf("failed to delete duplicate acls: delete from port groups failed: %w", err)
+		}
+		err = RemoveACLsFromAllSwitches(nbClient, duplicateACLs[1:])
```
jcaamano, line 294: In the 4.11 PR you remove from the first to the second to last. Here you remove from the second to the last. I would guess it doesn't matter but I would still keep it equal.

go-controller/pkg/libovsdbops/portgroup.go
```diff
@@ -313,6 +316,62 @@ func DeleteACLsFromPortGroupOps(nbClient libovsdbclient.Client, ops []libovsdb.O
...
+			return fmt.Errorf("failed to mutate port group: %w", err)
+		}
+		ops = append(ops, op...)
+	}
+	_, err = TransactAndCheck(nbClient, ops)
+	if err != nil {
+		return fmt.Errorf("failed to transact: %v", err)
+	}
+	return nil
+}
```
jcaamano, line 357: Why not just backport `DeleteACLsFromAllPortGroups` from the 4.11 PR?

jcaamano:
Out of curiosity, do we know how duplicated ACLS that we are removing here happened? Asking because if we are using `IsEquivalentACL` to find the duplicates, how had not `IsEquivalentACL` prevented the duplicates from happening in the first place? Did those duplicates happen in yet a previous version?

go-controller/pkg/libovsdbops/portgroup.go
npinaeva: good point, I didn't know all the same objects exist in this version with capitalized names :P

go-controller/pkg/libovsdbops/acl.go
npinaeva: there is a difference, because in the newer versions deleting duplicate acls is a side effect since we can't just safely retry the original operation. In this version findACL is a separate operation, and we can delete duplicate acls and return only 1 at the same time. It uses `acl.UUID = acls[0].UUID` to set the found UUID, so I decided to leave the first acl and delete all the others to leave this part of the code without changes

npinaeva:
>Out of curiosity, do we know how duplicated ACLS that we are removing here happened? Asking because if we are using IsEquivalentACL to find the duplicates, how had not IsEquivalentACL prevented the duplicates from happening in the first place? Did those duplicates happen in yet a previous version?

In general, it can happen when we update ACLs with UUID set, and the predicate is not used, another option is when acl becomes duplicate after the update, so the original ACL had `externalIDs=map1`, which was unique, but the update operation itself may update `externalIDs=map2`, and `map2` may already exist. So on the next predicate search we will find duplicate acls.
In this last bug, 4.9 `ovn-nbctl` is the source of duplicate acls (there are many ways of how that can happen in 4.9), and on update to 4.10 we see the duplicate acls error for the first time.
And thanks for review @jcaamano!

go-controller/pkg/libovsdbops/acl.go
npinaeva: now my memory comes back,  in this version we still have `DeleteACLs`, but we switched to `de-referencing only` in the later versions. Added a comment for that op

go-controller/pkg/libovsdbops/acl.go
```diff
@@ -249,6 +254,11 @@ func DeleteACLs(nbClient libovsdbclient.Client, acls []nbdb.ACL) error {
 		opModels = append(opModels, OperationModel{
 			Model:          &acl,
 			ModelPredicate: func(item *nbdb.ACL) bool { return IsEquivalentACL(item, &acl) },
+			// don't error on equivalent acls
+			// DeleteACLs will always be a noop, because deleted ACLs should be de-referenced in another transaction,
+			// and then they are already garbage-collected, or deleting ACL without dereferencing will have no effect.
+			// DeleteACLs has only effect for unit tests.
```
jcaamano, line 260: Do you mean to say we only use DeleteACLs for unit tests?

Note that `deleting ACL without dereferencing` will lead to an error, as ACLs references are strong. So not sure `will have no effect` is the coirrect description.

Sorry If I am not understanding what you mean.

go-controller/pkg/libovsdbops/acl.go
jcaamano: Confused as well here. I was comparing what I see in the 4.11 PR:
```
err = DeleteACLsFromAllPortGroups(nbClient, duplicateACLs[:len(duplicateACLs)-1]...)
```
to what I see here
```
err := DeleteACLsFromAllPortGroups(nbClient, duplicateACLs[1:]...)
```

npinaeva: that is what I tried to explain in the comment above: if I leave `duplicateACLs[:len(duplicateACLs)-1]`, I need to return acl after cleanup as `acl.UUID = acls[len(acls) - 1].UUID`. So the difference is where you make the change, in `findACL` or in `deleteDuplicateACLs`. I just prefer leaving the first element, since it looks cleaner to me.

PR #1672, opened 2023-05-15
OCPBUGS-13598: Call SyncEndpoints from AddService

tssurya:
There is a race condition observed when the
endpointslice is added before the service.
We will not pass
https://github.com/ovn-org/ovn-kubernetes/blob/e7a92dbdab7d2f0c056ae2c6e549140c1809c7c0/go-controller/pkg/node/healthcheck_service.go#L131 this check and hence will not call `syncEndPointSlices` since service doesn't exist yet. Later when we do
AddService, we were not calling syncEndpoints and so we end up in a situation where the localEndpoints count is not updated.

This PR fixes that by moving the logic to call syncEndpoints to AddService. In the reverse case of endpointslice being added after the service, its possible we call the syncEndpoints from two places - both AddService and AddEndpointSlice, but better to be functionally correct and avoid race before worrying about performance...

(cherry picked from commit dea653ab12d058fcc011529a2d1792cb69fc0a28)
Conflicts because https://github.com/ovn-org/ovn-kubernetes/pull/3485 does not exist.

(cherry picked from commit 79bf59d6c94cd5842d7188ccf758dd7c6d6ab39d)

**- How to verify it**

### DIFF
--- a/go-controller/pkg/node/healthcheck_service.go
+++ b/go-controller/pkg/node/healthcheck_service.go
@@ -46,7 +46,20 @@ func (l *loadBalancerHealthChecker) AddService(svc *kapi.Service) error {
 		defer l.Unlock()
 		name := ktypes.NamespacedName{Namespace: svc.Namespace, Name: svc.Name}
 		l.services[name] = uint16(svc.Spec.HealthCheckNodePort)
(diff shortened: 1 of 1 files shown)


trozet (approved):
/lgtm

PR #1698, opened 2023-06-06
matched comment to new function name

juanino:
All other comments match function names. There was a change in the function that was significant so referring to the old function in the comment is probably worth fixing to reduce confusion.

Simple comment change

**- How to verify it**
n/a

one liner change to a comment.

### DIFF
--- a/go-controller/pkg/cni/utils.go
+++ b/go-controller/pkg/cni/utils.go
@@ -59,7 +59,7 @@ func (c *ClientSet) getPod(namespace, name string) (*kapi.Pod, error) {
 	return pod, err
 }
 
-// GetPodAnnotations obtains the pod UID and annotation from the cache or apiserver
+// GetPodWithAnnotations obtains the pod UID and annotation from the cache or apiserver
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

go-controller/pkg/cni/utils.go
```diff
@@ -59,7 +59,7 @@ func (c *ClientSet) getPod(namespace, name string) (*kapi.Pod, error) {
 	return pod, err
 }
 
-// GetPodAnnotations obtains the pod UID and annotation from the cache or apiserver
+// GetPodWithAnnotations obtains the pod UID and annotation from the cache or apiserver
```
tssurya, line 62: please submit a PR upstream first: https://github.com/ovn-org/ovn-kubernetes and then it will be brought down from there automatically.

tssurya (changes requested):
Thanks for your contribution, Please close this PR and open one upstream first. We will then bring that in automatically.

juanino:
ok, if you have directions let me know, otherwise i will RTFM-- newb.

PR #1708, opened 2023-06-14
Dockerfile: bump to ovn23.03-23.03.0-69 for LB templates

dcbw:
We need at >= 23.03-14 for LB templates; bump to latest to get other fixes too.


dcbw:
/retest-required

Windows jobs failed on deprovision, not during the tests.

PR #1713, opened 2023-06-16
OCPBUGS-14982: Validate port before deleting conntrack flow

abhat:
For services and endpoints, make sure that the protocol and port are valid before deleting conntrack flows.
Under certain circumstances, the port may be the default value (meaning 0), indicating that all conntrack flows for a given IP shall be deleted. However, for endpoints and services, only ever delete exact matches. Otherwise, ovnkube-node might remove conntrack flows inserted by unrelated components (such as iptables rules). If for any reason protocol is invalid, skip conntrack deletion instead.

Reported-at: https://issues.redhat.com/browse/OCPBUGS-14769
Signed-off-by: Andreas Karis <ak.karis@gmail.com>
(cherry picked from commit eacdaa721dc3d4bd7a1b5b248aa94ac22a6d5da6)


tssurya (commented):
LGTM
Leaving approval to tim since its not clean pick.

PR #1722, opened 2023-06-23
OCPBUGS-15372: [release-4.13] Fix network policy to work with long namespace names

npinaeva:
Add default deny acl externalID to avoid duplicate acls for namespaces with long names.
every namespace has 4 default deny acls: ingress/egress and deny/allow. Direction was in the ExternalIDs before, but action was a suffix in acl name, that gets truncated for long namespace names, therefore deny and allow acls for the same direction would have the same name and ExternalIDs. New ExternalID should fix this case.

Multicast acls don't need extra ID, therefore I renamed the old function to `getDirectionPolicyACLExternalIDs` and multicast still uses that.
To check the difference on restart, you can run the new test on the old implementation and see duplicate acls error in the logs. Also create namespace with long name + networkpolicy and restart ovn-k master can be used as reproducer

### DIFF
--- a/go-controller/pkg/ovn/policy.go
+++ b/go-controller/pkg/ovn/policy.go
@@ -26,9 +26,12 @@ import (
 	utilnet "k8s.io/utils/net"
 )
 
+type netpolDefaultDenyACLAction string
+
 const (
 	// defaultDenyPolicyTypeACLExtIdKey external ID key for default deny policy type
-	defaultDenyPolicyTypeACLExtIdKey = "default-deny-policy-type"
+	defaultDenyPolicyTypeACLExtIdKey   = "default-deny-policy-type"
+	defaultDenyPolicyActionACLExtIdKey = "default-deny-policy-action"
 	// l4MatchACLExtIdKey external ID key for L4 Match on 'gress policy ACLs
 	l4MatchACLExtIdKey = "l4Match"
 	// ipBlockCIDRACLExtIdKey external ID key for IP block CIDR on 'gress policy ACLs
@@ -52,8 +55,10 @@ const (
 	// staleArpAllowPolicyMatch "was" the old match used when creating default allow ARP ACLs for a namespace
 	// NOTE: This is succeed by arpAllowPolicyMatch to allow support for IPV6. This is currently only
 	// used when removing stale ACLs from the syncNetworkPolicy function and should NOT be used in any main logic.
-	staleArpAllowPolicyMatch = "arp"
-	allowHairpinningACLID    = "allow-hairpinning"
+	staleArpAllowPolicyMatch                            = "arp"
+	allowHairpinningACLID                               = "allow-hairpinning"
+	denyAction               netpolDefaultDenyACLAction = "deny"
+	allowAction              netpolDefaultDenyACLAction = "allow"
 )
 
 // defaultDenyPortGroups is a shared object and should be used by only 1 thread at a time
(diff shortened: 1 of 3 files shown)


go-controller/pkg/ovn/policy.go
```diff
@@ -277,6 +282,28 @@ func (oc *DefaultNetworkController) updateStaleDefaultDenyACLNames(npType knet.P
...
+	for _, acl := range staleACLs {
+		var aclAction netpolDefaultDenyACLAction
+		if acl.Priority == types.DefaultDenyPriority {
+			aclAction = denyAction
+		} else {
+			aclAction = allowAction
+		}
+		acl.ExternalIDs[defaultDenyPolicyActionACLExtIdKey] = string(aclAction)
+	}
+	return libovsdbops.CreateOrUpdateACLs(oc.nbClient, staleACLs...)
```
trozet, line 304: Do we want to use CreateOrUpdateACLs here? The reason I ask is because that is the path where we will check is equivalentACLs predicate. So if we have the scenario where we have 2 duplicates in the cache, we would fail here, then the deleteDuplicate function would get called, syncNetworkPolicies would fail (and hopefully be retried).

I think it might be better to just directly do an update txn here, since we already fetched hte ACLs from the cache and this is a sync function so no one else should be messing with ACLs.

npinaeva: staleACLs are from the db, so they have UUID set, that is why predicate won't be used, update will happen based on UUID. You need to build update ops anyway, not sure how you see "directly do an update txn here", the best I can do is use `UpdateACLsOps` instead of `CreateOrUpdateACLsOps`, but the result should be exactly the same if no one deletes acls in between (which shouldn't happen as you said)

trozet: ah right we have uuid, so shoudl be good then

go-controller/pkg/ovn/policy_test.go
```diff
@@ -895,6 +888,44 @@ var _ = ginkgo.Describe("OVN NetworkPolicy Operations", func() {
...
+			app.Action = func(ctx *cli.Context) error {
+				namespace1 := *newNamespace("abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijk")
+				namespace2 := *newNamespace(namespaceName2)
+				networkPolicy := getMatchLabelsNetworkPolicy(netPolicyName1, namespace1.Name,
+					namespace2.Name, "", true, true)
+				// start with stale ACLs
+				// this function will cut off the namespace name to have no suffix
+				gressPolicyInitialData := getPolicyData(networkPolicy, nil, []string{namespace2.Name},
+					nil, "", true)
+				defaultDenyInitialData := getDefaultDenyData(networkPolicy, nil, "", true)
```
trozet, line 902: getDefaultDenyData now calls getDenyACLExternalIDs which has the new external Id direction in it already in the initial data. Dont you want the initial data to have the stale version?

npinaeva: last argument in both `getPolicyData` and `getDefaultDenyData` is `stale`, when set to true acls will be update the stale version, check `getStaleDefaultACL` change that deletes new externalID

trozet: then...
```
func getStaleDefaultACL(acls []*nbdb.ACL) []*nbdb.ACL {
	for _, acl := range acls {
		acl.Options = nil
		acl.Direction = nbdb.ACLDirectionToLport
		delete(acl.ExternalIDs, defaultDenyPolicyActionACLExtIdKey)
	}
	return acls
}
```

I don't really like this, but its just an old unit test...so I'm OK with it.

npinaeva: can you elaborate on how you would like it to look instead? maybe we can change it upstream

PR #1731, opened 2023-07-02
[release-4.13] OCPBUGS-15651: Improve syncNodes to remove stale data

flavio-fernandes:
Partial removal of nodes can leave OVN database with the gateway logical router and external logical switch of the removed node. This commit improves syncNode functionality, so these cases are also taken into consideration.

When deleting a port from a logical switch or router, it is possible that the logical switch or router themselves have been deleted. This commit changes the behavior of delete so it is not an error when if that happens.

Conflicts:
go-controller/pkg/ovn/master.go - cleanupNodeResources => deleteNode

Reported-at: https://issues.redhat.com/browse/OCPBUGS-13744
(cherry picked from commit 487bba32a5f6256e53b77f8f43808c74763507cf)

### DIFF
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -847,6 +847,7 @@ func (oc *DefaultNetworkController) addUpdateNodeEvent(node *kapi.Node, nSyncs *
 	if err != nil {
 		oc.recordNodeErrorEvent(node, err)
 	}
+
 	return err
 }
 
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -20,6 +20,7 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdbops"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/sbdb"
+	"github.com/pkg/errors"
 
 	hotypes "github.com/ovn-org/ovn-kubernetes/go-controller/hybrid-overlay/pkg/types"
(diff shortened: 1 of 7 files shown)


go-controller/pkg/ovn/master.go
```diff
@@ -639,11 +640,46 @@ func (oc *DefaultNetworkController) syncNodes(nodes []interface{}) error {
 	if err != nil {
 		return fmt.Errorf("failed to get node logical switches which have other-config set: %v", err)
 	}
+
+	staleNodes := sets.NewString()
```
bpickard22, line 644: sets.New[string]

bpickard22: could be on the old sets package in 4.13 so that might not be accepted but that is the proper way to use the sets package if its available

bpickard22 (commented):
/lgtm

go-controller/pkg/ovn/master.go
flavio-fernandes: Ah, generics notation. I think that is okay to use `NewString` and stay consistent in this file. Like line 692 below. Agree?

bpickard22: yes I think thats fine

go-controller/pkg/ovn/master.go
```diff
@@ -639,11 +640,46 @@ func (oc *DefaultNetworkController) syncNodes(nodes []interface{}) error {
...
 	for _, nodeSwitch := range nodeSwitches {
 		if !foundNodes.Has(nodeSwitch.Name) {
```
flavio-fernandes, line 646: **Note**: in ovn-ic, there is also another check here to ensure transit switches are never deleted by this logic.

trozet (approved):
/lgtm

PR #1732, opened 2023-07-03
[release-4.12] OCPBUGS-13744: Improve syncNodes to remove stale data

flavio-fernandes:
Cherry pick from https://github.com/openshift/ovn-kubernetes/pull/1731

Partial removal of nodes can leave OVN database with the gateway logical router and external logical switch of the removed node. This commit improves syncNode functionality, so these cases are also taken into consideration.

When deleting a port from a logical switch or router, it is possible that the logical switch or router themselves have been deleted. This commit changes the behavior of delete so it is not an error when if that happens.

Conflicts:
- go-controller/pkg/ovn/base_network_controller.go 
    - deleteNodeLogicalNetwork located in .../pkg/ovn/master.go
- go-controller/pkg/ovn/master_test.go

Signed-off-by: Flavio Fernandes <flaviof@redhat.com>
(cherry picked from commit 9fc87b22dad9919fd432d8dd20ed87f1605184e2)

PR #1734, opened 2023-07-03
[release-4.11] OCPBUGS-15731,OCPBUGS-15745: Improve syncNodes to remove stale data

flavio-fernandes:
Cherry pick of: https://github.com/openshift/ovn-kubernetes/pull/1732

Partial removal of nodes can leave OVN database with the gateway logical router and external logical switch of the removed node. This commit improves syncNode functionality, so these cases are also taken into consideration.

When deleting a port from a logical switch or router, it is possible that the logical switch or router themselves have been deleted. This commit changes the behavior of delete so it is not an error when if that happens.

Conflicts:
- go-controller/pkg/ovn/master_test.go


trozet (approved):
/lgtm

/label backport-risk-assessed

PR #1748, opened 2023-07-07
Dockerfile: build both RHEL8 and RHEL9 shims

dcbw:
Since the shim (ovn-k8s-cni-overlay) gets copied to the host OS and executed in the host mount namespace by CRIO/Multus it needs to be runtime compatible with the host OS. Running a RHEL9-built shim on a RHEL8 system doesn't work due to different shared library dependencies between the two OS versions.

So... build two separate versions of ovn-kubernetes in different layers, and copy the RHEL8 shim into a special location where our container startup scripts can find it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,14 +4,18 @@
 #
 # The standard name for this image is ovn-kube
 
+# Build RHEL-9 binaries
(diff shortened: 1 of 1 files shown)


mrunalp (approved):
/lgtm

wking:
/retitle OCPBUGS-15962: Dockerfile: build both RHEL8 and RHEL9 shims

PR #1759, opened 2023-07-14
[release-4.9] OCPBUGS-15865: Dockerfile: build both RHEL7 and RHEL8 shims

jcaamano:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,14 +9,18 @@
 # are built in this Dockerfile and included in the image (instead of the rpm)
 #
 
+# Build RHEL-8 binaries
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9 AS builder
-
 WORKDIR /go/src/github.com/openshift/ovn-kubernetes
 COPY . .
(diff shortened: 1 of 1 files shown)


jcaamano:
@dcbw PTAL

Dockerfile
```diff
@@ -9,15 +9,19 @@
...
 #
 
+# Build RHEL-8 binaries
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9 AS builder
-
 WORKDIR /go/src/github.com/openshift/ovn-kubernetes
 COPY . .
-
-# build the binaries
 RUN cd go-controller; CGO_ENABLED=0 make
```
dcbw, line 16: @jcaamano since we're building with CGO_ENABLED=1 anyway, should we just change it here?

dcbw:
@jcaamano did you mean to leave in the contrib/kind.yaml bits?

jcaamano:
> @jcaamano did you mean to leave in the contrib/kind.yaml bits?

no :facepalm:

PR #1760, opened 2023-07-14
Dockerfile: build both RHEL8 and RHEL9 shims

abhat:
Since the shim (ovn-k8s-cni-overlay) gets copied to the host OS and executed in the host mount namespace by CRIO/Multus it needs to be runtime compatible with the host OS. Running a RHEL9-built shim on a RHEL8 system doesn't work due to different shared library dependencies between the two OS versions.

So... build two separate versions of ovn-kubernetes in different layers, and copy the RHEL8 shim into a special location where our container startup scripts can find it.

(cherry picked from commit 528654874d921ee5054cd668e551ec5e1700b3bc)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,14 +4,19 @@
 #
 # The standard name for this image is ovn-kube
 
+# Build RHEL-9 binaries
 FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.20-openshift-4.14 AS builder
 
 WORKDIR /go/src/github.com/openshift/ovn-kubernetes
 COPY . .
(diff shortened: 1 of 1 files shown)


wking:
This is picking #1760 forward to the development branch, so retroactively associating with the 4.14.0 tracker:

/retitle OCPBUGS-15961: Dockerfile: build both RHEL8 and RHEL9 shims

Associating a pull request with a bug post-merge is too late to get that information into the merge commit message for the release controller, but :shrug:, not much we can do about that now, and it will still get the link to this pull request into the Jira ticket.

PR #1772, opened 2023-07-20
Hack azure vip

trozet:

**- How to verify it**


go-controller/pkg/ovn/gateway_init.go
```diff
@@ -556,6 +556,7 @@ func (oc *DefaultNetworkController) addPolicyBasedRoutes(nodeName, mgmtPortIP st
 	}
 
 	matches := sets.New[string]()
+	otherHostAddrs = append(otherHostAddrs, "10.0.0.4")
```
tssurya, line 559: haha but this is the fix for the customer bug i have... how is this related to the node disruptions we are seeing :D ?

tssurya:
/close
IC era done and dusted, stale PR cleanup

PR #1776, opened 2023-07-24
[release-4.8] OCPBUGS-16208: Dockerfile: build both RHEL7 and RHEL8 shims

jcaamano:

(cherry picked from commit 932ef31ed728426b7bb8c52b52ed52cb38d84a1d)

**- How to verify it**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,14 +9,18 @@
 # are built in this Dockerfile and included in the image (instead of the rpm)
 #
 
+# Build RHEL-8 binaries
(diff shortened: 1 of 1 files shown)


dcbw:
/override ci/prow/e2e-aws-ovn
/override ci/prow/e2e-azure-ovn
/override ci/prow/e2e-vsphere-ovn

Need 4.8 backport of https://github.com/openshift/origin/pull/27422, plus one image pull backoff issue for ingress-canary

PR #1778, opened 2023-07-24
[release-4.6] OCPBUGS-16210: Dockerfile: build both RHEL7 and RHEL8 shims

jcaamano:

(cherry picked from commit 932ef31ed728426b7bb8c52b52ed52cb38d84a1d) (cherry picked from commit 20f8d238c00fdfbb9c93689ed0c7f1d8b360a255) (cherry picked from commit 37a5ded3eec9faed3d187b9418303aa93dc7da64)

**- How to verify it**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,14 +9,18 @@
 # are built in this Dockerfile and included in the image (instead of the rpm)
 #
 
+# Build RHEL-8 binaries
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
-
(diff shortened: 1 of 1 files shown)


jcaamano:

Gather extra failed which I don't think has anything to do with this PR, but all tests passed.

dcbw:

error: unable to connect to image repository virthost.ostest.test.metalkube.org:5000/localimages/local-release-image@sha256:40175d5690e2d4691fc95e1cc1881892627d39ef29330d3b778f80aa981d5abe: Get "https://virthost.ostest.test.metalkube.org:5000/v2/": Internal Server Error

PR #1780, opened 2023-07-25
OCPBUGS-15811: Use NodePrimaryIp to set host-addresses in DPU mode

SalDaniele:
When running a 2-cluster design in DPU mode, currently host-addresses is set to the IP address of the Infrastructure node (i.e. Bluefield node) running of behalf of the tenant cluster. This will cause all NodePort service to fail in DPU mode.

Instead, we should calculate the host-addresses using the NodePrimaryIP which will set the host-address to the Node IP of the tenant node.

**- How to verify it**

### DIFF
--- a/go-controller/pkg/node/node_ip_handler_linux.go
+++ b/go-controller/pkg/node/node_ip_handler_linux.go
@@ -10,11 +10,14 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dcbw:
Let's take this patch through https://github.com/openshift/ovn-kubernetes/pull/1784

PR #1791, opened 2023-07-31
[WIP] Changes IC chassis update to be a mutate operation

trozet:
By updating the chassis with incomplete information, it was causing ovn-controler to rewrite the values in the DB, and furthermore causing northd to rewrite logical flows during ovnkube-master restart. The notable consequence of this would be that access to a service via loadbalancer type (using LB templates) would be momentarily disrupted. This is the root cause for the outage we see in Azure upgrades.

**- How to verify it**


tssurya:
/close

IC era done and dusted, stale PR cleanup

PR #1603, opened 2023-03-23
OCPBUGS-10799: [release-4.13] Netpol performance: shared peer address sets, add metrics

npinaeva:
Bring required changes to 4.13
perf metrics  https://github.com/ovn-org/ovn-kubernetes/pull/3450
downstream https://github.com/openshift/ovn-kubernetes/pull/1556

shared address sets https://github.com/ovn-org/ovn-kubernetes/pull/3329
downstream https://github.com/openshift/ovn-kubernetes/pull/1574

No conflicts, last commit fixes a "conflic" because k8s 1.26 bump is not yet merged to 4.13

PR #1814, opened 2023-08-14
OCPBUGS-17675: Don't return error on delete event that doesn't ave a chance to succeed.

npinaeva:
Besides unneeded logging it may cause update event that was received after delete event to be never handled, since retry framework will not proceed to the update handler until delete handler succeeds. Was fixed in newer versions with 120140fd670d8358959e6ea062564edd9d7d8841 (pod_selector_address_set)
Merged downstream here https://github.com/openshift/ovn-kubernetes/pull/1603 (4.13), https://github.com/openshift/ovn-kubernetes/pull/1603/files#diff-7375899468809f8394c8c50f436167aa32368e3dac1eb0c25d05e6b9b0e2cbe3R354-R359 
loglevel was updated to warning here https://github.com/ovn-org/ovn-kubernetes/commit/8e95a642698a763fb885fe318d3e9b20d8502205

### DIFF
--- a/go-controller/pkg/ovn/policy.go
+++ b/go-controller/pkg/ovn/policy.go
@@ -1525,7 +1525,10 @@ func (oc *Controller) handlePeerPodSelectorDelete(np *networkPolicy, gp *gressPo
 	if util.PodCompleted(pod) {
 		ips, err := util.GetAllPodIPs(pod)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


dcbw:
/override ci/prow/e2e-aws-ovn-local-to-shared-gateway-mode-migration

PR #1815, opened 2023-08-14
Fix encap port deadlock

tssurya:
If geneve encap port is different from the default, then we configure the SBDB chassis with the new port. After this we proceed to annotate the zone-name in ovnkube-node.
However in IC deployments, ovnkube-controller does not come up until the zone-name annotation is tagged on the nodes: https://github.com/ovn-org/ovn-kubernetes/blob/4bfacf905e84e9e1e37e3a9dda5b5321df22bc4f/go-controller/pkg/network-controller-manager/network_controller_manager.go#L337 This means ovnkube-controller is waiting on ovnkube-node to annotate stuff before starting to configure NBDB stuff and SBDB chassis creation won't happen because of that.

On the other hand, ovnkube-node is timing out trying to find the sbdb chassis to set the encap port since chassis is not existing in sbdb yet.

We end up in a deadlock situation.

(cherry picked from commit b18f577ebe2dc3ba3b3bb2e580ef8c6ecb6c0dd2)

**- How to verify it**

### DIFF
--- a/go-controller/pkg/node/default_node_network_controller.go
+++ b/go-controller/pkg/node/default_node_network_controller.go
@@ -324,27 +324,6 @@ func setupOVNNode(node *kapi.Node) error {
 	if err != nil {
 		return fmt.Errorf("error setting OVS external IDs: %v\n  %q", err, stderr)
 	}
-	// If EncapPort is not the default tell sbdb to use specified port.
-	if config.Default.EncapPort != config.DefaultEncapPort {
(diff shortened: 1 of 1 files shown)


flavio-fernandes:
/test prow/unit
is this a flake?

flavio-fernandes:
/test unit
flake?

flavio-fernandes:
> /test unit flake?

nope; not flake. ;P

PR #1838, opened 2023-08-28
OCPBUGS-17979: Fix egress firewall CRD

abhat:
As @tssurya found the min/max of 1 for destinations was hardcoded in the crd j2 file and was never in the types.go.

Fixed commit to remove node-selector feature which doesn't apply in 4.12

Co-authored-by: Aniket Bhat <anbhat@redhat.com>
(cherry picked from commit 41046e8c27715bd08ce8090ac8ae06d3cfc9e75c)


abhat:
@huiran0826 @trozet ptal. cc @anuragthehatter

abhat:
@trozet it is not a clean cherry-pick since there were node selector related changes in the original commit. I had to remove those and only have the minProperties=1/maxProperties=1 fix.

PR #1841, opened 2023-08-30
[release-4.13] DNM: Bump libovsdb client to send echo in async way

pperiyasamy:
This would send echo to ovsdb server in async way so that inactivity check might still go on in case of tcp write buf full etc.


tssurya:
/close

this PR seems stale? closing this as part of clean up to now trigger CI and skew presubmit results. Please open a new PR if this still needs to be fixed. I don't see any bug associated here.

PR #1847, opened 2023-09-01
OCPBUGS-17773: Perf increases to pod deletion

trozet:
Includes critical performance fixes for 4.14.

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller_namespace.go
+++ b/go-controller/pkg/ovn/base_network_controller_namespace.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"sync"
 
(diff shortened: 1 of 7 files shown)


jcaamano:
@trozet this might increase the probability for the duplicated IPs race condition to happen. The pod is updated with status IP later than the annotation, increasing the time window we might not notice another pod has been allocated with that IP.

trozet:
local gateway jobs are broken for a different reason. Overriding..

/override ci/prow/e2e-aws-ovn-local-to-shared-gateway-mode-migration
/override ci/prow/e2e-aws-ovn-upgrade-local-gateway

PR #1855, opened 2023-09-06
OCPBUGS-18585: Fix OVN SNATing on GR by enabling gateway_mtu on rtoe port of GR

tssurya:
CLEAN cherry pick
/assign @trozet


tssurya:
/retest
@jluhrsen : Where are we with the migration job fixes? Keep this red just to fix the timer bug is holding up signal on those lanes and so other regressions could slip in.
I don't like it, but in lieu of https://issues.redhat.com/browse/OCPBUGS-17391 we need to override these two lanes.

anuragthehatter:
upgrade and regression looks good on cluster bot build

/label qe-approved

PR #1860, opened 2023-09-07
[release-4.13] Serve NodePort services on secondary IP addresses

zeeke:
Backport of:
- https://github.com/ovn-org/ovn-kubernetes/pull/3328
- https://github.com/ovn-org/ovn-kubernetes/pull/3557


abhat:
@zeeke is this driven by some config at the user level? My worry is if we enable this by default, are we making the platform somewhat less secure?

zeeke:
> @zeeke is this driven by some config at the user level? My worry is if we enable this by default, are we making the platform somewhat less secure?

With this feature, the security attack surface surely increases. This is the first time we have approached this problem from the security point of view, as the initial request was to have the same feature as OpenshiftSDN, where kube-proxy accepts connections on every IP addresses [1]

```
--nodeport-addresses strings
A list of CIDR ranges that contain valid node IPs. If set, connections to NodePort services will only be accepted on node IPs in one of the indicated ranges. If unset, NodePort connections will be accepted on all local IPs. This parameter is ignored if a config file is specified by --config.
```

[1] https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/

tssurya:
/close
lack of cycles for SDN team to review this. This is a huge change and we need business justifications for backporting bugs.

PR #1874, opened 2023-09-14
[release-4.14] OCPBUGS-18977: Update bridge flow cache when the host address changes

flavio-fernandes:
When the addressManager detects changes in the node's address, it will update the node address annotations, but it was not reflecting the address changes in the OpenFlow rules within the node via openflowManager.updateBridgeFlowCache(). This could result in nodes having stale rules that affect functionalities such as hairpin.


trozet (approved):
/lgtm

PR #1887, opened 2023-09-18
OCPBUGS-19013: Dockerfile: Copy ovnkube-trace file for RHEL8 platform

pperiyasamy:
The `ovnkube-trace` binary file built for RHEL9 doesn't work anymore on RHEL8 system due to different shared library dependencies between the two OS versions. So this commit copies ovnkube-trace file built for RHEL8 into `/usr/lib/rhel8` directory so that user can download and run it on RHEL-8 platforms.


dcbw:
/retest-required

`{  failed to create or restart e2e-aws-ovn-gather-core-dump pod: unable to create pod: Internal error occurred: resource quota evaluation timed out}`

PR #1925, opened 2023-10-06
OCPBUGS-20274: dont quit if node does not have subnet annotation

jluhrsen:
in some cases, multiple nodes will have to sync and we may not have enough subnets available for all of them. In that case we don't want to quit the sync on other nodes if one is found to not have the subnet annotation as the others may still have it.

Also moves log message for duration of different syncs to the end of their sync so it will only print once.

JIRA: https://issues.redhat.com/browse/OCPBUGS-18598

**- How to verify it**


tssurya:
@jluhrsen : the payloads here are green, thanks for double confirming this, so that means something in CI is what is causing hiccups and not us.

jluhrsen:
this came in with the d/s merge

jluhrsen:
@dcbw , this commit was in the original d/s merge you had, but it was reverted. I think you are working on a new
d/s merge that would contain this. If you think that's gonna take some time, can we get this in by itself? there is
a [permafailing job](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-clusternetwork-cidr-expansion) it will fix.

jluhrsen:
i was confused. this actually already made it in. not sure why the job is still permafailing. need to investigate

PR #1927, opened 2023-10-06
[release-4.14] Use a private IPv4 address range for the transit switch subnet

ricky-rav:
The IP range 168.254.0.0/16 that we chose as default for the transit switch is a public one. Let's use a private one instead: I arbitrarily chose 100.88.0.0/16 from the 100.64.0.0/10 shared address space (https://datatracker.ietf.org/doc/html/rfc6598), as it didn't seem to be in use in the whole openshift or kubernetes orgs: https://github.com/search?q=org%3Aopenshift%20100.88&type=code https://github.com/search?q=org%3Akubernetes+100.88&type=code


ricky-rav:
closing in favour of https://github.com/openshift/ovn-kubernetes/pull/1933

PR #1928, opened 2023-10-06
[WIP] Use a private IPv4 address range for the transit switch subnet

ricky-rav:
The IP range 168.254.0.0/16 that we chose as default for the transit switch is a public one. Let's use a private one instead: I arbitrarily chose 100.88.0.0/16 from the 100.64.0.0/10 shared address space (https://datatracker.ietf.org/doc/html/rfc6598), as it didn't seem to be in use in the whole openshift or kubernetes orgs: https://github.com/search?q=org%3Aopenshift%20100.88&type=code https://github.com/search?q=org%3Akubernetes+100.88&type=code


ricky-rav:
We'll soon do a downstream merge that will include the commit used in this PR. 
Removing the bug label from this PR, as it served for testing upgrades.

ricky-rav:
Closing this as it was merged along with other commits in https://github.com/openshift/ovn-kubernetes/pull/1931

PR #1929, opened 2023-10-09
[release-4.12] OCPBUGS-19975: Update OVN Encap IP when node primary IP address changes

pperiyasamy:
When a node joined to the cluster, kubelet temporarily picked up secondary interface IP address as node's primary IP address into node.Status.Addresses (InternalIP) which made ovnkube-node to use it for OVN Encap IP address for geneve tunnel though kubelet changed to use primary interface IP address a bit later.

Hence this commit watches for node update event in address manager and when there is an update in node's primary ip address, then it would update ovn encap ip address via ovs command.

Manual backport of portion of changes from 4ce2fede6ede6c64fb1554424cf844366e35804c, 4622d618699ad5c16db28e3e099acf22a16caf7c and upstream unmerged change [fd9dac373add8fba54fa4d5955d2a411d6b40099](https://github.com/ovn-org/ovn-kubernetes/pull/3754).

### DIFF
--- a/go-controller/pkg/node/gateway_init_linux_test.go
+++ b/go-controller/pkg/node/gateway_init_linux_test.go
@@ -131,6 +131,17 @@ func shareGatewayInterfaceTest(app *cli.App, testNS ns.NetNS,
 			Cmd:    "ovs-vsctl --timeout=15 get interface eth0 ofport",
 			Output: "7",
 		})
+		if setNodeIP {
+			fexec.AddFakeCmdsNoOutputNoError([]string{
+				"ovs-vsctl --timeout=15 get Open_vSwitch . external_ids:ovn-encap-ip",
+			})
+			fexec.AddFakeCmdsNoOutputNoError([]string{
+				"ovs-vsctl --timeout=15 set Open_vSwitch . external_ids:ovn-encap-ip=192.168.1.10",
+			})
+			fexec.AddFakeCmdsNoOutputNoError([]string{
+				"ovn-appctl --timeout=5 -t ovn-controller exit --restart",
+			})
+		}
 		fexec.AddFakeCmd(&ovntest.ExpectedCmd{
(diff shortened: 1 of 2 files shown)


sdodson:
/hold
@pperiyasamy You cite an unmerged upstream change, how does that align with the fact that you say that 4.13 is already fixed. Is that unmerged upstream change already present in 4.13?

pperiyasamy:
> You cite an unmerged upstream change, how does that align with the fact that you say that 4.13 is already fixed. Is that unmerged upstream change already present in 4.13?

@sdodson The 4.13 commit https://github.com/openshift/ovn-kubernetes/commit/4ce2fede6ede6c64fb1554424cf844366e35804c is actually fixing the original problem, the remaining two commits are just an improvement change which doesn't unnecessarily restart ovn-controller process when there is no changes to node primary ip address.

sdodson:
@pperiyasamy But fd9dac373add8fba54fa4d5955d2a411d6b40099 is going to be a commit that skips 4.14 and 4.13 going straight into 4.12, we shouldn't have commits in 4.12 that aren't present in all upstream branches at the time it merges into 4.12.

pperiyasamy:
> @pperiyasamy But [fd9dac3](https://github.com/openshift/ovn-kubernetes/commit/fd9dac373add8fba54fa4d5955d2a411d6b40099) is going to be a commit that skips 4.14 and 4.13 going straight into 4.12? If so, we shouldn't have commits in 4.12 that aren't present in all upstream branches at the time it merges into 4.12.

@sdodson now this change is merged into master (4.15) addressing the bug [OCPBUGS-15538](https://issues.redhat.com/browse/OCPBUGS-15538). are you okay to unhold this PR once the bug is verified ?

tssurya:
/close

as per https://issues.redhat.com/browse/OCPBUGS-19975?focusedId=24016729&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24016729

PR #1930, opened 2023-10-09
OCPBUGS-20238: Fix webhook error & exit handling

jcaamano:
Fixed two isses that prevented the process to exit on error.

* We were waiting for idleWebhookConnectionsClosed even if l.Listen failed
* We would not exit if either runWebhook or runCSRApproverManager failed, only when both of them failed

Observed when the webhook cannot bind because the port was already in use, which may happen if an old instance of the webhook is running under a different user id.

Cherry-picked from https://github.com/ovn-org/ovn-kubernetes/pull/3951


jcaamano:
/retitle OCPBUGS-20238: Fix webhook error & exit handling

PR #1940, opened 2023-10-18
OCPBUGS-14799: Updating endpoints when using allocateLBNP=0, lgw mode fails

tssurya:
For unknown reasons, during the kube rebase, we just decided to flip from sorting the list of endpoints to not sorting it? : 3850450#diff-d3aa58d9b58a0a09264f072df46ab01d0501eb508c4656411ae2dc1ac68fb3c4L579 The allocateLBNodePorts feature depends on the list being sorted, else our logic is skewed and the random probability model we use to generate iptables gets skewed up as well. So during endpoints update and such we will behave badly and have left over DNATs towards pods that don't exist anymore which will lead to black-holing the traffic.

(cherry picked from commit d9d618783f373ac98d5203e17fe06915222fb926)

**- How to verify it**


trozet (commented):
/lgtm
/label backport-risk-assessed

PR #1947, opened 2023-10-27
Fix Sync netpol ACL filter

npinaeva:
Fix syncNetworkPolicy ACLs filter to not include multicast ACLs, since
multicast ACLs have similar ExternalIDs, but different priority.

Was introduced in 4.13 https://github.com/openshift/ovn-kubernetes/pull/1603 and not backported.
Fixed in 4.14 since https://github.com/ovn-org/ovn-kubernetes/commit/2aab19eef998d10addddfaca80d1c98ee4e3e457#diff-cc83e19af1c257d5a09b711d5977d8f8c20beb34b7b5d3eb37b2f2c53ded1bf7R293-R295

### DIFF
--- a/go-controller/pkg/ovn/policy.go
+++ b/go-controller/pkg/ovn/policy.go
@@ -327,7 +327,8 @@ func (oc *DefaultNetworkController) syncNetworkPolicies(networkPolicies []interf
 
 	// cleanup port groups based on acl search
 	p := func(item *nbdb.ACL) bool {
-		return item.ExternalIDs[policyACLExtIdKey] != "" || item.ExternalIDs[defaultDenyPolicyTypeACLExtIdKey] != ""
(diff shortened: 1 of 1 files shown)


npinaeva:
The intent was to fix network policy port groups being wrongly deleted because of interpreting multicast ACLs as netpol ACLs. Even though multicast ACL can actually give [here](https://github.com/openshift/ovn-kubernetes/blob/release-4.13/go-controller/pkg/ovn/policy.go#L351) either a real namespace name or "clusterRtrPortGroup", extra check https://github.com/openshift/ovn-kubernetes/blob/release-4.13/go-controller/pkg/ovn/policy.go#L352 won't let it delete network policy that actually exists. Also deleting a port group that doesn't exist also won't return an error https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/libovsdb/ops/portgroup.go#L263
So closing, since it doesn't cause a real problem

PR #1953, opened 2023-11-13
[wip] Strip managed fields to save memory

dcbw:
It's a lot of data we don't care about.


dcbw:
Moved upstream as https://github.com/ovn-org/ovn-kubernetes/pull/4012

PR #1757, opened 2023-07-14
SDN-3733: [DownstreamMerge] 7-13-23

trozet:
@npinaeva @tssurya @jcaamano @maiqueb

PR #1961, opened 2023-11-28
[release-4.12] OCPBUGS-18938: Wait for other SBDB objects

flavio-fernandes:
[Backporting from release 4.13](https://github.com/openshift/ovn-kubernetes/pull/1757/commits/17401b887d636c2784c098dcdfc05c767503366b)

There are other objects that we update in SBDB that depend on an object being created by northd and may race. Failures here can further delay node add by 30 seconds.

Ideally we should lower the retry mechanism interval from 30 seconds to something much smaller (now that we use a per key locked syncmap), however that could have even greater implications. So for now, just add in path delays for SBDB objects.

### DIFF
--- a/go-controller/pkg/util/ovn.go
+++ b/go-controller/pkg/util/ovn.go
@@ -6,8 +6,11 @@ package util
 import (
 	"fmt"
 	"net"
+	"time"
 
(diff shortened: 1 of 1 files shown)


flavio-fernandes:
Jira closed. Closing this PR.

PR #1973, opened 2023-12-05
OCPBUGS-24465: Removes Interconnect references from e2e tests

trozet:
IC doesn't exist in 4.13 and these functions were accidentally backported to 4.13.


trozet:
The release repo script pulls the respective branch, instead of using the code from the PR. I'll need to fix this in the release repo, and also enable this job as required in ovnk repo.

trozet:
oh wait this is only passing now because @jcaamano fixed the same thing here https://github.com/openshift/ovn-kubernetes/commit/27a0f9f5b4bc70e9e52ad3721fbbfc5b561b9bf8

I think the bug still exists in CI jobs.

PR #1977, opened 2023-12-08
[release-4.12] OCPBUGS-23981: Netpol retryFramework cleanup

npinaeva:
Kind of a backport of https://github.com/openshift/ovn-kubernetes/pull/1957
But with a full re-implementation because of 4.12 vs 4.13 differences

### DIFF
--- a/go-controller/pkg/retry/obj_retry.go
+++ b/go-controller/pkg/retry/obj_retry.go
@@ -78,6 +78,9 @@ type RetryFramework struct {
 	watchFactory      *factory.WatchFactory
 	ResourceHandler   *ResourceHandler
 	terminatedObjects sync.Map
+
+	// stopChan may be used to stop RetryFramework before watchFactory is shut down
+	stopChan <-chan struct{}
 }
 
 // NewRetryFramework returns a new RetryFramework instance, essential for the whole retry logic.
@@ -88,12 +91,22 @@ type RetryFramework struct {
 func NewRetryFramework(
 	watchFactory *factory.WatchFactory,
 	resourceHandler *ResourceHandler) *RetryFramework {
+	return NewRetryFrameworkWithStopChan(watchFactory, resourceHandler, nil)
+}
+
+// NewRetryFrameworkWithStopChan returns a new RetryFramework instance that may be stopped by closing a given stopChan.
+// It is also stopped on watchFactory shutdown.
+func NewRetryFrameworkWithStopChan(
+	watchFactory *factory.WatchFactory,
+	resourceHandler *ResourceHandler,
+	stopChan <-chan struct{}) *RetryFramework {
 	return &RetryFramework{
 		retryEntries:      syncmap.NewSyncMap[*retryObjEntry](),
 		retryChan:         make(chan struct{}, 1),
 		watchFactory:      watchFactory,
 		ResourceHandler:   resourceHandler,
 		terminatedObjects: sync.Map{},
+		stopChan:          stopChan,
 	}
 }
 
@@ -372,7 +385,7 @@ func (r *RetryFramework) periodicallyRetryResources() {
 	timer := time.NewTicker(RetryObjInterval)
 	waitCh := make(chan struct{})
 	go func() {
-		r.watchFactory.WaitForWatchFactoryStopChannel(waitCh)
+		r.watchFactory.WaitForWatchFactoryStopChannel(waitCh, r.stopChan)
 	}()
 	defer timer.Stop()
 	for {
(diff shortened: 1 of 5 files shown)


go-controller/pkg/retry/obj_retry.go
```diff
@@ -78,6 +78,9 @@ type RetryFramework struct {
 	watchFactory      *factory.WatchFactory
 	ResourceHandler   *ResourceHandler
 	terminatedObjects sync.Map
+
+	// extraStopChan may be used to stop RetryFramework before watchFactory is shut down
+	extraStopChan <-chan struct{}
```
jcaamano, line 83: `extraStopChan` sounds a bit weird given that there is no other stop chan, you could have just used `stopChan`

The fact that the `RetryFramework` has to stop the the `watchFactory` stops is kind of a given, so just `stopChan` sounds like a natural thing to have.

go-controller/pkg/retry/obj_retry.go
```diff
@@ -385,6 +396,9 @@ func (r *RetryFramework) periodicallyRetryResources() {
 			r.iterateRetryResources()
 			timer.Reset(RetryObjInterval)
 
+		case <-r.extraStopChan:
+			klog.V(5).Infof("Extra stop channel got triggered: will stop retrying failed objects of type %s", r.ResourceHandler.ObjType)
+			return
```
jcaamano, line 401: I think that this might be racey, we might notice `waitCh` closed before we notice `extraStopChan` closed.

A potentially more usable approach would be for the `watchFactory` to provide a context that is cancelled when the stopped. Then here you can check if that or either extraStopChan is closed.

go-controller/pkg/retry/obj_retry.go
npinaeva: I thought that stopChan as a name may give the wrong idea that it is required to stop RetryFramework (based on how we usually use this name in out code), but if you don't think it is confusing, I can rename it

go-controller/pkg/retry/obj_retry.go
npinaeva: I was trying to make the least number of changes possible, so sticking to stop channels. (up to switching to contexts if we find out stop channel don't work here)
Can you please elaborate on the race problem? not sure I understand the scenario and the problem it causes

go-controller/pkg/retry/obj_retry.go
jcaamano: It did confuse me for a sec and went looking for the other stop channel lol. 

The idea I get from from a `stopChan` it is that it is a way to stop it but not necessarily the only one. If there is more ways it could stop, I think that the user is the one that should be aware of it so that information should be exported in the doc of the retry framework and not really hidden in a private field name/comment. 

Anyway, just a nit. If you are happy with `extraStopChan` after my impression we can just let it be.

go-controller/pkg/retry/obj_retry.go
jcaamano: So you have three threads: the one that closes `extraStopChan`, the one that closes `waitCh` and this one here. This one here could be paused while the former thread do their thing resulting in both `extraStopChan` and `waitCh` being closed. Then when this thread awakes with both channels closed, are you sure which select branch is going to trigger? There might be an order to it but I am not aware.

npinaeva: I think the order should be random same as receiving messages from multiple channels, but the order doesn't change anything except the message that is logged?

jcaamano: Right but why would we add a log than can't be trusted. Perhaps remove then?

npinaeva: ah I see, I think we can just update the `waitCh` message to say that either watch factory was shut down or stop channel was triggered, then it will be true?

jcaamano: I just don't see the usefulness of having two distinct logs here when we don't know which might be triggered anyway for the same event. I would be happy to just have the single log that was already there, I think it could be easily correlated with other logs to know what did happen. Or perhaps you could add a log in a different place that could tell us more, like in `WaitForWatchFactoryStopChannel`.

npinaeva: Sure, I don't think distinctive messages are really useful, so what about the new version?

jcaamano: I think I might be missing something. Why do you want to have both cases
```
case <-r.stopChan:
			return
case <-waitCh:
			return
```
if the closure of `stopChan` will also cause the closure of `waitCh`? You could just leave what it was there before
```
		case <-waitCh:
			klog.V(5).Infof("Stop channel got triggered: will stop retrying failed objects of type %s", r.ResourceHandler.ObjType)
```

npinaeva: context switching side-effect :'(

PR #1981, opened 2023-12-11
OCPBUGS-25081: Update ACL syncer: make default deny acls filter more strict,

tssurya:
delete all acls that were not synced since they should be stale. Add unit test for a specific multicast leftover acl that was a result of libovsdb migration.

Signed-off-by: Nadia Pinaeva <npinaeva@redhat.com>
(cherry picked from commit 469e752fc8adc105fc23b061d9405b91e559b496)

**- How to verify it**

### DIFF
--- a/go-controller/pkg/ovn/external_ids_syncer/acl/acl_sync.go
+++ b/go-controller/pkg/ovn/external_ids_syncer/acl/acl_sync.go
@@ -136,6 +136,23 @@ func (syncer *aclSyncer) SyncACLs(existingNodes *v1.NodeList) error {
 		if err != nil {
 			return fmt.Errorf("cannot update stale ACLs: %v", err)
 		}
+
+		// There may be very old acls that are not selected by any of the syncers, delete them.
+		// One example is stale multicast ACLs with the old priority that was accidentally changed by
+		// https://github.com/ovn-org/ovn-kubernetes/commit/f68d302664e64093c867c0b9efe08d1d757d6780#diff-cc83e19af1c257d5a09b711d5977d8f8c20beb34b7b5d3eb37b2f2c53ded1bf7L537-R462
+		leftoverACLs, err := libovsdbops.FindACLsWithPredicate(syncer.nbClient, legacyAclPred)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


go-controller/pkg/ovn/external_ids_syncer/acl/acl_sync.go
```diff
@@ -402,8 +419,8 @@ func (syncer *aclSyncer) updateStaleDefaultDenyNetpolACLs(legacyACLs []*nbdb.ACL
 		// sync default Deny policies
 		// defaultDenyPolicyTypeACLExtIdKey ExternalID was used by default deny and multicast acls,
 		// but multicast acls have specific DefaultMcast priority, filter them out.
-		if acl.ExternalIDs[defaultDenyPolicyTypeACLExtIdKey] == "" || acl.Priority == types.DefaultMcastDenyPriority ||
-			acl.Priority == types.DefaultMcastAllowPriority {
+		if acl.ExternalIDs[defaultDenyPolicyTypeACLExtIdKey] == "" || (acl.Priority != types.DefaultAllowPriority &&
+			acl.Priority != types.DefaultDenyPriority) {
 			// not default deny policy
```
jcaamano, line 424: I don't understand this. Reads tome like now we are filtering out default deny ACLs instead of multicast ACLs. It is also strange that this was not bundled with an update in the comment above.

tssurya: cc @npinaeva should respond to this?
I am merely backporting a fix she did. This fix has no conflicts. I don't want to overstep here.

npinaeva: this condition tries to filter out "not default deny policy", the easiest way to do that is filter everything that has `not default deny policy priority`. The previous condition implies that the only non-default-deny priority is multicast, which is true in the current version, but in the previous versions (which may leave stale acls) `DefaultRoutedMcastAllowPriority` was also used. So instead of matching on all non-default-deny-priorities we match on `!= default-deny-priority`
The comment was just missed, I will try to replace "specific DefaultMcast" to "different" upstream

PR #1988, opened 2023-12-12
OCPBUGS-25224: [release-4.13] fixes MTU configuration on gateway router

trozet:
Trivial conflicts with references to things that pertain to IC or code comments that dont exist in 4.13 code. Only significant conflict was missing the GetIPNetFullMask function which I had to make slightly different from upstream due to function dependencies.


trozet:
@weliang1 PTAL

PR #2002, opened 2023-12-29
Support permanent session affinity

tssurya:
OVN drops the first packet from a given client
for every new session which is undesirable and
a regression compared to how SDN behaves (it can
be argued that this is an implementation detail
but ultimately users want same behaviour and no
drops). This is a complicated fix and will take
time.
The OVN fix will take time and in K8s there is no
affinity without a timeout unfortunately. So in
OVNKube we will introduce permanent session affinity as an alternative. Hence if user sets 86000 which is the highest timeout value it would mean no timeout. This is not how the K8s definition works, this will be an ovnkube implementation detail.
There is no harm in changing what a 1 day timeout means from it being a day to it being infinite affinity because either ways OVN only supports upto 18hours of timeout value max (UINT_MAX). So currently range 18-24 is not used and is configured as 18 itself in OVN.
So from this change forward, if timeout is set to 86000 by the user, they will get permanent session affinity.

(cherry picked from commit 357f63ac91fc44b7873e910b8a9c7837f7063e17)

**- How to verify it**
A service with affinity timeout set to 86000:
```
_uuid               : 820724ad-1d5d-4575-a8d6-0bf4a04c791e
external_ids        : {"k8s.ovn.org/kind"=Service, "k8s.ovn.org/owner"="foo/hello-world-2"}
health_check        : []
ip_port_mappings    : {}
name                : "Service_foo/hello-world-2_TCP_cluster"
options             : {event="false", hairpin_snat_ip="169.254.169.5 fd69::5", neighbor_responder=none, reject="true", skip_snat="false"}
protocol            : tcp
selection_fields    : [ip_dst, ip_src]
vips                : {"10.96.253.40:80"="10.244.0.6:8080,10.244.1.3:8080,10.244.2.3:8080"}
```
A service with timeout set to any other number:
```
_uuid               : 6094234e-251d-41ca-88b5-95b0984b25d4
external_ids        : {"k8s.ovn.org/kind"=Service, "k8s.ovn.org/owner"="foo1/hello-world-2"}
health_check        : []
ip_port_mappings    : {}
name                : "Service_foo1/hello-world-2_TCP_cluster"
options             : {affinity_timeout="10800", event="false", hairpin_snat_ip="169.254.169.5 fd69::5", neighbor_responder=none, reject="true", skip_snat="false"}
protocol            : tcp
selection_fields    : []
vips                : {"10.96.128.49:80"="10.244.0.7:8080,10.244.1.4:8080,10.244.2.4:8080"}
```

### DIFF
--- a/go-controller/pkg/ovn/controller/services/lb_config.go
+++ b/go-controller/pkg/ovn/controller/services/lb_config.go
@@ -768,7 +768,7 @@ func getSessionAffinityTimeOut(service *v1.Service) int32 {
 
 func hasSessionAffinityTimeOut(service *v1.Service) bool {
 	return service.Spec.SessionAffinity == v1.ServiceAffinityClientIP &&
-		getSessionAffinityTimeOut(service) > 0
+		getSessionAffinityTimeOut(service) != 86000
 }
 
 // lbOpts generates the OVN load balancer options from the kubernetes Service.
--- a/go-controller/pkg/ovn/controller/services/loadbalancer.go
+++ b/go-controller/pkg/ovn/controller/services/loadbalancer.go
@@ -362,8 +362,16 @@ func buildLB(lb *LB) *templateLoadBalancer {
 	// Session affinity
 	// If enabled, then bucket flows by 3-tuple (proto, srcip, dstip) for the specific timeout value
 	// otherwise, use default ovn value
+	selectionFields := []nbdb.LoadBalancerSelectionFields{}
 	if lb.Opts.AffinityTimeOut > 0 {
-		options["affinity_timeout"] = fmt.Sprintf("%d", lb.Opts.AffinityTimeOut)
+		if lb.Opts.AffinityTimeOut != 86000 {
+			options["affinity_timeout"] = fmt.Sprintf("%d", lb.Opts.AffinityTimeOut)
+		} else {
+			selectionFields = []string{
+				nbdb.LoadBalancerSelectionFieldsIPSrc,
+				nbdb.LoadBalancerSelectionFieldsIPDst,
+			}
+		}
 	}
 
 	if lb.Opts.Template {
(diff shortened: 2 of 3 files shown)


go-controller/pkg/ovn/controller/services/lb_config.go
```diff
@@ -768,7 +768,7 @@ func getSessionAffinityTimeOut(service *v1.Service) int32 {
 
 func hasSessionAffinityTimeOut(service *v1.Service) bool {
 	return service.Spec.SessionAffinity == v1.ServiceAffinityClientIP &&
-		getSessionAffinityTimeOut(service) > 0
+		getSessionAffinityTimeOut(service) != 86000
```
dcbw, line 771: @tssurya you could use `core.MaxClientIPServiceAffinitySeconds` here from `k8s.io/kubernetes/pkg/apis/core`

go-controller/pkg/ovn/controller/services/loadbalancer.go
```diff
@@ -362,8 +362,16 @@ func buildLB(lb *LB) *templateLoadBalancer {
 	// Session affinity
 	// If enabled, then bucket flows by 3-tuple (proto, srcip, dstip) for the specific timeout value
 	// otherwise, use default ovn value
+	selectionFields := []nbdb.LoadBalancerSelectionFields{}
 	if lb.Opts.AffinityTimeOut > 0 {
-		options["affinity_timeout"] = fmt.Sprintf("%d", lb.Opts.AffinityTimeOut)
+		if lb.Opts.AffinityTimeOut != 86000 {
```
dcbw, line 367: same here for `core.MaxClientIPServiceAffinitySeconds`

go-controller/pkg/ovn/controller/services/lb_config.go
tssurya: yep thanks @dcbw : will update it here: https://github.com/ovn-org/ovn-kubernetes/pull/4066/files

PR #2041, opened 2024-01-31
Add .snyk file with ignore rules

pperiyasamy:


tssurya:
what? why isn't security ci not here... that is strange...

PR #2076, opened 2024-02-19
[release-4.14] OCPBUGS-29342: AdminPolicyBasedExternalRoute CRD failing to watch and reconcile routes for later pods

jordigilh:
Fix backported from 4.15 and merged in https://github.com/ovn-org/ovn-kubernetes/pull/3934

The issue is that pods that don't have an `IP` in their status when being processed by the APB controller fail to trigger a new reconciliation loop, thus missing the logic that adds their configuration to the north bound DB.

@npinaeva @jcaamano PTAL.

@josecastillolema FYI.


trozet (approved):
/label backport-risk-assessed
/lgtm

PR #2079, opened 2024-02-20
Reuse node-subnet from cache if exists

jluhrsen:
This PR is to return an already allocated subnet instead of creating a new one if a combination of subnet and node name already exists in the cache.

This PR is to stop multiple allocation of node-subnet from ClusterNetwork CIDR for any node if addition of
'k8s.ovn.org/node-subnets' annotation gets delayed but does not reach default BackOff[1] time limit for retry framework and meanwhile syncNodeNetworkAnnotations[2] gets executed multiple times.

JIRA: https://issues.redhat.com/browse/OCPBUGS-25733
[1] - https://github.com/kubernetes/client-go/blob/v0.26.1/util/retry/util.go#L38
[2] - https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/clustermanager/node/node_allocator.go#L178

**- How to verify it**

### DIFF
--- a/go-controller/pkg/clustermanager/node/subnet_allocator.go
+++ b/go-controller/pkg/clustermanager/node/subnet_allocator.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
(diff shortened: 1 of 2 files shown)


jluhrsen:
this does not help the trouble with the cidr-expansion job perma-failing on 4.14

PR #2090, opened 2024-03-01
OCPBUGS-29851: Egress IP, Services: use all node IP addresses

pperiyasamy:
Manual cherry-pick of 4.14 commit 5c6edd6271697912d2fd732fc796a60a88981954.

Resolved conflicts:

```
go-controller/pkg/ovn/egressservices_test.go
go-controller/pkg/ovn/ovn.go
go-controller/pkg/util/node_annotations.go
go-controller/pkg/util/util.go
```

Match on all node's IP addresses on the no re-route policies so that intra cluster traffic is not routed through the egress or service IP even when the destination IP address is a secondary node IP.

Secondary node IPs might be VIPs that move across nodes.

(cherry picked from commit 5c6edd6271697912d2fd732fc796a60a88981954)

### DIFF
--- a/go-controller/pkg/util/node_annotations.go
+++ b/go-controller/pkg/util/node_annotations.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 
 	kapi "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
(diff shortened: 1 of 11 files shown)


go-controller/pkg/util/node_annotations.go
```diff
@@ -540,6 +541,10 @@ func SetNodeHostAddresses(nodeAnnotator kube.Annotator, addresses sets.Set[strin
 	return nodeAnnotator.Set(ovnNodeHostAddresses, addresses.UnsortedList())
 }
 
+func NodeHostAddressesAnnotationChanged(oldNode, newNode *v1.Node) bool {
+	return oldNode.Annotations[ovnNodeHostAddresses] != newNode.Annotations[ovnNodeHostAddresses]
```
tssurya, line 545: which PR introduced this change in 4.14 and which one wasn't backported to 4.13 due to which this is something you are having to add here?

tssurya (commented):
I spotted the conflicts for the 1st files, I can't figure out the difference in `go-controller/pkg/util/util.go` @pperiyasamy can you please help me here?

tssurya:
/lgtm
rest look ok based on https://github.com/openshift/ovn-kubernetes/commit/5c6edd6271697912d2fd732fc796a60a88981954
I will add risk-assessed after Peri runs a test for this bit on 4.13 cluster.

PR #2105, opened 2024-04-01
[release-4.15] OCPBUGS-30899: use a forked version of j-keck/arping that fixes a threading issue

JacobTanenbaum:
clean cherry-pick of 1aa87b4

here is the PR for the upstream version https://github.com/j-keck/arping/pull/17

using sock.deinitialize() outside of the goroutine that is using the socket leaves the potential for erroneous and unexpected behavior. When calling Ping() very quickly the socket fd can be reused while a goroutine is still trying to read from it and cause the goroutine to deadlock.

removing the sock.deinitialize() from the timeout select case as it will be taken care of by the goroutine upon its return'

**- How to verify it**


JacobTanenbaum:
samplers seem to not deploy properly 

/retest-required

PR #2112, opened 2024-04-05
[4.16][WIP] Fix endpoint selection for externalTrafficPolicy=local and no downstream hacks

ricky-rav:


ricky-rav:
`DNS should answer queries using the local DNS endpoint` is failing as expected, since I removed the downstream hack for it. I'm interested in e2e-metal-ipi-ovn-ipv6.

PR #2118, opened 2024-04-11
OCPBUGS-29397: 4.14 High CPU usage with APB CRD

jordigilh:
Fixes an issue when using APB in a cluster with high number of pods where the APB controller would hit the KAPI for each pod event to update the APB status, causing a high cpu usage.
The fix resolves in checking if the last message in the APB status in the informer matches the same message generated for the pod event, and if different or different status (succeeded or failed) then proceed to request the latest copy of the APB CR from the KAPI server and update it.

### DIFF
--- a/go-controller/pkg/ovn/controller/apbroute/master_controller.go
+++ b/go-controller/pkg/ovn/controller/apbroute/master_controller.go
@@ -516,7 +516,7 @@ func (c *ExternalGatewayMasterController) updateStatusAPBExternalRoute(policyNam
 	}
 	successMessage := fmt.Sprintf("Configured external gateway IPs: %s", strings.Join(sets.List(gwIPs), ","))
 	var lastMessage, failedMessage string
-	if !status.LastTransitionTime.Time.IsZero() {
+	if len(status.Messages)>0 {
 		lastMessage = status.Messages[len(status.Messages)-1]
 	}
 	if processedError != nil {
--- a/go-controller/pkg/ovn/controller/apbroute/master_controller.go
+++ b/go-controller/pkg/ovn/controller/apbroute/master_controller.go
@@ -1,16 +1,19 @@
 package apbroute
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
 	nettypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	libovsdbclient "github.com/ovn-org/libovsdb/client"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/controller/apbroute/master_controller.go
```diff
@@ -496,34 +496,53 @@ func (c *ExternalGatewayMasterController) processNextPodWorkItem(wg *sync.WaitGr
...
-				// policy doesn't exist, no need to update status
-				return nil
+	status, err := c.GetAPBRoutePolicyStatus(policyName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		klog.Warningf("AdminPolicyBasedExternalRoute %s has been deleted before being able to update its status", policyName)
+		return nil
+	}
+	successMessage := fmt.Sprintf("Configured external gateway IPs: %s", strings.Join(sets.List(gwIPs), ","))
+	var lastMessage, failedMessage string
+	if !status.LastTransitionTime.Time.IsZero() {
```
npinaeva, line 519: won't `if len(status.Messages) > 0` be more reliable here?

jordigilh: What makes you think so? Both objects are initialized in the status at the same time (i.e. they are not pointers) so either one will do just fine for this check. What am I missing?

jordigilh: Are you saying this because it might be safer to check the slice before retrieving the last message, regardless of the time value? To avoid a potential error if the slice is empty but the time has been initialized? If so yeah, maybe that would be a safer approach.

jordigilh: Updated to check for the slice length instead of time. I don't think it's a realistic case today since both fields are updated in the same transaction, but better to be safe than sorry 😄

go-controller/pkg/ovn/controller/apbroute/master_controller.go
```diff
@@ -496,34 +496,53 @@ func (c *ExternalGatewayMasterController) processNextPodWorkItem(wg *sync.WaitGr
...
-				// policy doesn't exist, no need to update status
-				return nil
+	status, err := c.GetAPBRoutePolicyStatus(policyName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		klog.Warningf("AdminPolicyBasedExternalRoute %s has been deleted before being able to update its status", policyName)
+		return nil
+	}
+	successMessage := fmt.Sprintf("Configured external gateway IPs: %s", strings.Join(sets.List(gwIPs), ","))
+	var lastMessage, failedMessage string
+	if len(status.Messages)>0 {
```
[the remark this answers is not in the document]
jordigilh: Done

trozet (approved):
/lgtm

trozet:
/retitle OCPBUGS-29397: 4.14 High CPU usage with APB CRD

josecastillolema:
From a perf/scale perspective the fix was successfully validated on the ScaleLab at a 120 node scale.

PR #2161, opened 2024-05-09
OKD-40: Add ovnver_okd and ovsver_okd to Dockerfile.base

aleskandro:

The current process involving a reconciliation of the openshift/release from ART does not allow overwriting the build arguments in the OKD variant Prow Config for ovn-kubernetes. This commit adds two additional build-args with default values for OKD and let the RUN layer decide on whether to use them as ovn and ovs versions to consider for installation.

To simplify (avoid redundant shell code), this PR also proposes to flatten the second RUN layer into the first.

**- How to verify it**

- [OCP] ci/prow/images should succeed => 
- [OKD] ci/prow/okd-scos-images should succeed => https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_ovn-kubernetes/2161/pull-ci-openshift-ovn-kubernetes-master-okd-scos-images/1788571045293199360

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -14,17 +14,19 @@ RUN dnf install -y --nodocs \
 
 ARG ovsver=3.3.0-2.el9fdp
 ARG ovnver=24.03.1-36.el9fdp
+# NOTE: Ensure that the versions of OVS and OVN are overriden for OKD in each of the subsequent layers.
+ARG ovsver_okd=3.3.0-2.el9s
+ARG ovnver_okd=24.03.1-5.el9s
 
 RUN INSTALL_PKGS="iptables nftables" && \
+    source /etc/os-release && \
+    [ "${ID}" == "centos" ] && ovsver=$ovsver_okd && ovnver=$ovnver_okd; \
 	ovsver_short=$(echo "$ovsver" | cut -d'.' -f1,2) && \
 	ovnver_short=$(echo "$ovnver" | cut -d'.' -f1,2) && \
 	dnf install -y --nodocs $INSTALL_PKGS && \
 	dnf install -y --nodocs "openvswitch$ovsver_short = $ovsver" "python3-openvswitch$ovsver_short = $ovsver" && \
 	dnf install -y --nodocs "ovn$ovnver_short = $ovnver" "ovn$ovnver_short-central = $ovnver" "ovn$ovnver_short-host = $ovnver" && \
-	dnf clean all && rm -rf /var/cache/*
-
-RUN ovsver_short=$(echo "$ovsver" | cut -d'.' -f1,2) && \
-	ovnver_short=$(echo "$ovnver" | cut -d'.' -f1,2) && \
+	dnf clean all && rm -rf /var/cache/* && \
 	sed 's/%/"/g' <<<"%openvswitch$ovsver_short-devel = $ovsver% %openvswitch$ovsver_short-ipsec = $ovsver% %ovn$ovnver_short-vtep = $ovnver%" > /more-pkgs
 
 RUN mkdir -p /var/run/openvswitch && \


Prashanth684:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-master-okd-images

martinkennelly:
Added to our team agenda

aleskandro:
Note: `okd-e2e-gcp-ovn` consumes the rhel-based image built by `okd-images`. The changes in this PR affect the result of `okd-scos-images` and there are no tests consuming the `okd-scos-images` (as we still miss a first payload to be composed by the release controller due to the ovn and a few other missing images).

Prashanth684:
> /test okd-e2e-gcp-ovn

@martinkennelly the image we are testing here is the `okd-scos-images`, which is an effort we are going through to build each container image on top of centos ([OKD-210](https://issues.redhat.com/browse/OKD-210)). This okd test uses the old okd container image built on top of rhel and so this is unrelated. Once we build all container images for OKD on top of centos stream, these jobs will use those containers.

LorbusChris:
As noted above, the OKD e2e tests are currently not expected to succeed and can be ignored on this PR for now.

/lgtm
/assign @jacobtanenbaum
for approval

Dockerfile.base
```diff
@@ -14,17 +14,19 @@ RUN dnf install -y --nodocs \
 
 ARG ovsver=3.3.0-2.el9fdp
 ARG ovnver=24.03.1-36.el9fdp
+# NOTE: Ensure that the versions of OVS and OVN are overriden for OKD in each of the subsequent layers.
+ARG ovsver_okd=3.3.0-2.el9s
+ARG ovnver_okd=24.03.1-5.el9s
```
JacobTanenbaum, line 19: I would rather see two different Dockerfiles, one for rhel and another for coreos. @tssurya is that request to pedantic?  
but if one file is fine could you be more clear in what is happening, possibly with a link to  reference the fact that you override the base image when doing okd builds

tssurya: so I am ok with same docker file but I don't understand the https://github.com/openshift/release/blob/master/ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-master__okd-scos.yaml#L24 override part.

I also don't see why okd is pinning different versions here compared to default ocp, I'd rather not have new variables; OKD should just be using the default ARG values otherwise its going to get difficult to maintain and update these.

aleskandro: Hi @JacobTanenbaum @tssurya, 

> so I am ok with same docker file (though traditionally like Jake said rhel is main and centos being just another distro can inherit the base and then build differently, but yeah one dockerfile per distro seems more proper) but I don't understand the https://github.com/openshift/release/blob/master/ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-master__okd-scos.yaml#L24 override part. I agree with Jacob we perhaps need more info here on what's happening

The images for the OKD components are all being rebuilt using CentOS Stream 9 as a base image. The override in the Prow Config is happening as it always did for the RHEL-based images too. That `from:` field will let the image build on top of a CS9 base instead of the _placeholder_ you have in your dockerfile for the runtime stage.

We would avoid having separate Dockerfiles for RHEL-based and CS9-based images, as it increases the maintenance efforts for you to ensure both the Dockerfiles are always successfully building the image and in sync in particular. The process would be error-prone as we could end up with updates to your code AND the RHEL Dockerfile, but not the CS9 one. Finally, the CS9 image could still be built successfully but perhaps missing some system-level configuration. Finally, we get a non-working image for OKD, although everything looks "green". 

So, unless there are specific reasons that we don't see here, we are avoiding doing it for all the Openshift component's Dockerfiles.

> I also don't see why okd is pinning different versions here compared to default ocp, I'd rather not have new variables; OKD should just be using the default ARG values otherwise its going to get difficult to maintain and update these.

The default values you set for the RHEL-based images target RPMs that cannot be installed in CS9 as we can't use the FDP repositories.

I would also prefer not to use an additional build argument, but we are currently using automation to build the OKD configuration, and having them set in the prow config (overriding the current ones) would mean updating this doozer-based automation config (in openshift-eng/ocp-build-data) instead of just doing it in the same PR that will bump the version for RHEL when you'll need to do it (cc @jupierce).

tssurya:
@jupierce and @aleskandro : there is a presubmit for OKD we run in ovnkube which is always red `okd-e2e-gcp-ovn`. Shouldn't that work with this PR?

aleskandro:
Hi @tssurya, this PR is unrelated to that failure. The output artifacts from these PRs will land in the okd-scos stream, which is not the target for the okd-e2e-gcp-ovn pre-submit.

tssurya:
/hold
till https://github.com/openshift/ovn-kubernetes/pull/2171merges (the deadline for 2172 is 8AM UTC which is in a few hours) This is to avoid merge pool churn
I will remove hold on this one asap. I hope that is ok @aleskandro

Prashanth684:
> /hold till https://github.com/openshift/ovn-kubernetes/pull/2171merges (the deadline for 2172 is 8AM UTC which is in a few hours) This is to avoid merge pool churn I will remove hold on this one asap. I hope that is ok @aleskandro

we are hoping that we can get it merged before 4.16 branching :)

tssurya:
> > /hold till https://github.com/openshift/ovn-kubernetes/pull/2171merges (the deadline for 2172 is 8AM UTC which is in a few hours) This is to avoid merge pool churn I will remove hold on this one asap. I hope that is ok @aleskandro
> 
> we are hoping that we can get it merged before 4.16 branching :)

yes already spoke to @jupierce in DM before putting hold, happy to approve the backport to 4.16 worstcase.

PR #2170, opened 2024-05-15
[release-4.14] OCPBUGS-33721: use a forked version of j-keck/arping that fixes a threading issue

JacobTanenbaum:
clearn cherry-pick of https://github.com/openshift/ovn-kubernetes/commit/1aa87b44c9a64370191945b2c3d57f1f7fc3830b

here is the PR for the upstream version https://github.com/j-keck/arping/pull/17

'
using sock.deinitialize() outside of the goroutine that is using the socket leaves the potential for erroneous and unexpected behavior. When calling Ping() very quickly the socket fd can be reused while a goroutine is still trying to read from it and cause the goroutine to deadlock.

removing the sock.deinitialize() from the timeout select case as it will be taken care of by the goroutine upon its return'

**- How to verify it**


JacobTanenbaum:
CI failed to retrieve the image from the registry 

/retest-required

npinaeva:
/hold
till the urgent fix is merged https://github.com/openshift/ovn-kubernetes/pull/2183

trozet (approved):
/lgtm

PR #1697, opened 2023-06-06
SDN-3733,SDN-3838,SDN-3840,OCPBUGS-11180,OCPBUGS-4370,OCPBUGS-1715,OCPBUGS-10592,OCPBUGS-12747,OCPBUGS-4485,OCPBUGS-10841: Downstream Merge 6th June 2023

tssurya:
**- What this PR does and why is it needed**
This PR brings in OVN-Interconnect feature, multiple bug fixes and commits, Egress Service CRD feature, CPU Pinning feature.
Not clean,
Conflicts in merge PR:

Conflicts:
	go-controller/pkg/config/config_test.go
because we already brought down
https://github.com/openshift/ovn-kubernetes/commit/d7e5bb9fc45eed2d01bdb9e9dcd4ea862a921b56
	go-controller/pkg/ovn/gateway_init.go
because we already brought down
https://github.com/openshift/ovn-kubernetes/commit/fe63418d89e376a74b11013c37552decc9a3ffc5

in the previous merge

**- Special notes for reviewers**

1. There is one downstream specific commit: https://github.com/openshift/ovn-kubernetes/pull/1697/commits/dcd500354c3ed7a11a5ef36dbc014eead7a06291 that is required to fix tests with the new changes coming in from egress svc CRD feature
2. I had to cherry-pick https://github.com/openshift/ovn-kubernetes/pull/1697/commits/5d2d4730f3e5c0b7960b6cb4ae340f9bd4bf9d46 out of order to ensure unit tests don't panic
3. I had to cherry-pick https://github.com/openshift/ovn-kubernetes/pull/1697/commits/a416d67fcc059d1657579736a6a78dfda220a0c2 an unmerged PR: https://github.com/ovn-org/ovn-kubernetes/pull/3653 from upstream -> as discussed with @trozet we can remove the conflict next time during the merge
4. NEW EDIT AFTER PREVIOUS REVIEW: Had to cherry-pick https://github.com/ovn-org/ovn-kubernetes/pull/3671 to ensure we don't have 2 LE's running at the same time during upgrades.

**- How to verify it**
All tests have to be green

commit owners: @numansiddique @oribon @zeeke @MikeZappa87 @bpickard22 @dceara @trozet @wizhaoredhat @cathy-zhou @ricky-rav @maiqueb @pperiyasamy @martinkennelly @jcaamano @bn222 @JacobTanenbaum @dcbw @tssurya @kyrtapz @DmytroLinkin

PR #2172, opened 2024-05-16
[release-4.13] OCPBUGS-33730: Improves service iptables efficiency on start up

trozet:
Includes manual backport of ec5685b as first commit. This needs close review. The rest of the commits were pretty clean. A few trivial things that don't need close attention.

Note, @tssurya brought up a valid question to why https://github.com/openshift/ovn-kubernetes/pull/2172/commits/461b9882bdd6dee89a5846e8e227068060849c16 was necessary (CARRY: fix ordering of MCS blocking iptables rules).

Due to the forwarding rules added in 4.14 and later in initExternalBridgeServiceForwardingRules, the downstream tests have been broken:
```
[Fail] Gateway Init Operations Setting up the gateway bridge [It] sets up a local gateway with predetermined interface 
/mnt/crucial/go/src/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/gateway_init_linux_test.go:1207

[Fail] Gateway Init Operations Setting up the gateway bridge [It] sets up a local gateway with predetermined interface and no default route 
/mnt/crucial/go/src/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/gateway_init_linux_test.go:1207
```

This has gone unnoticed in our downstream, because these tests require root and are skipped in CI. @jechen0648 @anuragthehatter we should add some downstream QE coverage for this.

The commit I added fixes the tests and the ordering of the MCS rules. However, it is only required in 4.14 and later. I think it is a trivial change that we can bring into 4.13 and 4.12 to just provide consistency in our downstream only code.

### DIFF
--- a/go-controller/go.mod
+++ b/go-controller/go.mod
@@ -112,6 +112,7 @@ require (
 
 replace (
 	github.com/Microsoft/hcsshim => github.com/Microsoft/hcsshim v0.8.20
+	github.com/coreos/go-iptables => github.com/trozet/go-iptables v0.0.0-20240328221912-077e672b3808
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
 	github.com/vishvananda/netlink => github.com/jcaamano/netlink v1.1.1-0.20220831114501-3a761ed61db6
 	k8s.io/api => k8s.io/api v0.26.0
--- a/go-controller/pkg/node/gateway_init_linux_test.go
+++ b/go-controller/pkg/node/gateway_init_linux_test.go
@@ -330,7 +330,7 @@ func shareGatewayInterfaceTest(app *cli.App, testNS ns.NetNS,
 				"OVN-KUBE-SNAT-MGMTPORT": []string{},
 				"OVN-KUBE-ETP":           []string{},
 				"OVN-KUBE-ITP":           []string{},
-				"OVN-KUBE-EGRESS-SVC":    []string{"-m mark --mark 0x3f0 -m comment --comment Do not SNAT to SVC VIP -j RETURN"},
+				"OVN-KUBE-EGRESS-SVC":    []string{"-m mark --mark 0x3f0 -m comment --comment DoNotSNAT -j RETURN"},
 			},
 			"filter": {},
 			"mangle": {
@@ -1045,7 +1045,7 @@ OFPT_GET_CONFIG_REPLY (xid=0x4): frags=normal miss_send_len=0`,
 				"OVN-KUBE-SNAT-MGMTPORT": []string{},
 				"OVN-KUBE-ETP":           []string{},
 				"OVN-KUBE-ITP":           []string{},
-				"OVN-KUBE-EGRESS-SVC":    []string{"-m mark --mark 0x3f0 -m comment --comment Do not SNAT to SVC VIP -j RETURN"},
+				"OVN-KUBE-EGRESS-SVC":    []string{"-m mark --mark 0x3f0 -m comment --comment DoNotSNAT -j RETURN"},
 			},
 			"filter": {
 				"FORWARD": []string{
--- a/go-controller/pkg/node/gateway_iptables.go
+++ b/go-controller/pkg/node/gateway_iptables.go
@@ -59,10 +59,12 @@ func insertIptRules(rules []nodeipt.Rule) error {
 	return nodeipt.AddRules(rules, false)
 }
 
-// insertIptRulesFiltered adds the provided rules in an insert fashion with a filter for table/chain
+// restoreIptRulesFiltered restores the provided rules in an insert fashion with a filter for table/chain
 // i.e each rule gets added at the first position in the chain
-func insertIptRulesFiltered(rules []nodeipt.Rule, filter map[string]map[string]bool) error {
-	return nodeipt.AddRulesFiltered(rules, filter)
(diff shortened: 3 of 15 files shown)


go-controller/pkg/node/gateway_init_linux_test.go
```diff
@@ -330,7 +330,7 @@ func shareGatewayInterfaceTest(app *cli.App, testNS ns.NetNS,
 				"OVN-KUBE-SNAT-MGMTPORT": []string{},
 				"OVN-KUBE-ETP":           []string{},
 				"OVN-KUBE-ITP":           []string{},
-				"OVN-KUBE-EGRESS-SVC":    []string{"-m mark --mark 0x3f0 -m comment --comment Do not SNAT to SVC VIP -j RETURN"},
+				"OVN-KUBE-EGRESS-SVC":    []string{"-m mark --mark 0x3f0 -m comment --comment DoNotSNAT -j RETURN"},
```
tssurya, line 333: hmm why is this different from https://github.com/openshift/ovn-kubernetes/pull/1697/commits/ec5685bf762c942e37b36e2a7ec19ab926f2836f#diff-baef649410d0c75a89ece69e5a452be4a68d9fb8f8febed16657ff4fc2e6b2a4R361 ? there if I see it correctly we don't have any rule in that chain right?

go-controller/pkg/node/gateway_iptables.go
```diff
@@ -5,19 +5,17 @@ package node
...
+	nodeipt "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/iptables"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
-	"github.com/pkg/errors"
 	kapi "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
+	"net"
```
tssurya, line 18: nit: hmm this should not change.. keep it back up with the do dev pkgs?

go-controller/pkg/node/gateway_iptables.go
```diff
@@ -576,30 +505,30 @@ func egressSVCIPTRulesForEndpoints(svc *kapi.Service, v4Eps, v6Eps []string) []i
...
-		chain: iptableESVCChain,
-		args: []string{
+func egressSVCIPTDefaultReturnRule() nodeipt.Rule {
+	return nodeipt.Rule{
+		Table: "nat",
+		Chain: iptableESVCChain,
+		Args: []string{
 			"-m", "mark", "--mark", string(ovnKubeNodeSNATMark),
-			"-m", "comment", "--comment", "Do not SNAT to SVC VIP",
+			"-m", "comment", "--comment", "DoNotSNAT",
```
tssurya, line 549: oh I get it, its some esvc feature we have not brought in yet so in 4.13 we still expect this DoNotSNAT rule to exist?

go-controller/pkg/node/iptables/iptables.go
```diff
@@ -0,0 +1,81 @@
...
+				r.Table, r.Chain, strings.Join(r.Args, " "), err)
+		}
+	}
+	if addErrors.Error() == "" {
+		addErrors = nil
+	}
+	return addErrors
+}
+
+func DelRules(rules []Rule) error {
```
tssurya, line 140: nit https://github.com/openshift/ovn-kubernetes/pull/1697/commits/ec5685bf762c942e37b36e2a7ec19ab926f2836f missing doc string; IDC just saying that's where this file diverges from original file :D

tssurya (commented):
commit1 with conflicts lgtm

go-controller/go.mod
tssurya, line 1: @trozet did you go go mod tidy and go mod vendor after this cherry-pick? as in whilst cherrypicking module updates sometimes dependencies changes wanna make sure in 4.14 and 4.13 we are ok -> I get that its pinned to your local branch but wanted to double check

go-controller/pkg/node/gateway_iptables.go
trozet: oh, its actually cause the spaces break iptables-restore because it sees it as separate arguments. I noticed in:
https://github.com/openshift/ovn-kubernetes/commit/ec5685bf762c942e37b36e2a7ec19ab926f2836f#diff-a470e68c8afee98bb50622182d2df9d466c3ce459e0ebf9e35066fc4a75864b6R977

it was changed to be generic anyway so I just also brought that in. egress service is in 4.13 code:
https://github.com/openshift/ovn-kubernetes/tree/release-4.13/go-controller/pkg/ovn/controller/egress_services

go-controller/pkg/node/iptables/iptables.go
trozet: oops. I missed that. I'd prefer not to restart all of CI again for this though.

tssurya (commented):
commits look good: https://github.com/openshift/ovn-kubernetes/pull/2172/commits/461b9882bdd6dee89a5846e8e227068060849c16 is same as https://github.com/openshift/ovn-kubernetes/pull/2164/commits/011ab4a5df8b59795944e754fab3179d8ed02fc0 although I don't understand the history behind the ordering and why we never caught it/need it here in 4.13

PR #2178, opened 2024-05-22
[release-4.13] OCPBUGS-24651: Handle old EgressIP status gracefully

pperiyasamy:
When EIP fails over to another node, then ovnkube-master takes a while to update EIP status with newly assigned node, but before that if there is a pod update event occurs, it still uses previously assigned node for programming SNAT entry, This never gets removed while handling EIP update event once EIP is updated with new node name in the status. This commit removes those stale entries while handing EIP update event.


pperiyasamy:
closing it in favor of https://github.com/openshift/ovn-kubernetes/pull/2189.

PR #2189, opened 2024-05-29
[release-4.13] OCPBUGS-33355,OCPBUGS-24651: EgressIP: Validate cache state when reconciling EgressIP pods and namespaces

kyrtapz:
Note: This change is <4.14 specific because of the EgressIP redesign that happened with IC.

When an egressIP moves from nodeA to nodeB the following happens in `reconcileEgressIP`:
1. Allocate the new node and store it in cache
2. Reassign all matching pods to the new node in nbdb
3. Update the EgressIP status with the new node

There is a chance that `reconcileEgressIPPod`/`reconcileEgressIPNamespace` will run at the same time. If it happens to run before step 3 completes and the informers are updated with the new status there it will use the stale status to configure the nbdb. 
To avoid this issue verify that the cache state matches the current egressIP status.

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -424,6 +424,9 @@ func (oc *DefaultNetworkController) reconcileEgressIP(old, new *egressipv1.Egres
 }
 
 func (oc *DefaultNetworkController) reconcileEgressIPNamespace(old, new *v1.Namespace) error {
+	oc.eIPC.egressIPAssignmentMutex.Lock()
+	defer oc.eIPC.egressIPAssignmentMutex.Unlock()
+
 	// Same as for reconcileEgressIP: labels play nicely with empty object, not
 	// nil ones.
 	oldNamespace, newNamespace := &v1.Namespace{}, &v1.Namespace{}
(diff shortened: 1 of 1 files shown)


pperiyasamy:
Discussed about this PR with @kyrtapz, This is also fixing other bug [OCPBUGS-24651](https://issues.redhat.com//browse/OCPBUGS-24651) which I'm trying to fix via another [PR](https://github.com/openshift/ovn-kubernetes/pull/2178).

 We decided to go ahead with this PR as its very much only tied to cache state matches whereas other PR has changes on `natPredicates` for deleting SNAT entries, We are uncertain this may cause any other side side effect for EgressIP delete cases though unit tests are successful, so will close the PR https://github.com/openshift/ovn-kubernetes/pull/2178 without merging it.

@kyrtapz may update this PR to reduce the scope of `egressIPAssignmentMutex` locking if needed.

/assign @tssurya

pperiyasamy:
/retitle [release-4.13] OCPBUGS-33355,OCPBUGS-24651: EgressIP: Validate cache state when reconciling EgressIP pods and namespaces

huiran0826:
please be noted that pre-merge testing failed with this pr, details were added in the bug.

go-controller/pkg/ovn/egressip.go
```diff
@@ -3169,3 +3193,28 @@ func ipStringToCloudPrivateIPConfigName(ipString string) (name string) {
 	}
 	return
 }
+
+// getEgressIPAllocations returns a map of all allocations found in the cache
+func (oc *DefaultNetworkController) getEgressIPAllocations() map[string]egressIPNodeStatus {
+	allAllocations := make(map[string]egressIPNodeStatus)
+	for _, eNode := range oc.eIPC.allocator.cache {
```
pperiyasamy, line 3200: it's better to read allocator cache after acquiring allocator lock.

huiran0826:
Retested PR, looks good. 
/label qe-approved

PR #2218, opened 2024-07-03
OCPBUGS-36486: Fix race condition when creating/deleting namespace address set

jcaamano:
Fix race condition:

1. deleteNamespaceLocked, after the 20s delay, checks that the current nsInfo for the namespace is nil
2. ensureNamespaceLockedCommon adds a new nsInfo referencing the existing address set
3. deleteNamespaceLocked destroys the existing address set

(cherry picked from commit b1bb8edf95f14c1afc0454296f1f704225e9c51a)


tssurya (approved):
/lgtm
/label backport-risk-assessed
(touches important piece of central codebase but this one was already verified by scale team on 4,17 and has had soak time of over a month so good to go.)

PR #2223, opened 2024-07-10
SDN-4919: Downstream Merge July 10th

tssurya:
@jcaamano @npinaeva @kyrtapz @qinqon


tssurya:
/label acknowledge-critical-fixes-only

All the pieces going in are on a tech preview gated feature. SDN Team is not observing Shift Week because of feature deadlines.

PR #2240, opened 2024-07-29
OCPBUGS-37541: Fix per-pod MCS/metadata blocking

cgoncalves:
The conversion from former iptables rule to nftables rule failed to include further matching criteria [1][2].

TCP SYN-ACK packets from Pod in response to client-side initiating TCP connections on source ports :22623 and :22624 are being rejected.

[1] Commit 90696a9
[2] https://github.com/openshift/ovn-kubernetes/pull/1946/


danwinship:
yeah, looks right.
/lgtm
/approve

danwinship:
/override ci/prow/4.17-upgrade-from-stable-4.16-e2e-gcp-ovn-rt-upgrade

PR #2268, opened 2024-08-20
SDN-4930: Downstream Merge 20th august

tssurya:
/hold

The previous merge has not landed in a CI or nightly build therefore not in a payload yet because of known infrastructure issues, so I won't merge this till that happens....
But I want to get a read of the openflow manager PR

PR #2269, opened 2024-08-21
Dockerfile.base: Bump OVN to ovn24.09-24.09.0-beta.23.el9fdp.

dceara:
This is a pre-release version of the upcoming OVN v24.09.0 release. It's expected the actual release won't be very different from this version.

Relevant OVN features/performance improvements:
- NATs can now be given an arbitrary match condition and priority. This allows for conditional NATs to be configured.
- Added support to define boundaries (min and max values) for selected ct zones.
- Add support for ACL sampling through the new Sample_Collector and Sample tables.  Sampling is supported for both traffic that creates new connections and for traffic that is part of an existing connection.
- Added a new column in the southbound database "flow_desc" to provide human readable context to flows.
- Added two new experimental logical router port options, "routing-protocol-redirect" and "routing-protocols", that allow redirection of routing protocol traffic received by a router port to a different logical switch port.

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-9-release-golang-1.22-openshift-4.17
+  tag: rhel-9-release-golang-1.22-openshift-4.18
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # The standard name for this image is ovn-kube
 
 # Build RHEL-9 binaries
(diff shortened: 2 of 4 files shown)


tssurya:
/hold
I will remove it when we have gotten a green payload with the previous UDN PR, I am controlling the pace of PRs by leaving ample gap between them to ensure we get green nightlies/ci before merging next set.

This PR will go in after the openflow manager DS Merge https://github.com/openshift/ovn-kubernetes/pull/2268

tssurya:
blocked on https://github.com/ovn-org/ovn-kubernetes/pull/4658#issuecomment-2314884878

tssurya:
/close
fixed the conflict here: https://github.com/openshift/ovn-kubernetes/pull/2284

PR #2282, opened 2024-09-02
[release-4.16] OCPBUGS-38697: Use more exact name match when deleting static routes to HO nodes.

pliurh:
It can help to avoid the unintended match when the '-gr' substring exists in a node name.

This is manual cherrypick of commit 6dd663235dab22377971f2b9e109b7af327fd9fb


tssurya (approved):
looks like clean cherry-pick, no conflicts
code looks same as from 4.17

PR #2315, opened 2024-10-09
[release-4.17] OCPBUGS-42931: Add static route to the hairpin masquerade IPs to pod

pliurh:
When users attach pod to a secondary network and override the default route pod. It will cause the assymetric routing for service haripin traffic.

We add static routes to ensure the traffic to the hairpin masquerade IP always goes to OVN.

(This is a clean cherry-pick of 6cdf7ecdff5ea9017c09ca8b35acc3de944a6eff)

#### Which issue(s) this PR fixes

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```


jluhrsen:
the hypershift-conformance-techpreview job is new and we hope that it can pass, but I think it has some permafailing
tests that we may need to dig in to. I started a slack thread [here](https://redhat-external.slack.com/archives/C01C8502FMM/p1730746026912429).

PR #2367, opened 2024-11-26
SDN-5498: make dnf get through repo troubles faster

jluhrsen:
- connection timeouts are now 5s instead of default 30s
- adding flag to skip_if_unavailable so it will not retry on failure

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,8 +45,8 @@ RUN INSTALL_PKGS=" \
 	ethtool conntrack-tools \
 	openshift-clients \
 	" && \
-	dnf install -y --nodocs $INSTALL_PKGS && \
-	eval "dnf install -y --nodocs $(cat /more-pkgs)" && \
+	dnf --setopt=retries=2 --setopt=timeout=2 install -y --nodocs $INSTALL_PKGS && \
+	eval "dnf --setopt=retries=2 --setopt=timeout=2 install -y --nodocs $(cat /more-pkgs)" && \
 	dnf clean all && rm -rf /var/cache/*
 
 COPY --from=builder /go/src/github.com/openshift/ovn-kubernetes/go-controller/_output/go/bin/ovnkube /usr/bin/
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -8,7 +8,7 @@
 FROM registry.ci.openshift.org/ocp/4.17:base-rhel9
 
 # install selinux-policy first to avoid a race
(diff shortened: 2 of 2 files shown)


jluhrsen:
I think this will be helpful. [this images job](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_ovn-kubernetes/2367/pull-ci-openshift-ovn-kubernetes-master-images/1861203399916130304/artifacts/build-logs/ovn-kubernetes-amd64.log) hit the "Timeout" issue which like we can see in [this job](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_ovn-kubernetes/2364/pull-ci-openshift-ovn-kubernetes-master-images/1861137239195521024/artifacts/build-logs/ovn-kubernetes-amd64.log) was defaulting to 30s per try. with this PR that is down to 5s. the first iteration of those timeouts cost us ~4m compared to ~15. It also happens twice so that will cut the time from ~30m lost to closer to ~8m.

trozet (approved):
/lgtm

jluhrsen:
/hold

something broken. some jobs can't find selinux-policy

Dockerfile
```diff
@@ -32,6 +32,33 @@ USER root
 
 ENV PYTHONDONTWRITEBYTECODE yes
 
+# ignore dnf repositories that are not reachable in openshift CI clusters
```
tssurya, line 35: this commented out block is intentional?

jluhrsen: no. thanks for looking. PR is updated.

jluhrsen:
depending on the base image the dnf repos that are configured are different and possibly some are or are not reachable because of some other config. I did not get that deep, but I noticed that just removing all repos that we have seen to show as unreachable or timing out ended up causing some jobs to not be able to install selinux-policy. It's too messy to worry about that, IMHO, so I think the best answer for now is just to have it timeout faster and only retry a couple of times. That's the current stage of this PR.

tssurya (commented):
/lgtm
I am OK in getting this in as a stop gap, but I feel something must have changed somewhere cause say 2-3 months ago I don't remember hitting this issue, I wonder if we can pin point to what introduced this issue in the first place?

PR #2375, opened 2024-12-03
OCPBUGS-45314: pin libreswan to 4.6-3.el9_0.3

zshi-redhat:
For a crash fix and CVE backports in libreswan

NOTE: This is a temporary fix until the RHEL fix lands because we have time constrants and this will be reverted in the future on all the pinning PRs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ RUN INSTALL_PKGS=" \
 	openssl firewalld-filesystem \
 	libpcap iproute iproute-tc strace \
 	tcpdump iputils \
-	libreswan \
+	libreswan-4.6-3.el9_0.3 \
 	ethtool conntrack-tools \
 	openshift-clients \
 	" && \


huiran0826:
IPSec function and regression testing for this PR looks good, no issue found.

/label qe-approved

tssurya (approved):
/approve
I am approving this based on https://redhat-internal.slack.com/archives/C07S54F9URX/p1731335471886419
assuming CVEs etc are handled already
@zshi-redhat / @pperiyasamy can you please link JIRA EPIC cards to revert these changes in the future? Are we tracking the revert of this when the proper fixes have landed?

tssurya:
for visibility,  RHEL9.2  is using swan 4.9 and RHEL9.4 is using swan 4.12
we are downgrading swan to avoid the crash bug and we are pinning it to a version that has all the CVE fixes backported between these two releases i.e 4.6 and 4.9/4.12 swans
the downgrade is temporary till we get the RHEL fix and we can revert this change.

PR #2398, opened 2024-12-17
[release-4.18] OCPBUGS-37193: fixes overzealous deletion of SNAT in egressIP

JacobTanenbaum:
currently when creating egressIP in a dualstack cluster regardless of if the egressip is ipv4 or ipv6 when disable-snat-multiple-gws is set the code removes both ipv4 and ipv6 snats from the database. This means that the pod will not be able to communicate with the cluster correctly becuase the Gateway Router is missing an SNAT and the traffic will be dropped.

Additionally the testing looks a little different from the others because we do not correctly setup dualstack clusters when running unit tests

## 📑 Description

Fixes #

## ✅ Checks

- [x] My code requires changes to the documentation
- [x] if so, I have updated the documentation as required
- [x] My code requires tests


JacobTanenbaum:
closing in favor of https://github.com/openshift/ovn-kubernetes/pull/2403 which brings in this change as https://github.com/openshift/ovn-kubernetes/pull/2403/commits/dfa9d8783069285181243c7328b900fd52cffea1

PR #2403, opened 2024-12-19
SDN-5297,OCPBUGS-46527: DownStream Merge Sync from 4.19 [12-19-2024]

jluhrsen:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI

PR #2409, opened 2025-01-06
OCPBUGS-47800: Fixes unexpected mp0 route removal during start up

trozet:
Routes via mp0 were being deleted on every ovnkube-node restart: [root@ovn-worker ~]# ip monitor route
Deleted 192.72.3.0/24 dev ovn-k8s-mp0 proto kernel scope link src 192.72.3.2 Deleted broadcast 192.72.3.255 dev ovn-k8s-mp0 table local proto kernel scope link src 192.72.3.2 Deleted local 192.72.3.2 dev ovn-k8s-mp0 table local proto kernel scope host src 192.72.3.2 local 192.72.3.2 dev ovn-k8s-mp0 table local proto kernel scope host src 192.72.3.2 broadcast 192.72.3.255 dev ovn-k8s-mp0 table local proto kernel scope link src 192.72.3.2

This causes traffic outage during upgrade, as well as other unwanted side effects when pod-destined traffic is routed via default gateway route in the host. This is especially disruptive in local gateway mode.

This patch removes the teardown, and then makes the synchronization of addresses and routes more robust, so that we can safely handle changes to MTU or mp0 addresses.

(cherry picked from commit d604d597b0065016ee2ed7c907b4b84bd78fc89d)

### DIFF
--- a/go-controller/pkg/node/management-port_linux.go
+++ b/go-controller/pkg/node/management-port_linux.go
@@ -160,48 +160,27 @@ func tearDownInterfaceIPConfig(link netlink.Link, ipt4, ipt6 util.IPTablesHelper
 	return nil
 }
 
-func tearDownManagementPortConfig(mpcfg *managementPortConfig) error {
-	// for the initial setup we need to start from the clean slate, so flush
-	// all (non-LL) addresses on this link, routes through this link, and
-	// finally any IPtable rules for this link.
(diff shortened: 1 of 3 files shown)


trozet:
commit already merged into 4.18 via: https://github.com/openshift/ovn-kubernetes/pull/2403

PR #2425, opened 2025-01-24
[release-4.17] OCPBUGS-48828: fixes overzealous deletion of SNAT in egressIP

jluhrsen:
currently when creating egressIP in a dualstack cluster regardless of if the egressip is ipv4 or ipv6 when disable-snat-multiple-gws is set the code removes both ipv4 and ipv6 snats from the database. This means that the pod will not be able to communicate with the cluster correctly becuase the Gateway Router is missing an SNAT and the traffic will be dropped.

Additionally the testing looks a little different from the others because we do not correctly setup dualstack clusters when running unit tests

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests


trozet (approved):
/lgtm

looks like just one small conflict in deletePodSNATOps where in 4.17 it does not include match criteria

PR #2467, opened 2025-02-19
[WIP] OCPBUGS-43004: Allow default network pods to reach localnet on the same node

ricky-rav:
Add a flow at prio=101 to allow traffic from the OVN network to the host network to be handled normally by the bridge instead of forwarding it directly to the NIC (which would be done by the flow at prio=100). This allows pods in the default network to reach localnet pods on the same node.

Testing upstream changes https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5041


ricky-rav:
The current changes don't work with local gateway mode because the `in_port` doesn't match.
The traffic goes pod->mp0->breth0 and when it's on breth0 it's already SNAT'ed to the node IP.
Having `in_port=LOCAL` breaks the cluster, as well as removing `in_port` and adding `nw_dst=$nodeIP`.

PR #2478, opened 2025-03-06
[release-4.17] OCPBUGS-52480: Handles unspecified protocol in network policy port

trozet:
Fixes a null pointer exception when network policy port has no protocol. If the protocol is missing in the network policy port definition, it should be assumed to be TCP.

(cherry picked from commit bc7e86e468a30200e5867aaf7904429d3f8948ea)

Clean cherry-pick.


tssurya (approved):
/lgtm

PR #2504, opened 2025-04-02
OCPBUGS-54497: [release-4.18] ovspinning: Enable the feature at runtime

zeeke:
A configuration manager like the `machine-config-operator` is often used to enable the OVS CPU pinning feature. During reconciliation loop, the configurator might remove and recreate a file for a very short time. When it comes to the `/etc/openvswitch/enable_dynamic_cpu_affinity` file, it might disable the CPU pinning permanently and an ovnkube-controller restart is needed.

Make the OVS CPU pinning be enabled without restarting `ovnkube-controller` pod. Watch the parent folder instead of the enabler file because if the file gets deleted, the file watch will not emit any further fsevent.

Backport of
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5015

### DIFF
--- a/go-controller/pkg/node/ovspinning/ovspinning_linux.go
+++ b/go-controller/pkg/node/ovspinning/ovspinning_linux.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


zeeke:
@ricky-rav @tssurya please take a look when you have a chance

PR #2495, opened 2025-03-19
[release-4.18][WIP][DNM] Update OVN to FDP25.A.1 24.09.2-41 for testing.

dceara:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI

PR #2506, opened 2025-04-03
Clean git merge (sort of)

jluhrsen:
somehow we let in a single commit [0] to 4.18 to bump OVN and that is now also coming
as a duplicate, but we may need to just accept these one-offs if/when they happen until
auto branch sync is really happening and people are used to it.

[0] https://github.com/openshift/ovn-kubernetes/pull/2495


jluhrsen:
wrong repo, meant for my fork as a POC

PR #2523, opened 2025-04-28
WIP: OCPBUGS-54806: Add metrics for UDN

danwinship:
/cc


jan--f:
This looks good from the metrics perspective. One possible improvement would be to track two counter metrics per networktype (user and cluster): `user_defined_network_created_total` and `user_defined_network_destroyed_total`.
That would allow to get the current count (created-destroyed) but also rates at which networks are created and destroyed. 

If those setups are expected to change regularly this could be helpful. For largely static configurations this doesn't add much.

danwinship:
ClusterUserDefinedNetworkPolicies shouldn't change much once the cluster is stable. UDNs will either also be stable-ish, or else rate of UDN change is just going to match rate of Namespace creation/deletion (which we are presumably either already tracking, or else don't care about).

(Also, I just copied what we're doing for AdminNetworkPolicy.)

PR #2525, opened 2025-04-29
Dockerfile: Unpin OVS and consume the latest from FDP.

igsilya:
OVN-Kubernetes is always lagging behind on the version of OVS it pins. This is causing a lot of trouble with keeping up with bug fixes and especially CVE fixes on older branches, resulting in scanners constantly flagging this image with poor security grades.

OVS package inside the container is responsible for the following:

  1. Command line utilities to talk with OVS from the host.
  2. ovsdb-server processes serving OVN databases.
  3. ovs-monitor-ipsec script for managing ipsec configuration on OVN tunnels.

These tools/programs are not changing that much between patch releases, and bug fix releases in FDP are going through a lot of testing before becoming available in the repo.  So, benefits of timely delivery of bug and CVE fixes significantly outweighs the small risks that automatic consumption of new builds incurs.  Main OVS is working on the host and follows FDP for a very long time now, and it's also better to keep the minor versions between host and container in sync, just to decrease the amount of variables in the system.

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -12,25 +12,28 @@ RUN dnf --setopt=retries=2 --setopt=timeout=2 install -y --nodocs \
 	selinux-policy procps-ng && \
 	dnf clean all
 
-ARG ovsver=3.5.0-0.9.el9fdp
+# NOTE: OVS is not pinned to a particular patch version in order to stay in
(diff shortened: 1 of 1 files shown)


igsilya:
Cc: @tssurya

igsilya:
Failures do not seem related.
/retest

PR #2583, opened 2025-05-22
WIP: ovspinning for scheduable control plane

Tal-or:
## 📑 Description
dynamic ovs pinning feature works as long as ovnkube-node process is running on the cpus which are dedicated
for bustable/best-effort pods.

as part of the a schedulable contorl-plane development, we should guard the critical infra
pods from starving, by enabling workload-partitioning.

That means the ovnkube-node process will be affine to the reserved-cpus and that will break
the dynamic ovs pinning.

To overcome this problem, we're suggesting a different way to detect the non-pinned cpus
which is using podresourceAPI.
podresourceAPI allows ovnkube-node to query kubelet for all the cpus which are not
affine to guaranteed workloads, pin ovsvswitch and ovsdbserver process to those CPUs,
as was done before.

In addition, reaffine the CPUs will be triggered by event and not by time interval as before.
ovnkube-node watches the cgroup kubepods.slice and reaffine the processes everytime a change
occurs under the slice.

Fixes # https://issues.redhat.com/browse/CNF-11587

N/A

## ✅ Checks

- [X] My code requires tests

## How to verify it
Unit tests included.
E2E tests are WIP


go-controller/pkg/node/ovspinning/ovspinning_podresourceapi.go
```diff
@@ -0,0 +1,77 @@
...
+func (c *PodResClient) Client() podresourcesapi.PodResourcesListerClient {
+	return c.client
+}
+
+// GetNonPinnedCPUs calculates and returns all allocatable CPUs on the node which are not
+// exclusively pinned to any container. IOW it returns the CPUs that are dedicated for
+// Burstable and BestEffort QoS containers
+func (c *PodResClient) GetNonPinnedCPUs(ctx context.Context) (cpuset.CPUSet, error) {
+	// Get allocatable CPUs
+	allocatableResp, err := c.client.GetAllocatableResources(ctx, &podresourcesapi.AllocatableResourcesRequest{})
```
MarSik, line 47: Will this include reserved cpus when Workload partitioning is in use? Or after the containers are restricted to isolated only by the new kubelet option https://github.com/kubernetes/kubernetes/pull/127483?

Tal-or: Independent of Workload partitioning the allocatable cpus are always minus reserved:
https://github.com/kubernetes/kubernetes/blob/e51804a16b09ca4914c90e1ee45bfa4b174f7cf1/pkg/kubelet/cm/cpumanager/policy_static.go#L274

Tal-or: Done.
See new commit

PR #2628, opened 2025-06-11
QE/DNM: EgressIP: fix startup syncer to parse IP instead of CIDR

martinkennelly:
Initial implementations erroneously assumed a CIDR.

This was not caught by unit tests because the unit test also contained the assumption of only logical IP with no mask.

It was not caught by upstream CI because we have no reboot tests.
 
/hold

/cc


jechen0648:
/label qe-approved
did pre-merge verification again after new commits, verification passed

PR #2697, opened 2025-07-29
OCPBUGS-59934: Dockerfile: Remove ovs version pinning

sdodson:


sdodson:
/retitle OCPBUGS-59934: Dockerfile: Remove ovs version pinning

sdodson:
@igsilya Hmm, it looks like it's trying to include 9.2 repos but failing, also this is only for CI builds, when we get to the nightly builds those are functioning properly and have newer libibverbs-44 :man_shrugging: 

```
2025-07-29T19:11:01.739187728Z localdev-rhel-9-baseos-rpms                     0.0  B/s |   0  B     00:00    
2025-07-29T19:11:01.739357961Z Errors during downloading metadata for repository 'localdev-rhel-9-baseos-rpms':
2025-07-29T19:11:01.739357961Z   - Curl error (6): Couldn't resolve host name for https://rhsm-pulp.corp.redhat.com/content/e4s/rhel9/9.2/x86_64/baseos/os/repodata/repomd.xml [Could not resolve host: rhsm-pulp.corp.redhat.com]
2025-07-29T19:11:01.739467854Z Error: Failed to download metadata for repo 'localdev-rhel-9-baseos-rpms': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

The same error shows up in 4.14 builds as well but doesn't seem to affect things there, so not sure what's going on.

Given 4.13 is EOL and the current version shipped at the same time as the el8 version used in 4.12 I'm going to back burner this effort for 4.13 and 4.12 until image grading issues show up again.

Do you know if moving the buildroot forward is an intentional strategy likely to be repeated in future versions? Do we expect openvswitch3.3 to move beyond 9.4 build root and we run into the same problems on 4.16-4.18?

sdodson:
Thanks, I was going to chase that today.

For the latest branches where we're looking at 9.4 vs 9.6 lets wait on those, we're actually shipping 4.19 based on 9.4 container content right now though 4.20 is supposed to be moving to 9.6 and 4.19 will come after that.

PR #2698, opened 2025-07-29
OCPBUGS-59935: Dockerfile: Remove ovs version pinning

sdodson:


igsilya:
These failing required jobs are failing since at least February.  Some of them have no record of successful runs at all.

PR #2741, opened 2025-09-02
[DNM] [release-4.18] EgressIP: Lock the nodes in lexicographical order

pperiyasamy:
The EgressIP doesn't acquire node lock in the same order while processing pod events, this may lead to deadlock scenario when EgressIP and Pod are placed on two different nodes. Hence this commit ensures node are locked in lexicographical order so that two goroutines never get into deadlock scenario while processing pod event for two different EIP objects.

This PR is just created to review and test this potential fix on a setup where issue can be reproduced.

cc @martinkennelly @tssurya


martinkennelly:
If you cannot produce a unit test to minic it without changing the product, I guess we'd need to describe why upstream. Thanks peri

martinkennelly:
Please document for support how a cu can get into this situation and how they can leave it.

pliurh:
Why locking the nodes in lexicographical order may not help here?
For example:
- Goroutine 1: Locks nodeA (status.Node), then tries to lock nodeB (pod.Spec.NodeName).
- Goroutine 2: Locks nodeB (status.Node), then tries to lock nodeA (pod.Spec.NodeName).

We have a deadlock now. But if we lock the nodes in lexicographical order, the deadlock can be avoided. Since a goroutine must get the nodeA lock first.

pperiyasamy:
@martinkennelly @pliurh posted an upstream PR https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5555, PTAL. We need get this into d/s ASAP and get this synced until 4.18 to fix a customer bug https://issues.redhat.com/browse/OCPBUGS-59744. 
The fix is validated and it's working fine as per comment [here](https://issues.redhat.com/browse/OCPBUGS-59744?focusedId=27991159&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27991159).

PR #2782, opened 2025-10-06
[DNM][IC][EIP][release-4.21]  Fix race between SB DB and OVN-Controller

martinkennelly:
Ensure ovn-controller has processed the SB DB updates before removing the GARP drop flows by utilizing the hv_cfg field in NB_Global [1]

OVNKube controller increments the nb_cfg value post sync, which is copied to SB DB by northd. OVN-Controllers copy this nb_cfg value from SB DB and write it to their chassis_private tables nb_cfg field after they have processed the SB DB changes. Northd will then look at all the chassis_private tables nb_cfg value and set the NB DBs Nb_global hv_cfg value to the min integer found.

Since IC currently only supports one node per zone, we can be sure ovn-controller is running locally and therefore its ok to block removing the drop GARP flows.

[1] https://man7.org/linux/man-pages/man5/ovn-nb.5.html

(cherry picked from commit 2d70a9f7bfc5eb20f048c7d2850913c6ba9d55cf)

/hold

### DIFF
--- a/go-controller/cmd/ovnkube/ovnkube.go
+++ b/go-controller/cmd/ovnkube/ovnkube.go
@@ -536,12 +536,12 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 			metrics.MetricOVNKubeControllerReadyDuration.Set(time.Since(startTime).Seconds())
 
 			if isOVNKubeControllerSyncd != nil {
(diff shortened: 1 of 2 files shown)


martinkennelly:
/retest

Saw nothing strange just infra setup issues.

martinkennelly:
/close

Will be in latest downstream merge

PR #3050, opened 2026-03-12
OCPBUGS-72533, OCPBUGS-78771, OCPBUGS-99477: Branch Sync release-4.22 to release-4.21 [03-11-2026]

jluhrsen:
Branch sync of first 1/3 of release-4.22 into release-4.21 using `git merge` (not cherry-pick).

  414 total commits to sync, split into thirds (~138 each). This PR merges up to `52eee0790` ("E2E, networksegmentation: Use 172.16.0.0/16 for UDN tests to avoid conflicts").

  `git merge` auto-resolved all conflicts, but `-X theirs` silently broke one file:

  | File | Issue | Fix |
  |---|---|---|
  | `go-controller/pkg/node/default_node_network_controller.go` | Incoming code re-added `node`/`nodeAnnotator` in `Start()`, but their consumer (IC zone migration hack) was already removed from 4.21. Build failed with `undefined: node` and `declared and not used: nodeAnnotator`. | Removed the 5 incoming lines to match 4.21. |


jluhrsen:
> @jluhrsen is it possible to remove the ovn-large and ovn-xlarge perfscale jobs from the list of the jobs for the rerun so they are not triggered every time? They were originally triggered due to the issue fixed now in [openshift/release#76149](https://github.com/openshift/release/pull/76149) , but the rerun still triggers them. I remember that there was a way to block things from the rerun on a PR, but I don't remember what it was...

I don't think there is a way. we could create a new PR and those jobs would no longer.

Otherwise, if we only retest by issuing `/test <JOBNAME>` and avoid those perfscale jobs we should be ok. I'll try to stay on top of that so others won't have to worry about that.

arkadeepsen:
Since only a placeholder bug is linked to the PR adding the verified label

/verified by ci

RamLavi:
/retitle OCPBUGS-72533, OCPBUGS-78771: Branch Sync release-4.22 to release-4.21 [03-11-2026]

PR #3093, opened 2026-03-26
[release-4.20] OCPBUGS-78777: Fix NAD Controller syncAll for networkID upgrade from node->NAD

qinqon:
## 📑 Description

Minimal surgical backport of upstream [ovn-org/ovn-kubernetes#5705](https://github.com/ovn-org/ovn-kubernetes/pull/5705) to release-4.20, without pulling in the tunnel-keys (#5561) or callback registration (#5688) dependencies.

Related Jira: [OCPBUGS-78777](https://issues.redhat.com/browse/OCPBUGS-78777) — Upgrading with NADs using same network name causes VMs using a ovn-k8s-cni-overlay localnet NAD to lose connectivity due to a missing logical port on some of the VMs.

The fix is present in `openshift/release-4.21+` but not backported to 4.20 or earlier.

The core problem is that with the change to move the networkID from nodes to NADs, the upgrade logic left a gap in time where pods could not start. This is because cluster-manager was responsible for migrating the networkID from the node→NAD, and in our upgrade strategy, workers upgrade before control plane nodes. This would leave worker nodes in a state where new OVNK code was running that was only looking for the networkID on the NAD, but it had not yet been migrated.

### Changes:
- **`networkIDAllocator` is now always created** (unconditionally) for all controller modes. It serves as a cache for zone/node modes, and as the actual allocator for cluster manager. An explicit `isClusterManager` flag replaces the old nil-check pattern.
- **`syncAll` now behaves the same for all modes**: NADs missing the ID annotation are skipped in the first pass, their IDs are looked up per-NAD from node annotations (via new `getNetworkIDFromNode` helper), reserved in the allocator, and only then synced.
- **`handleNetworkID`** checks the allocator cache first, then falls back to the NAD annotation. Non-cluster-manager modes set the found ID and return without trying to allocate new IDs or annotate NADs.
- **`GetID()`** added to the `id.Allocator` interface to read existing mappings without allocating.

### Dependency chain (upstream):
```
#5561 — tunnel keys allocator for layer2 networks
  └─▶ #5688 — register callbacks API in NetworkManager
        └─▶ #5705 — NAD controller syncAll: unify cluster-manager / node behavior for networkID upgrade
```

This backport avoids the tunnel-keys and callback dependencies by using an `isClusterManager` bool flag instead of `tunnelKeysAllocator \!= nil`, and omitting tunnel-key annotation handling entirely.

## ✅ Checks
- [x] My code requires tests

## How to verify it

Existing unit tests updated and passing:
- `TestNADController` — all 16 subtests pass
- `TestSyncAll` — both subtests pass
- Full `go build ./...` passes with no errors

### Manual verification on an OpenShift cluster

The bug: during upgrade, workers run new OVNK code that looks for `networkID` on the NAD annotation, but cluster-manager (still on old control plane) hasn't migrated it from node annotations yet — pods on localnet NADs lose connectivity.

To exercise the fix:

1. **Create a localnet NAD** with `type: ovn-k8s-cni-overlay`

2. **Create a pod attached to it**, confirm it works

3. **Remove the `k8s.ovn.org/network-id` annotation** from the NAD to simulate the pre-migration state:
   ```
   oc annotate net-attach-def <name> -n <ns> k8s.ovn.org/network-id-
   ```

4. **Restart `ovnkube-node`** on a worker to trigger `syncAll`:
   ```
   oc delete pod -n openshift-ovn-kubernetes -l app=ovnkube-node --field-selector spec.nodeName=<worker>
   ```

5. **Verify the pod still has connectivity** after the restart

**Without the fix:** ovnkube-node can't find the networkID and the network breaks.
**With the fix:** `syncAll` falls back to reading the networkID from the node annotation (`getNetworkIDFromNode`), reserves it in the allocator, and everything keeps working.

### DIFF
--- a/go-controller/pkg/allocator/id/allocator.go
+++ b/go-controller/pkg/allocator/id/allocator.go
@@ -16,6 +16,7 @@ type Allocator interface {
 	AllocateID(name string) (int, error)
 	ReserveID(name string, id int) error
 	ReleaseID(name string)
+	GetID(name string) int
 	ForName(name string) NamedAllocator
 }
 
@@ -91,6 +92,16 @@ func (idAllocator *idAllocator) ReleaseID(name string) {
 		idAllocator.idBitmap.Release(v.(int))
 		idAllocator.nameIdMap.Delete(name)
 	}
+}
+
+// GetID returns the id for the resource 'name' if it has been previously
+// allocated or reserved. Returns invalidID if no id is found.
+func (idAllocator *idAllocator) GetID(name string) int {
+	v, ok := idAllocator.nameIdMap.Load(name)
+	if ok {
+		return v.(int)
+	}
+	return invalidID
 }
 
 func (idAllocator *idAllocator) ForName(name string) NamedAllocator {
--- a/go-controller/pkg/clustermanager/pod/allocator_test.go
+++ b/go-controller/pkg/clustermanager/pod/allocator_test.go
@@ -146,6 +146,10 @@ func (a *idAllocatorStub) ReleaseID(string) {
 	a.released = true
 }
 
+func (a *idAllocatorStub) GetID(string) int {
+	panic("not implemented") // TODO: Implement
+}
+
 func (a *idAllocatorStub) ForName(string) id.NamedAllocator {
 	panic("not implemented") // TODO: Implement
 }
--- a/go-controller/pkg/networkmanager/nad_controller.go
+++ b/go-controller/pkg/networkmanager/nad_controller.go
@@ -3,6 +3,7 @@ package networkmanager
 import (
 	"fmt"
 	"reflect"
+	"sort"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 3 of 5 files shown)


jcaamano:
@qinqon why are you not cherry-picking the unit tests which should be testing what you are actually cherry-picking?

qinqon:
> @qinqon why are you not cherry-picking the unit tests which should be testing what you are actually cherry-picking?

@jcaamano done

jcaamano:
> > @qinqon why are you not cherry-picking the unit tests which should be testing what you are actually cherry-picking?
> 
> @jcaamano done

lint issues

qinqon:
> > > @qinqon why are you not cherry-picking the unit tests which should be testing what you are actually cherry-picking?
> > 
> > 
> > @jcaamano done
> 
> lint issues

Fixed

servolkov:
/verified by @servolkov 

https://redhat.atlassian.net/browse/OCPBUGS-78777?focusedCommentId=16770537

PR #3141, opened 2026-04-15
Add node IP to HostNetworkNamespace address_set

asood-rh:
When NoOverlay mode is used for a network, it uses learned route with proto bgp and that sets Node IP as source IP.

10.129.2.0/23 nhid 157 via 192.168.100.100 dev br-ex proto bgp metric 20

So, it is essential to add node IP to HostNetworkNamespace address_set to let host network POD use network-policy while using NoOverlay mode.

## 📑 Description
This PR is just to pre merge upstream commit with CNO PR https://github.com/openshift/cluster-network-operator/pull/2960 for bug https://redhat.atlassian.net/browse/OCPBUGS-83406

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Host‑network namespace address set in NoOverlay mode now includes nodes' primary interface addresses.

* **Tests**
  * Added a test verifying the host‑network address set contains the node primary IPs when NoOverlay mode is enabled.

### DIFF
--- a/go-controller/pkg/ovn/master_test.go
+++ b/go-controller/pkg/ovn/master_test.go
@@ -1838,6 +1838,66 @@ var _ = ginkgo.Describe("Default network controller operations", func() {
 		})
 		gomega.Expect(err).NotTo(gomega.HaveOccurred())
 	})
+
+	ginkgo.It("includes node primary IP in host-network-namespace address_set when NoOverlay mode is enabled", func() {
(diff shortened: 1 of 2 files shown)


asood-rh:
/close

The commit in this PR is downstream in https://github.com/openshift/ovn-kubernetes/pull/3169

PR #3167, opened 2026-04-29
Fix IPv6 CIDR flapping in k8s.ovn.org/host-cidrs annotation

smulje:
## Summary

Fixes a race condition where the `k8s.ovn.org/host-cidrs` annotation continuously flaps between dual-stack and IPv4-only states on nodes with IPv6 enabled (particularly with SLAAC).

## Problem

On nodes with IPv6 enabled, the `k8s.ovn.org/host-cidrs` annotation exhibits unstable behavior, alternating between:
- `["10.46.xx.62/21","2620:52:0:2ef8:7058:xxx:3bd9:e3dc/64"]` (dual-stack)
- `["10.46.xx.62/21"]` (IPv4-only)

This flapping occurs every 30-60 seconds and impacts workloads that depend on stable host network information.

### Root Cause

The `addressManager.sync()` function has a race condition with IPv6 Duplicate Address Detection (DAD):

1. **Periodic sync** scans network interfaces to rebuild the address map
2. **IPv6 DAD timing**: IPv6 addresses may still be in "tentative" state during the scan
3. **Result**: sync() misses the IPv6 address and updates annotation to IPv4-only
4. **Netlink watcher** later discovers the IPv6 address via kernel events
5. **Result**: Annotation updated back to dual-stack
6. **Repeat**: Next sync cycle triggers the same race

## Solution

Perform **two interface scans with a 100ms delay** and take the **union** of both results. This ensures IPv6 addresses that are completing DAD during the first scan are captured in the second scan.

### Changes

1. **New helper function** `scanInterfaceAddresses()`: Extracts interface scanning logic for reuse and returns `(sets.Set[string], error)` to propagate failures

2. **Modified sync() function**: Performs double-scan with union and includes error handling with fallback logic:
   - Both scans succeed: use union (catches IPv6 during DAD)
   - One scan fails: use the successful scan result
   - Both scans fail: abort early to preserve existing annotation

3. **Better error handling**: Continues on per-interface errors instead of aborting entire sync, and prevents transient `netlink.LinkList()` failures from incorrectly clearing the annotation

The 100ms delay accommodates typical IPv6 DAD completion times (most complete within 100-200ms) without significantly impacting sync performance (runs every 30s).

### Reproduction Environment
- **Cluster**: OpenShift 4.20.1, OVN-Kubernetes
- **Network**: Dual-stack (IPv4 + IPv6 via SLAAC)
- **Affected nodes**: All nodes with IPv6 enabled
- **Kernel**: RHEL CoreOS 9.6

### Reproduction Steps
1. Enable IPv6 on cluster nodes (via SLAAC or static)
2. Monitor annotation: `oc get node <node> -o jsonpath='{.metadata.annotations.k8s\.ovn\.org/host-cidrs}'`
3. Observe flapping between dual-stack and IPv4-only every 30-70 seconds

### Verification

**Before fix**: Flapping detected every 40-70 seconds  
**After fix**: Annotation remains stable with dual-stack

## Impact

- **Positive**: Eliminates annotation instability on IPv6-enabled clusters
- **Performance**: Negligible (adds 100ms to sync interval, which runs every 30s)
- **Compatibility**: No breaking changes, backward compatible
- **Side effects**: None identified

## Related Issues

- Affects OpenShift 4.15+ (when k8s.ovn.org/host-cidrs annotation was introduced)
- Impacts egress IP feature on dual-stack clusters
- May affect any controller watching node annotations for network changes

### DIFF
--- a/go-controller/pkg/node/node_ip_handler_linux.go
+++ b/go-controller/pkg/node/node_ip_handler_linux.go
@@ -548,7 +548,7 @@ func (c *addressManager) sync() {
 			return
 		}
 		for _, link := range links {
-			foundAddrs, err := util.GetNetLinkOps().AddrList(link, getSupportedIPFamily())
+			foundAddrs, err := util.GetNetLinkOps().AddrList(link, 0)
 			if err != nil {
 				klog.Errorf("Failed sync due to being unable to list addresses for %q: %v", link.Attrs().Name, err)
 				return


tssurya:
all fixes must go upstream
coming as part of project maintainence here, PR is too long in stale state in draft mode
please close PR after use @smulje

smulje:
Hi @tssurya, thanks for the reminder. This was mainly for learning purposes, so I’ll close this one.

PR #3183, opened 2026-05-11
OCPBUGS-85366: ANP-Skip redundant status updates to reduce API server load

tssurya:
When reconciling AdminNetworkPolicy or BaselineAdminNetworkPolicy, the controller updates the status condition on every successful sync, even when the status is already in the desired state. At scale (500-1000 nodes), this causes unnecessary API server load spikes since each zone's controller repeatedly patches the same CRD.

Add a doesStatusNeedAnUpdate() utility function that compares the existing condition with the new condition. If Status, Reason, and Message are all unchanged, skip the ApplyStatus call entirely. This optimization applies to both ANP and BANP status update paths.

Also move the "Patched status" log messages into the functions that perform the actual API call, so they only log when an update actually happens (not when skipped due to no-op).

(cherry picked from commit 322e391160cb6186750feb713db22681c5d47b77)
no conflicts

4.21 merge: https://github.com/openshift/ovn-kubernetes/pull/3080
4.22 merge: https://github.com/openshift/ovn-kubernetes/pull/2980

### DIFF
--- a/go-controller/pkg/ovn/controller/admin_network_policy/status.go
+++ b/go-controller/pkg/ovn/controller/admin_network_policy/status.go
@@ -46,6 +46,22 @@ const (
 	policyNotReadyReason = "SetupFailed"
 )
 
+// doesStatusNeedAnUpdate compares the existing condition with the new condition
+// and returns true if an update is needed, false if the status is already in the desired state.
+// This helps avoid unnecessary API server calls when the status hasn't changed.
+func doesStatusNeedAnUpdate(existingCondition *metav1.Condition, newCondition metav1.Condition) bool {
+	if existingCondition == nil {
+		return true // condition doesn't exist yet, needs to be created
+	}
+	// Check if Status, Reason, and Message are all the same - if so, no update needed
+	if existingCondition.Status == newCondition.Status &&
+		existingCondition.Reason == newCondition.Reason &&
+		existingCondition.Message == newCondition.Message {
+		return false
+	}
+	return true
+}
+
 // updateANPStatusToReady updates the status of the policy to reflect that it is ready
 // Each zone's ovnkube-controller will call this, hence let's update status using server-side-apply
 func (c *Controller) updateANPStatusToReady(anpName string) error {
@@ -59,8 +75,6 @@ func (c *Controller) updateANPStatusToReady(anpName string) error {
 	if err != nil {
 		return fmt.Errorf("unable to update the status of ANP %s, err: %v", anpName, err)
 	}
-	klog.V(5).Infof("Patched the status of ANP %v with condition type %v/%v",
-		anpName, policyReadyStatusType+c.zone, metav1.ConditionTrue)
 	return nil
 }
 
@@ -83,8 +97,6 @@ func (c *Controller) updateANPStatusToNotReady(anpName, message string) error {
 	if err != nil {
 		return fmt.Errorf("unable update the status of ANP %s, err: %v", anpName, err)
 	}
-	klog.V(3).Infof("Patched the status of ANP %v with condition type %v/%v and reason %s/%s",
-		anpName, policyReadyStatusType+c.zone, metav1.ConditionFalse, policyNotReadyReason, message)
 	return nil
 }
 
@@ -94,6 +106,10 @@ func (c *Controller) updateANPZoneStatusCondition(newCondition metav1.Condition,
 		return err
 	}
 	existingCondition := meta.FindStatusCondition(anp.Status.Conditions, newCondition.Type)
+	if !doesStatusNeedAnUpdate(existingCondition, newCondition) {
+		// status is already in the desired state, skip the update to reduce API server load
+		return nil
+	}
 	if existingCondition == nil {
 		newCondition.LastTransitionTime = metav1.NewTime(time.Now())
 	} else {
@@ -109,6 +125,10 @@ func (c *Controller) updateANPZoneStatusCondition(newCondition metav1.Condition,
 		WithStatus(anpapiapply.AdminNetworkPolicyStatus().WithConditions(newCondition))
 	_, err = c.anpClientSet.PolicyV1alpha1().AdminNetworkPolicies().
 		ApplyStatus(context.TODO(), applyObj, metav1.ApplyOptions{FieldManager: c.zone, Force: true})
+	if err == nil {
+		klog.V(5).Infof("Patched the status of ANP %s with condition type %s/%s, reason %s, message: %s",
+			anpName, newCondition.Type, newCondition.Status, newCondition.Reason, newCondition.Message)
+	}
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


jcaamano:
@tssurya I know this is a backport but I have to ask.

Since we are comparing the status with cache which might not be current, could this led to incorrect status?
For example, considering this timeline:

T0 status ready
T1 update status not ready
T2 after a namespace event, attempt to update status to ready but our cache is at T0 so we don't update status because we see it is the same status

tssurya:
> Since we are comparing the status with cache which might not be current, could this led to incorrect status? For example, considering this timeline:
> 
> T0 status ready T1 update status not ready T2 after a namespace event, attempt to update status to ready but our cache is at T0 so we don't update status because we see it is the same status

So the controller itself has a central lock so no concurrent status writes can happen, however if the informer cache itself which I use to determine the status to compare is lagged then we can hit the race condition which I believe shouldn't be too frequent unless its a high churn cluster - even then I would take that informer v/s API lag bug causing a incorrect state until next reconcile solves it v/s the scale issue we have right now (if we want to prevent the total race with informer cache lag I'd need to use API list directly OR revert this change to always apply which brings us back to the scale issue).

But in general we have agreed upstream that per-zone status is a bad idea based on experience and feedback we have gotten, so all this is going to go away soon with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6418 - we are moving towards metrics and alerts more and more for the zone side things..and removing https://github.com/ovn-kubernetes/ovn-kubernetes/tree/master/go-controller/pkg/clustermanager/status_manager

tssurya:
> @tssurya I know this is a backport but I have to ask.
> 
> Since we are comparing the status with cache which might not be current, could this led to incorrect status? For example, considering this timeline:
> 
> T0 status ready T1 update status not ready T2 after a namespace event, attempt to update status to ready but our cache is at T0 so we don't update status because we see it is the same status

actually is this bug even possible? cause my controller has a central lock, so if status was updated to not ready.. it was through a specific reconcile.. here if there is a lag then the next reconcile will fix it ? I doubt when the namespace even t comes in if there is going to be that much lag ? I do think theoretically informer bug is possible, but practically finding it hard to hit/repro

jcaamano:
I don't think a central lock prevents it. Yes, subsequent reconciles can straighten the issue ... if they come ... or if you are reconciling on your own updates (not great).

> even then I would take that informer v/s API lag bug causing a incorrect state until next reconcile solves it v/s the scale issue we have right now

Subjective, not sure I would be of the same opinion. But just because it is subjective and the incorrect status issue has not actually been observed, I don't think I have much issue with your opinion either. Just wanted to make you aware.

> (if we want to prevent the total race with informer cache lag I'd need to use API list directly OR revert this change to always apply which brings us back to the scale issue)

As an alternative example, the RA controller tracks if the reconciliation triggered any actual downstream change, and if it didn't does not update the status.

In general, I think comparing with your previous state to inhibit updates should be avoided.

Anyway, proceeding with this.

tssurya:
> As an alternative example, the RA controller tracks if the reconciliation triggered any actual downstream change, and if it didn't does not update the status.
> 
> In general, I think comparing with your previous state to inhibit updates should be avoided.

ah you use some kind of hadUpdates v/s not ? how does RA controller avoid the informer lag in the subsequent update?

tssurya:
> I don't think a central lock prevents it. Yes, subsequent reconciles can straighten the issue ... if they come ... or if you are reconciling on your own updates (not great).

yea its the "if they come" case, we don't trigger reconcile on our own status updates..

SachinNinganure:
sninganu@sninganu-thinkpadt14gen4:~/Downloads$ date;oc version
Fri May 29 03:06:12 PM IST 2026
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: 4.20.0-0-2026-05-29-083655-test-ci-ln-dsft1ib-latest
Kubernetes Version: v1.33.12  

 

sninganu@sninganu-thinkpadt14gen4:~/Devel/workdir/anp-bug$ oc apply -f templates/banp-template.yaml
[baselineadminnetworkpolicy.policy.networking.k8s.io/default](http://baselineadminnetworkpolicy.policy.networking.k8s.io/default) created
sninganu@sninganu-thinkpadt14gen4:~/Devel/workdir/anp-bug$  # Create 70 projects (namespaces)

sninganu@sninganu-thinkpadt14gen4:~/Devel/workdir/anp-bug$ for i in {1..70}; do
  NS_NAME="bug-desc-ns-${i}"
  TEAM_ID=$((i % 5 + 1))
  echo "Creating namespace ${NS_NAME}..."
  cat <<EOF | oc apply -f -
apiVersion: v1
kind: Namespace
metadata:
  name: ${NS_NAME}
  labels:
    bug-test: "true"
    team: "team-${TEAM_ID}"
EOF
done
Creating namespace bug-desc-ns-1...
namespace/bug-desc-ns-1 created
Creating namespace bug-desc-ns-2...
..................
Creating namespace bug-desc-ns-70...
namespace/bug-desc-ns-70 created

 

sninganu@sninganu-thinkpadt14gen4:~/Devel/workdir/anp-bug$   echo "Created namespace ${NS_NAME}"
  done
   echo "Namespaces created. Now creating ANPs..."

Namespaces created. Now creating ANPs...

[adminnetworkpolicy.policy.networking.k8s.io/bug-desc-anp-1](http://adminnetworkpolicy.policy.networking.k8s.io/bug-desc-anp-1) created
Created ANP bug-desc-anp-1 (priority 101)
[adminnetworkpolicy.policy.networking.k8s.io/bug-desc-anp-2](http://adminnetworkpolicy.policy.networking.k8s.io/bug-desc-anp-2) created
.......................................
Created ANP bug-desc-anp-69 (priority 169)
[adminnetworkpolicy.policy.networking.k8s.io/bug-desc-anp-70](http://adminnetworkpolicy.policy.networking.k8s.io/bug-desc-anp-70) created
Created ANP bug-desc-anp-70 (priority 170)
sninganu@sninganu-thinkpadt14gen4:~/Devel/workdir/anp-bug$   echo ""
  echo "========================================="
  echo "Verification"
  echo "========================================="
  echo "Namespaces created: $(oc get namespaces -l bug-test=true --no-headers | wc -l)"
  echo "ANPs created: $(oc get anp --no-headers | wc -l)"
  echo "BANP created: $(oc get banp --no-headers | wc -l)"
  echo ""

=========================================
Verification

Namespaces created: 70
ANPs created: 70
BANP created: 1

  NO CPU spikes observed

    70 namespaces

    70 ANPs

    1 BANP

cronjob run every 5 minutes

108 | Time | Event | Master CPU Peak | kube-apiserver Peak | ANP Logs | Pattern |
  109 |------|-------|----------------|---------------------|----------|---------|
  110 | 15:25:05 | :25 CronJob trigger | 588m (3%) | 619m | 115 |  Moderate |
  111 | 15:30:24 | :30 CronJob trigger | 1248m (8%) | 359m | 130 |  Normal |
  112 | 15:35:07 | :35 CronJob trigger | 500m (3%) | 196m | 173 |  Normal |
 
  with 140 CronJobs creating pod churn, no massive CPU spikes observed

PR #3189, opened 2026-05-14
DNM: e2e: replace random allocations with ConfigMap-backed allocators

jcaamano:


test/e2e/allocators/bgp.go
```diff
@@ -0,0 +1,105 @@
...
+	bgpOnce.Do(func() {
+		bgpPeerV4 = newSubnetSpec(bgpPeerSubnets, nil)
+		bgpPeerV6 = newSubnetSpec(bgpPeerSubnets6, nil)
+		ipvrfV4 = newSubnetSpec(ipvrfSubnets, nil)
+		ipvrfV6 = newSubnetSpec(ipvrfSubnets6, nil)
+		vtepV4 = newSubnetSpec(vtepSubnets, nil)
+		vtepV6 = newSubnetSpec(vtepSubnets6, nil)
+	})
+
+	maxUsable := min(
```
pperiyasamy, line 67: here maxUsage is derived from min value of various subnets, vid and VNI, now let's say one subnet has a wider usable CIDRs and another subnet has a narrow usable CIDRs. so doesn't it impose restriction  in allocating CIDR for every subnet (meaning some CIDRs never allocated from a subnet which has wider allocatable CIDRs) ?

test/e2e/allocators/int.go
```diff
@@ -0,0 +1,120 @@
...
+	_, err = client.Create(ctx, cm, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return 0, fmt.Errorf("failed to ensure allocator config map: %w", err)
+	}
+
+	var allocated int
+	err = retry.OnError(
+		allocatorBackoff,
+		func(err error) bool {
+			return apierrors.IsConflict(err) || errors.Is(err, errAllocationFull)
```
pperiyasamy, line 66: so return true for `errors.Is(err, errAllocationFull) ` make allocation to retry until another test to clean up the allocated id ?

jcaamano: yes

test/e2e/allocators/bgp.go
jcaamano: Yup.
The compromise is allocating a bunch of stuff from a single int which I find acceptable, don't see an issue with this.
Right now I am just going with the ranges we had but of course it would make sense to make some of them narrower & aligned now that we are not randomizing.
@pperiyasamy review https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6404, this is just a DNM PR

PR #3217, opened 2026-06-02
CORENET-7129 Add OVN metrics tests as OTP tests

asood-rh:
Adds 8  metrics tests -
- 8 tests as informing.
[sig-networking] OVN metrics [OTP] 45841-Add OVN flow count metric
[sig-networking] OVN metrics [OTP] 47471-Record update to cache versus port binding
[sig-networking] OVN metrics [OTP] 52072-Add mechanism to record duration for k8 kinds
[sig-networking] OVN metrics [OTP] 47524-Metrics for ovn-appctl stopwatch/show command
[sig-networking] OVN metrics [OTP] 45689-Metrics for idling enable/disabled
[sig-networking] OVN metrics [Serial] [OTP] 60539-Verify metrics ovs_vswitchd_interfaces_total
[sig-networking] OVN metrics [Serial] [OTP] 60704-Verify metrics ovs_vswitchd_interface_up_wait_seconds_total
[sig-networking] OVN metrics [Serial] [Slow] [OTP] 60708-Verify metrics ovnkube_resource_retry_failures_total

## Summary by CodeRabbit

* **Tests**
  * Added extensive OTP-marked OVN-Kubernetes e2e suites (NetworkPolicy, IPBlock, DNS/connectivity, ACL audit logging, stability/perf) and OVN metrics validation; test harness updated to discover/classify OTP tests with adjusted lifecycles and naming.
  * Added a validating admission policy protection test and serial stress/perf sampling.

* **Chores**
  * Updated module dependency versions and refreshed transitive dependency graph.

* **Tests**
  * Added many networking test fixtures, templates and test data assets; embedded testdata now extracts via a single cached fixture path.

* **Style**
  * Added approvers to testdata OWNERS.


asood-rh:
/retitle [CORENET-7129](https://redhat.atlassian.net/browse/CORENET-7129) Add OVN metrics tests as blocking OTP tests

asood-rh:
Tests are passing e2e-aws-ovn job
<pre>
started: 0/151/510 "[sig-networking] OVN metrics [OTP] 45841-Add OVN flow count metric"
passed: (20.2s) 2026-06-08T23:33:06 "[sig-networking] OVN metrics [OTP] 45841-Add OVN flow count metric"

started: 0/257/510 "[sig-networking] OVN metrics [OTP] 47471-Record update to cache versus port binding"
passed: (1m7s) 2026-06-08T23:34:55 "[sig-networking] OVN metrics [OTP] 47471-Record update to cache versus port binding"

started: 2/436/510 "[sig-networking] OVN metrics [OTP] 52072-Add mechanism to record duration for k8 kinds"
passed: (17s) 2026-06-08T23:36:16 "[sig-networking] OVN metrics [OTP] 52072-Add mechanism to record duration for k8 kinds"

</pre>

openshift/test/otp/metrics.go
```diff
@@ -0,0 +1,423 @@
...
+			if len(podLogs) == 0 || logErr != nil {
+				e2e.Logf("did not get expected podLogs, or have err: %v, try again", logErr)
+				return false, nil
+			}
+			return true, nil
+		})
+		o.Expect(checkErr).NotTo(o.HaveOccurred(), "fail to get expected log in pod %v", ovnMasterPodName)
+
+		g.By("4. Get the metrics of " + metricName + " again when resource retry failure occur")
+		metricValue1Int, _ := strconv.Atoi(metricValue1)
```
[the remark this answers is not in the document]
asood-rh: @coderabbitai resolve this if fixed

openshift/test/otp/networkpolicy.go
```diff
@@ -0,0 +1,2117 @@
+package otp
```
anuragthehatter: @asood-rh Thanks. Seems like networkpolicy.go commit came here by mistake? I guess you just meant metrics.go only

openshift/test/ote/metrics.go
```diff
@@ -0,0 +1,428 @@
...
+	var oc = exutil.NewCLI("networking-ovn-metrics")
+
+	g.BeforeEach(func() {
+		networkType := otputils.CheckNetworkType(oc)
+		if !strings.Contains(networkType, "ovn") {
+			g.Skip("Skip testing on non-ovn cluster!!!")
+		}
+	})
+
+	g.It("[OTP] 45841-Add OVN flow count metric", func() {
```
anuragthehatter, line 39: @asood-rh Thanks, Please add [JIRA:Networking] tag as well to all cases as per OTE guide requirement.

openshift/test/ote/metrics.go
```diff
@@ -0,0 +1,428 @@
...
+			if currentValue > baselineValue {
+				return true, nil
+			}
+			e2e.Logf("Can't get correct metrics of openshift_unidle_events and try again")
+			return false, nil
+		})
+		o.Expect(metricsOutput).NotTo(o.HaveOccurred(), "Fail to get updated metric openshift_unidle_events_total")
+	})
+
+	g.It("[Serial] [OTP] 60539-Verify metrics ovs_vswitchd_interfaces_total", func() {
```
anuragthehatter, line 257: [serial] wont work here. If we want to make all cases serial, I would just add `[Suite:openshift/conformance/serial]` tag in g.Describe.

anuragthehatter: Or if we want to make specific cases serial, I would siffix them with `g.serial` suffix tag like you have done in one of them

anuragthehatter: ah i might be wrong [serial] should also work. Unless dev has any recommendation to be consistent with g.serial or openshift/conformance/seria logic.

openshift/test/otp/networkpolicy.go
asood-rh: @anuragthehatter In order to use the utils library I created branch off your branch. I do have few changes to ensure metrics tests executed successfully. Will need to deal with merges unless you want to pick up changes in utils.go. I can rebase my branch with yours.

anuragthehatter: @asood-rh np at all, as long as you are aware of

openshift/test/ote/metrics.go
asood-rh: Done

anuragthehatter:
Thanks Arti.
lgtm to me, lets wait for results

asood-rh:
> Thanks Arti. lgtm to me, lets wait for results
Could we assign dev to it as well?
I believe we will have to deal with CI failures as they may not be related to tests.

asood-rh:
@pperiyasamy 
All 8 [OTP] tests pass consistently across all jobs. Zero PR-related failures.                                                                                                                                   
                                                                                                                                                                                                                   
  [OTP] Test Results (all PASS):                                                                                                                                                                                   
  <b style="font-weight:normal;" id="docs-internal-guid-1e295289-7fff-9095-fd58-39e447eb2ab6"><div dir="ltr" style="margin-left:0pt;" align="center">
Test | e2e-aws | e2e-gcp | e2e-upgrade
-- | -- | -- | --
[OTP] 45841 Add OVN flow count metric | 20s | 33s | 21s
[OTP] 47471 Record update to cache vs port binding | 1m8s | 1m9s | 1m8s
[OTP] 52072 Add mechanism to record duration for k8 kinds | 17s | 14s | 18s
[OTP] 47524 Metrics for ovn-appctl stopwatch/show | 4m47s | 5m11s | 4m54s
[OTP] 45689 Metrics for idling enable/disabled | 7m28s | 7m34s | 7m26s
[OTP] 60704 Verify ovs_vswitchd_interface_up_wait | 41s | 43s | 42s
[OTP] 60539 Verify ovs_vswitchd_interfaces_total | 58s | 1m24s | 58s
[OTP] 60708 Verify ovnkube_resource_retry_failures_total | 16m31s | 16m26s | 16m31s

</div></b>
  What's actually failing (all unrelated to PR):

  <b style="font-weight:normal;" id="docs-internal-guid-52f1bd0b-7fff-0d10-a5e6-57ebec7c7bc6"><div dir="ltr" style="margin-left:0pt;" align="center">
Job | Failure | Cause
-- | -- | --
e2e-aws-ovn | 5 blocking: AWS NLB + OAuth server tests | AWS cloud provider infrastructure flakes
e2e-gcp-ovn | 16 blocking across sig-storage, sig-node, sig-cli, sig-auth | GCP infrastructure instability (mass failure flag triggered: 17 > threshold 10)
e2e-aws-ovn-upgrade | MonitorTest failure | Cluster monitoring infrastructure flake, not a test assertion
security | Snyk scan: 2 LOW findings | Hardcoded userName in otp/utils/utils.go lines 3310/3334 — test fixture data, not real credentials
All e2e jobs | 10 informing [Feature:NetworkSegmentation] failures | Pre-existing UDN issues, informing only (don't block)

</div><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">&nbsp; </span></p></b><br class="Apple-interchange-newline">

Snyk security note

  The security job flagged userName variables in test utility code as "hardcoded credentials". These are test fixtures. If the security job is required to pass, renaming the variable (e.g., testUserName) would
  clear it.

asood-rh:
The following tests would need architect approval
<pre>
[OTP] 45689 Metrics for idling enable/disabled takes around 7.5 mins
[OTP] 60708 Verify ovnkube_resource_retry_failures_total	 takes 16.5 mins
</pre>

openshift/cmd/ovn-kubernetes-tests-ext/main.go
pperiyasamy, line 1: Are these numbers 45841, 47471 and 52072 in the commit message convey any info ? I can't make anything out of it.

pperiyasamy, line 1: The commit message says ``Depends-On: #3200" for utils.go", so how do you make that to be part of this PR ? I don't see any of the commits from #3200 included in this PR, so how does it work ?

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -36,6 +36,21 @@ import (
 
 var ocpInfra *ocpinfraprovider.OpenshiftInfraProvider
 
+var otpBlockingTests = []string{
+	"Add OVN flow count metric",
+	"Record update to cache versus port binding",
+	"Add mechanism to record duration for k8 kinds",
+}
+
+func isOTPBlocking(name string) bool {
```
pperiyasamy: what is the purpose of `isOTPBlocking` method ? i don't see it being invoked from anywhere in this commit. so move it to right commit where it's used and we can discuss about the need for this function.

openshift/test/otp/metrics.go
```diff
@@ -0,0 +1,106 @@
+package otp
```
pperiyasamy: why package name otp ? is it still relevant now ?

openshift/test/otp/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+
+	exutil "github.com/openshift/origin/test/extended/util"
+
+	otputils "github.com/ovn-kubernetes/ovn-kubernetes/openshift/pkg/otp"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = g.Describe("[sig-networking] OVN metrics", func() {
```
pperiyasamy: I don't think we need [sig-networking] here, all ote test are going to prepended with "[ovn-kubernetes-ote][sig-network]" label ,so this would be redunant. but let's have a feature label.

openshift/test/otp/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+	defer g.GinkgoRecover()
+
+	var oc = exutil.NewCLI("networking-ovn-metrics")
+
+	g.BeforeEach(func() {
+		networkType := otputils.CheckNetworkType(oc)
```
pperiyasamy: where is otputils ?

openshift/test/ote/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+
+var _ = g.Describe("[sig-networking] OVN metrics", func() {
+	defer g.GinkgoRecover()
+
+	var oc = exutil.NewCLI("networking-ovn-metrics")
+
+	g.BeforeEach(func() {
+		networkType := otputils.CheckNetworkType(oc)
+		if !strings.Contains(networkType, "ovn") {
+			g.Skip("Skip testing on non-ovn cluster!!!")
```
pperiyasamy, line 35: is this check still necessary ? the test is meant to run ovn type cluster

openshift/test/otp/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+		if !strings.Contains(networkType, "ovn") {
+			g.Skip("Skip testing on non-ovn cluster!!!")
+		}
+	})
+
+	g.It("[OTP] 45841-Add OVN flow count metric", func() {
+		var (
+			namespace = "openshift-ovn-kubernetes"
```
pperiyasamy: namespace can be retrieved from deploymentconfig's `OVNKubernetesNamespace()` API

openshift/test/otp/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+			ovncmName = "kube-rbac-proxy-ovn-metrics"
+			podLabel  = "app=ovnkube-node"
+		)
+
+		podName := otputils.GetLeaderInfo(oc, namespace, podLabel, "ovnkubernetes")
```
pperiyasamy: isn't `otputils.GetLeaderInfo` specific non-ic ?

openshift/test/ote/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+					return true, nil
+				}
+				e2e.Logf("Can't get correct metrics value of %s and try again", m)
+				return false, nil
+			})
+			o.Expect(metricsOutput).NotTo(o.HaveOccurred(), "Fail to get metric %s", m)
+		}
+	})
+
+	g.It("[OTP] 52072-Add mechanism to record duration for k8 kinds", func() {
```
pperiyasamy, line 96: why can't query metrics with single curl call and check expected metrics on a single test ?

openshift/pkg/ote/utils/utils.go
pperiyasamy, line 1: so this is the anurag's commit you were mentioning in the previous commit ? can you do your changes on top of his commits ?

openshift/pkg/otp/utils/utils.go
```diff
@@ -1231,6 +1231,28 @@ func GetOVNMetrics(oc *exutil.CLI, url string) string {
 	return metricsLog
 }
 
+// ExtractMetricValue reads a metrics file and returns the value (second field) from the Nth line
+// matching metricName. This replaces shell pipelines like: cat file | grep metric | awk 'NR==N{print $2}'
+func ExtractMetricValue(filePath, metricName string, matchIndex int) string {
```
pperiyasamy: have you explored retrieving metrics in json format and parsing it ? 
parsing curl output is too hard.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -165,14 +173,24 @@ func main() {
...
+			for _, label := range getPrependLabels(spec.Labels) {
+				spec.Labels.Insert(label)
+			}
 
-		spec.Name = generatePrependedLabelsStr(spec.Labels) + " " + spec.Name // prepend ginkgo labels to test name
+			spec.Name = generatePrependedLabelsStr(spec.Labels) + " " + spec.Name
+		}
 
 		switch {
+		case isOTP && isOTPBlocking(spec.Name):
```
pperiyasamy: so you are using isOTPBlocking method here, I think you may need to organize commits with self contained changes to review better.
why do you add it as blocking test right away ? shouldn't we start with informing test ? FYI, we already handle assigning lifecycle for tests here https://github.com/openshift/ovn-kubernetes/blob/main/openshift/test/tests.go. can reuse it.

pperiyasamy (commented):
@asood-rh some initial comments.

anuragthehatter:
@asood-rh Can you ci/prow/security failures? Just run go vet/go fmt. Thanks

openshift/pkg/otp/testdata/networking/hello-pod-daemonset.yaml
```diff
@@ -0,0 +1,29 @@
...
+  selector:
+    matchLabels:
+      name: hello-pod
+  template:
+    metadata:
+      labels:
+        name: hello-pod
+    spec:
+      nodeSelector:
+        beta.kubernetes.io/os: linux
```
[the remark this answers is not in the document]
asood-rh: @coderabbitai Resolve if fixed

openshift/pkg/otp/testdata/networking/networkpolicy/ingress_49696.yaml
```diff
@@ -0,0 +1,18 @@
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: allow-all-ingress
```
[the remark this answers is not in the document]
asood-rh: @coderabbitai Resolve if fixed

openshift/pkg/otp/testdata/networking/networkpolicy/ipblock/ipBlock-egress-except-dual-CIDRs-template.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: ipBlock_egress_dual_cidrs_template
```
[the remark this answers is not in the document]
asood-rh: @coderabbitai Resolve if fixed

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: @pperiyasamy These are test case ID that we are carrying over for book keeping migration effort.
In polarion they will have prefix of OCP- e.g. OCP-45041

openshift/pkg/otp/testdata/networking/networkpolicy/ipblock/ipBlock-egress-except-single-CIDR-template.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: ipBlockEgressSingle_cidr_template
```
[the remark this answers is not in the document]
asood-rh: @coderabbitai Resolve if fixed

openshift/test/otp/networkpolicy.go
```diff
@@ -0,0 +1,2117 @@
...
+		o.Expect(string(PreGoroutineOut) == string(PostGoroutineOut)).To(o.BeTrue())
+
+		g.By("6. Verify ovnk-master pod memory usage should be the same as previous.\n")
+		// wait for ovnk-master leader pod to be stable
+		checkErr = wait.PollUntilContextTimeout(context.Background(), 10*time.Second, 300*time.Second, false, func(cxt context.Context) (bool, error) {
+			checkMemory2, err := exec.Command("bash", "-c", checkMemoryCmd).CombinedOutput()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if string(checkMemory2) == string(checkMemory1) {
+				e2e.Logf("Memory usage is the same as previous.")
+				return true, nil
```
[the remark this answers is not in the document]
asood-rh: Not relevant

openshift/test/otp/networkpolicy.go
```diff
@@ -0,0 +1,2121 @@
...
+		}()
+		g.By("Set properties of duplicate ACL")
+		setAclPropertiesCmd := fmt.Sprintf("ovn-nbctl set acl %s  match='%s' priority=%s meter=%s", idOutput, aclMap["match"], aclMap["priority"], aclMap["meter"])
+		_, listErr = otputils.RemoteShPodWithBashSpecifyContainer(oc, "openshift-ovn-kubernetes", ovnKNodePod, "ovnkube-controller", setAclPropertiesCmd)
+		o.Expect(listErr).NotTo(o.HaveOccurred())
+		defer func() {
+			if os.Getenv("DELETE_NAMESPACE") != "false" {
+				oc.AsAdmin().WithoutNamespace().Run("delete").Args("networkpolicy", policyName, "-n", ns).Execute()
+			}
+		}()
```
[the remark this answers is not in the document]
asood-rh: networkpolicy.go, no longer in PR.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: @pperiyasamy I started with creating my branch off '#3200' to leverage the library. I will get rid of that in the commit message as I have a version utils.go that works for these tests.

openshift/pkg/ote/utils/utils.go
asood-rh: I started off with that but later cleaned up and working off main branch. Will fix commit message.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: There are tests that are blocking to start with as the TRT team made a determination on the basis of the pass rate of the test.

If you have concerns we can make then informing.

openshift/pkg/otp/utils/utils.go
asood-rh: I have not as most of the test code is migrated as it is with just the changes using the OTE where needed.

anuragthehatter:
@asood-rh When you get time, can you investigate if your cases are being picked up in `e2e-metal-ipi-ovn-ipv6`  ipv6 job and passing or skipped wherever applicable. Thanks

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: Fixed

openshift/pkg/otp/utils/utils.go
asood-rh: The api that is queried to obtain the metrics does not return the output in JSON format.

openshift/pkg/ote/utils/utils.go
asood-rh: Fixed

openshift/test/ote/metrics.go
asood-rh: The function is called only once. Looking at pros and cons changing implementation does not make sense. Would like to keep it as such.

openshift/test/otp/metrics.go
asood-rh: Fixed

openshift/test/otp/metrics.go
asood-rh: Fixed

openshift/test/otp/metrics.go
asood-rh: Changed to ote package

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: Fixed

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: Fixed

openshift/cmd/ovn-kubernetes-tests-ext/main.go
pperiyasamy: can you fix it in the same commit ? that would be cleaner and easy for review.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -128,14 +147,19 @@ func main() {
...
 
 	informingTests := sets.New(test.InformingTests...)
 	blockingTests := sets.New(test.BlockingTests...)
 
 	specs.Walk(func(spec *extensiontests.ExtensionTestSpec) {
-		for _, label := range getTestExtensionLabels() {
-			spec.Labels.Insert(label)
+		isOTP := strings.Contains(spec.Name, "[OTP]")
+
+		if !isOTP {
```
pperiyasamy, line 159: why don't you need test extension labels "sig-network", "ovn-kubernetes-ote"  for OTP tests ?

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -148,14 +172,24 @@ func main() {
 			spec.Name += " " + annotations
 		}
 
-		// prepend other labels by matching on existing spec labels
-		for _, label := range getPrependLabels(spec.Labels) {
-			spec.Labels.Insert(label)
-		}
+		if isOTP {
+			if spec.Labels.Has("Level0") {
```
pperiyasamy, line 176: why do want to handle prepend label for OTP tests separately ?
The `spec.Name = generatePrependedLabelsStr(spec.Labels) + " " + spec.Name already` already takes care of whatever you are doing it here.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
pperiyasamy: do the fix in the same commit.

openshift/pkg/otp/utils/utils.go
pperiyasamy, line 1: this file has so many util functions and ~5000 lines, are all used by metrics tests ? otherwise please keep only relevant ones which helps for better review

openshift/pkg/otp/utils/utils.go
pperiyasamy: have you explored prom2json https://github.com/prometheus/prom2json#json-format ?

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -4,13 +4,17 @@ import (
...
 	"github.com/ovn-kubernetes/ovn-kubernetes/openshift/test"
 	_ "github.com/ovn-kubernetes/ovn-kubernetes/openshift/test/deploymentconfig"
 	"github.com/ovn-kubernetes/ovn-kubernetes/openshift/test/generated"
 	ocpinfraprovider "github.com/ovn-kubernetes/ovn-kubernetes/openshift/test/infraprovider"
 
 	// import ovn-kubernetes tests
 	_ "github.com/ovn-kubernetes/ovn-kubernetes/test/e2e"
+	// import OTE migrated tests
+	_ "github.com/ovn-kubernetes/ovn-kubernetes/openshift/test/ote"
+	_ "github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/deploymentconfig"
```
pperiyasamy, line 17: what is the need for this ?

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -128,14 +133,19 @@ func main() {
 		if err := initializeTestFramework(os.Getenv("TEST_PROVIDER"), cfg); err != nil {
 			panic(err)
 		}
+		exutil.WithCleanup(func() {})
```
pperiyasamy, line 136: same here.

openshift/test/ote/metrics.go
pperiyasamy: I meant `oteutils.CheckovnkubeMasterNetworkProgrammingetrics` is making one curl call per metric. so it can be optimized with single curl call and retrieve desired metrics from the output.

openshift/pkg/otp/utils/utils.go
asood-rh: This is PR worked simultaneously with other PRs where we are trying to land code. Once first PR is merged, utils.go will have minimal changes.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: We would need distinction between the tests that are migrated versus upstream tests running downstream.

asood-rh: After thought the IDs in test case can be make them distinct, therefore using same labels as other tests.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: Removed, can build and link successfully without it. The downstream blank import already in this file at line 9 (openshift/test/deploymentconfig) has an init() that calls deploymentconfig.Set() with
  the OCP config, making the upstream import redundant

openshift/test/ote/metrics.go
asood-rh: @pperiyasamy historically networking tests could run against other plugins like cilium. There is a team that uses them to validate other plugins. This check was added to ensure these tests are skipped for other plugins.

openshift/test/ote/metrics.go
asood-rh: I understood what you were referring to, using single curl to retrieve multiple metrics. Here is the reasons for not modifying the implementation.  The optimization is valid but the benefit is modest — these are e2e tests, not production code. The round-trip overhead is small relative to overall test duration.
                                             
  Pros of single curl call:                                                                                                                                                                                        
  - Fewer oc exec + curl round-trips to the prometheus pod — less load on the API server
  - Metrics are captured at the same point in time — no skew between values if something changes between calls                                                                                                     
  - Faster test execution                                                                                                                                                                                          
  - Cleaner test code at the call site — one call instead of N

  Cons of single curl call:
  - Changes the function signature — callers would need to pass a list of metric names instead of one, or the function returns a map
  - If one metric is missing from the output, error handling becomes more complex (fail the whole check vs. report which specific metric is absent)
  - The current function is simple and self-contained — easy to understand, each call has a clear pass/fail for one metric
  - Harder to debug failures — "metric X missing" is clearer when each metric has its own poll+assertion than when buried in a batch result

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: 1. Sets testsStarted = true — This flips exutil out of dry-run mode. Without this, utility methods that check this flag may no-op during test execution, since package-level Ginkgo initialization happens before tests actually run.                                                             
  2. Initializes the fixture directory — Creates a temp directory and sets OS_TEST_FIXTURE_DIR so that exutil.FixturePath() calls (used by OTE tests to locate YAML templates and test fixtures) resolve correctly. The defer inside WithCleanup cleans up that directory after tests complete.     

  The empty func() {} is intentional — the OTE binary doesn't need to run anything inside the callback. It just needs the side effects that WithCleanup triggers before invoking the callback.

  This is called inside specs.AddBeforeAll() so it runs once before all test specs execute, which is the correct timing for this initialization.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: Thanks @pperiyasamy  Fixed it.

openshift/pkg/otp/utils/utils.go
asood-rh: The current implementation uses ExtractMetricValue() which does in-Go string parsing (replaced the old bash -c "cat | grep | awk" pipeline per CodeRabbit feedback). For the metrics we check here (simple name
  match → extract value), structured JSON parsing adds complexity without benefit. prom2json is a CLI tool that would need to be available in the pod, and the Go library equivalent
  (github.com/prometheus/common/expfmt) would be a heavier change for simple scalar lookups. Happy to switch if there's a team preference for consistency.

openshift/pkg/otp/utils/utils.go
asood-rh: @pperiyasamy Trimmed the utils.go

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: Fixed it

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -36,6 +39,29 @@ import (
...
+	"Add mechanism to record duration for k8 kinds",
+}
+
+func isOTPBlocking(name string) bool {
+	for _, title := range otpBlockingTests {
+		if strings.Contains(name, title) {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
asood-rh: @coderabbitai resolve it

openshift/pkg/otp/testdata/networking/service-generic-template.yaml
```diff
@@ -0,0 +1,34 @@
...
+parameters:
+- name: SERVICENAME
+- name: NAMESPACE
+- name: internalTrafficPolicy
+- name: externalTrafficPolicy
+- name: ipFamilyPolicy
+- name: PROTOCOL
+- name: SELECTOR
+- name: serviceType
+- name: PORT
```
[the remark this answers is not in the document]
asood-rh: File not part of PR any more

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: @pperiyasamy Ok to resolve this, we will carry test case ID as indicator that it is migrated test.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: Should be really fixed now as commits are cleaner now  with three commits.
go.mod/go.sum 
utils.go
rest of the files.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: Done

openshift/pkg/ote/utils/utils.go
```diff
@@ -0,0 +1,416 @@
...
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	exutil "github.com/openshift/origin/test/extended/util"
+	"k8s.io/apimachinery/pkg/util/wait"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+type EgressIPResource1 struct {
```
pperiyasamy, line 16: any specific reason for using this struct and template to construct and create EgressIP object ? 

can't just use u/s approach to create EIP something like below in every test ?
```
					var egressIPConfig = `apiVersion: k8s.ovn.org/v1
kind: EgressIP
metadata:
    name: ` + egressIPName + `
spec:
    egressIPs:
    - ` + egressIP1.String() + `
    - ` + egressIP2.String() + `
    podSelector:
        matchLabels:
            wants: egress
    namespaceSelector:
        matchLabels:
            name: ` + f.Namespace.Name + `
`

					if err := os.WriteFile(egressIPYaml, []byte(egressIPConfig), 0644); err != nil {
						framework.Failf("Unable to write CRD config to disk: %v", err)
					}
					defer func() {
						if err := os.Remove(egressIPYaml); err != nil {
							framework.Logf("Unable to remove the CRD config from disk: %v", err)
						}
					}()

					framework.Logf("Create the EgressIP configuration")
					e2ekubectl.RunKubectlOrDie("default", "create", "-f", egressIPYaml)
```

openshift/pkg/ote/utils/utils.go
```diff
@@ -0,0 +1,416 @@
...
+	chars := "abcdefghijklmnopqrstuvwxyz0123456789"
+	seed := rand.New(rand.NewSource(time.Now().UnixNano()))
+	buffer := make([]byte, 8)
+	for index := range buffer {
+		buffer[index] = chars[seed.Intn(len(chars))]
+	}
+	return string(buffer)
+}
+
+func CheckPlatform(oc *exutil.CLI) string {
```
pperiyasamy, line 104: The OpenshiftInfraProvider in openshift.go already contains operNetwork, clusterFeatureGate. so it can be made to hold infrastructure object as well, so can these utility methods can be from openshift.go itself ?

openshift/pkg/ote/testdata/networking/metrics/metrics-pod.yaml
pperiyasamy, line 1: why do we use hello pod which uses a fixed image `quay.io/openshifttest` registry ? can't we use `e2epod.NewAgnhostPod` instead ? same for `testpod.yaml`.

anuragthehatter: @pperiyasamy Your concern is valid and thanks for the feedback. Many of our existing tests depend on this image due to the miscellaneous utilities bundled in it. At this point, we don't have the bandwidth to re-engineer 100+ tests to use different image and validate that. However, for new tests going forward, we'll be more careful about selecting upstream pod images wherever applicable.

Additionally, this image has been synced to `quay.io/openshift/community-e2e-images` (https://quay.io/repository/openshift/community-e2e-images), so it will continue to be available long-term.

openshift/pkg/ote/utils/utils.go
pperiyasamy, line 1: As @anuragthehatter mentioned in other comment, I'm not sure if this effort is also considering otp tests to adapt with utils that used by u/s E2Es. That requires some refactoring around your OTP tests. have you also considered that ? have you checked how much time it took to complete each test ?

openshift/pkg/ote/utils/utils.go
pperiyasamy, line 16: why `EgressIPResource1` ? can't it be `EgressIPResource` ? can you add some comment about this struct why it is needed ?

openshift/pkg/ote/utils/utils.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+	if output == "" {
+		return []string{}, nil
+	}
+	return strings.Split(output, " "), nil
+}
+
+func GetFirstWorkerNode(oc *exutil.CLI) (string, error) {
+	nodes, err := GetClusterNodesBy(oc, "worker")
+	if err != nil || len(nodes) == 0 {
```
pperiyasamy, line 412: what if nodes are present, but err is due to something else ?

openshift/pkg/ote/utils/utils.go
```diff
@@ -0,0 +1,416 @@
...
+	// holderIdentity format is "<pod-name>_<leader-election-id>", extract just the pod name
+	if parts := strings.SplitN(leaderPodName, "_", 2); len(parts) > 1 {
+		leaderPodName = parts[0]
+	}
+	e2e.Logf("The leader pod name is %s", leaderPodName)
+	leaderPodIP := GetPodIPv4(oc, "openshift-controller-manager", leaderPodName)
+	return leaderPodIP
+}
+
+func GetOVNKMasterPod(oc *exutil.CLI) string {
```
pperiyasamy, line 388: is `GetOVNKMasterPod` still valid ? this is not valid anymore with IC deployment.

openshift/pkg/ote/utils/utils.go
```diff
@@ -0,0 +1,416 @@
...
+			}
+		}
+	}
+	return ""
+}
+
+func GetOVNMetricsInSpecificContainer(oc *exutil.CLI, containerName string, podName string, url string, metricName string) string {
+	var metricValue string
+	metricsErr := wait.Poll(5*time.Second, 10*time.Second, func() (bool, error) {
+		output, err := oc.AsAdmin().WithoutNamespace().Run("exec").Args("-n", "openshift-ovn-kubernetes", "-c", containerName, podName, "--", "curl", url).Output()
```
pperiyasamy, line 239: why does it invoke `WithoutNamespace()`, but in the args it passes `-n `parameter ?

pperiyasamy, line 239: ovnk namespace can be retrieved from deploymentconfig API

openshift/pkg/ote/utils/utils.go
```diff
@@ -0,0 +1,416 @@
...
+		} else {
+			e2e.Logf("Can't get metrics for %s:", metrics)
+			return false, nil
+		}
+		return true, nil
+	})
+	o.Expect(metricsErr).NotTo(o.HaveOccurred(), fmt.Sprintf("Fail to get metric and the error is:%s", metricsErr))
+}
+
+func GetSvcIP(oc *exutil.CLI, namespace string, svcName string) (string, string) {
```
pperiyasamy, line 286: I'm bit worried about timings with usage of multiple oc commands just to retrieve svc IP from a service, can't we just use kube API ?

openshift/cmd/ovn-kubernetes-tests-ext/main.go
pperiyasamy: ok, can you move this into `initializeTestFramework` with a comment ?

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -153,7 +158,7 @@ func main() {
 			spec.Labels.Insert(label)
 		}
 
-		spec.Name = generatePrependedLabelsStr(spec.Labels) + " " + spec.Name // prepend ginkgo labels to test name
+		spec.Name = generatePrependedLabelsStr(spec.Labels) + " " + spec.Name
```
pperiyasamy, line 160: why the comment is removed ?

openshift/test/ote/metrics.go
```diff
@@ -0,0 +1,448 @@
...
+	oteutils "github.com/ovn-kubernetes/ovn-kubernetes/openshift/pkg/ote/utils"
+	"github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/deploymentconfig"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
+	e2eoutput "k8s.io/kubernetes/test/e2e/framework/pod/output"
+)
+
+var _ = g.Describe("OVN metrics", g.Label("JIRA:Networking", "Feature:Metrics"), func() {
```
pperiyasamy, line 27: why the label `JIRA:Networking` ?

openshift/test/ote/metrics.go
pperiyasamy: but ote tests in ovnk is not meant to run on cilium

openshift/test/ote/metrics.go
```diff
@@ -0,0 +1,448 @@
...
+			"ovn_controller_pinctrl_run_total_samples",
+			"ovn_controller_pinctrl_run_long_term_avg",
+			"ovn_controller_ofctrl_seqno_run_total_samples",
+			"ovn_controller_ofctrl_seqno_run_long_term_avg",
+			"ovn_controller_patch_run_total_samples",
+			"ovn_controller_patch_run_long_term_avg",
+			"ovn_controller_ct_zone_commit_total_samples",
+			"ovn_controller_ct_zone_commit_long_term_avg",
+		}
+		for _, m := range metrics {
```
pperiyasamy, line 157: are these tests not intended to be refactored to improve timings ? there is a curl for every metric, i think all metrics are retrieved with one curl call and parse the result for desired metric, as i mentioned earlier prom2json library would be an appropriate for this.

openshift/pkg/ote/utils/utils.go
asood-rh: @pperiyasamy The effort is to migrate the  openshift-tests-private tests to use openshift-test-extension. It is not objective to adapt it to use upstream test pattern. It has been determined one of the test is slow therefore it has label, others are within the limit.

openshift/pkg/ote/utils/utils.go
asood-rh: @pperiyasamy The objective is not t refactor the tests to use upstream test pattern.

openshift/test/ote/metrics.go
asood-rh: It would not make sense to introduce additional dependency for one test. Retrieving one metric with a curl keeps the parsing of output simple.

openshift/test/ote/metrics.go
asood-rh: Its purpose is to declare test ownership: it maps the test to the Networking component in the OCPBUGS Jira project
• Sippy and Component Readiness can attribute failures and regressions to the right component for reporting and triage.
• Without the annotation, the test can still execute and be imported into Sippy, but it will not be correctly mapped in Component Readiness dashboards.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
asood-rh: Added it back

openshift/pkg/ote/utils/utils.go
asood-rh: It finds the ovn-K cluster manager master pod. Added comment in the code for it.

openshift/pkg/ote/utils/utils.go
asood-rh: It can be but that is how it was named in original test code.  It is used for creating egress IP object using template YAML. It is passed to constructor and destructor functions. This is how it is done for other objects.

openshift/pkg/ote/utils/utils.go
asood-rh: Improved it so that failure to get nodes and no nodes error do not mask.

openshift/pkg/ote/utils/utils.go
asood-rh: As the intent is to migrate tests and not refactor them as upstream tests, this could continue to use existing function.

openshift/pkg/ote/utils/utils.go
asood-rh: WithoutNamespace() disables the CLI wrapper’s default --namespace=<e2e project>. -n in Args is the real namespace for that oc call. It is not redundant.

asood-rh: Done

PR #3226, opened 2026-06-04
OCPBUGS-87020: fix non-atomic NorthDB port-group writes causing persistent NodePort traffic loss

joshbranham:
## 📑 Description
When ovnkube-controller reconnects to kube-apiserver after a transient outage (e.g. KAS revision rollout) and re-syncs NetworkPolicy state, two bugs combine to cause persistent NodePort/LB traffic loss:

1. CreateOrUpdatePortGroupsOps clears existing port membership during Update. getAllUpdatableFields() unconditionally returns Ports in the OnModelUpdates list. When BuildPortGroup() is called with nil ports (the normal path in createNetworkPolicy and createDefaultDenyPGAndACLs), and the port group already exists in NorthDB (during retry/re-sync), the OVSDB Update operation replaces existing ports with an empty set. The allow ACLs on the now-empty port group match no pods, while default-deny ACLs still apply — dropping all traffic.

2. The local pod sync function silently discards errors. In addLocalPodHandler, the syncFunc ignores the return value of handleLocalPodSelectorAddFunc. If the NorthDB transaction fails during sync (e.g. connection instability during API server recovery), ports are never added to the port group and no retry is triggered. Since resyncInterval=0, the stale state persists indefinitely.

Fix getAllUpdatableFields for PortGroup to skip nil fields — nil means "not specified" (preserve existing value), while non-nil means "set to this value". This is safe for all callers: NetworkPolicy code passes nil ports (preserved on update), ANP code passes explicit ports (replaced as before), and the Create path is unaffected (uses the full model, not OnModelUpdates).

Fix the syncFunc to return errors from handleLocalPodSelectorAddFunc. The function is idempotent (pods already in np.localPods are skipped), and the WatchFactory retries the sync for up to 60 seconds.

Both bugs were introduced in 2022 (commits 5d84deb81 and c973d615c) and are present in all release branches from 4.14 through 4.21.

Fixes: https://issues.redhat.com/browse/OCPBUGS-87020

## ✅ Checks

- [x] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Pod sync now surfaces sync errors so failed syncs are retried.
  * Port group updates now preserve existing ports when ports are unspecified; explicitly provided ports replace existing ones; creating with unspecified ports yields an empty ports list. ACLs and external IDs are consistently applied.

* **Tests**
  * Added tests covering port group create/update behaviors for nil, explicit, and create-with-nil ports.

### DIFF
--- a/go-controller/pkg/libovsdb/ops/model.go
+++ b/go-controller/pkg/libovsdb/ops/model.go
@@ -597,7 +597,16 @@ func getAllUpdatableFields(model model.Model) []interface{} {
 	case *nbdb.LogicalSwitchPort:
 		return []interface{}{&t.Addresses, &t.Type, &t.TagRequest, &t.Options, &t.PortSecurity}
 	case *nbdb.PortGroup:
-		return []interface{}{&t.ACLs, &t.Ports, &t.ExternalIDs}
+		fields := []interface{}{&t.ACLs, &t.ExternalIDs}
+		// Only include Ports when explicitly set (non-nil). Callers like
+		// createNetworkPolicy pass nil Ports via BuildPortGroup and rely on
+		// a separate local-pod handler to populate them. Including a nil
+		// Ports here would generate an OVSDB Update that clears existing
+		// port membership, breaking traffic until pods are re-added.
+		if t.Ports != nil {
+			fields = append(fields, &t.Ports)
+		}
+		return fields
 	default:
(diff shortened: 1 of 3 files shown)


joshbranham:
/jira refresh
@coderabbitai review

joshbranham:
@coderabbitai review

PR #3227, opened 2026-06-05
CORENET-6746: Enable EgressIP E2E tests on baremetal, AWS, Azure and GCP for OTE

pperiyasamy:
Enable EgressIP E2E tests on OpenShift CI across multiple platforms.

**Baremetal**: Add secondary network support for EgressIP on secondary host interfaces (secondary-host-eip tests), including firewall configuration, IP forwarding, and host-networked container caching.
**AWS/Azure/GCP**: Add bastion-based platform support using a shared infrastructure pattern (SSH + podman on bastion host).
**E2E fixes**: Reserve primary network IP range to avoid node IP conflicts, fix namespace label update races, abstract NBDB container name, and bind explicit UDP ports for agnhost containers (E2E fixes are candidate for a upstream PR, https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6783 will be used for this).


openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -128,6 +132,13 @@ func main() {
 		if err := initializeTestFramework(os.Getenv("TEST_PROVIDER"), cfg); err != nil {
 			panic(err)
 		}
+		client, err := kclientset.NewForConfig(cfg)
+		if err != nil {
+			panic(fmt.Sprintf("failed to create k8s clientset: %v", err))
+		}
+		if err := ipalloc.InitPrimaryIPAllocator(client.CoreV1().Nodes()); err != nil {
+			panic(fmt.Sprintf("failed to initialize node primary IP allocator: %v", err))
+		}
```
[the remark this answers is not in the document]
pperiyasamy: this is not correct find, `specs.AddBeforeAll` runs only once for all the tests and does the same thing as upstream `InitPrimaryIPAllocator` is called suite-wide in [e2e_suite_test.go:58](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/test/e2e/e2e_suite_test.go#L58) inside `ginkgo.BeforeSuite`.

openshift/test/infraprovider/base.go
```diff
@@ -0,0 +1,252 @@
...
+func (p *baseContextProvider) DeleteNetwork(network api.Network) error {
+	return p.engine.DeleteNetwork(network)
+}
+
+func (p *baseContextProvider) AttachNetwork(network api.Network, instance string) (api.NetworkInterface, error) {
+	if _, err := p.parent.runner.Run(containerRuntime, "container", "inspect", instance); err != nil {
+		return api.NetworkInterface{}, nil
+	}
+	return p.engine.AttachNetwork(network, instance)
+}
```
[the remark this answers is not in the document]
pperiyasamy: this is fixed now with using `p.engine.GetContainerState` function.

pperiyasamy:
@coderabbitai addressed most of your Nitpick comments, PTAL

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/e2e-aws-core-networking-serial-ote openshift/origin#31614

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/e2e-aws-core-networking-serial-ote https://github.com/openshift/origin/pull/31614

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/e2e-aws-core-networking-serial-ote https://github.com/openshift/origin/pull/31614

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/e2e-aws-core-networking-serial-ote https://github.com/openshift/origin/pull/31614

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -64,7 +69,28 @@ func shouldIncludeTest(spec *extensiontests.ExtensionTestSpec) bool {
...
 
-	// Future feature-based filters can be added here
+	// If platform infra node (hypervisor node for baremetal, bastion host
+	// for cloud platforms) doesn't exist, then ignore running EgressIP tests.
+	if !ocpInfra.HasPlatformInfra() && spec.Labels.Has(featureLabelEgressIP) {
+		return false
+	}
+	// secondary-host-eip tests: only include on platforms with a
+	// pre-configured secondary network (currently baremetal only)
+	if strings.Contains(spec.Name, "secondary-host-eip") && !ocpInfra.HasSecondaryHostEIPSupport() {
```
arkadeepsen, line 79: Will it be better to use a constant for `secondary-host-eip` which can be referred everywhere?

openshift/test/infraprovider/openshift.go
```diff
@@ -73,30 +84,78 @@ func (o *OpenshiftInfraProvider) initClusterObjects(config *rest.Config) error {
...
+		o.operNetwork.Spec.DefaultNetwork.OVNKubernetesConfig.GatewayConfig != nil &&
 		o.operNetwork.Spec.DefaultNetwork.OVNKubernetesConfig.GatewayConfig.RoutingViaHost {
 		// The E2E utility method isLocalGWModeEnabled depends on the
 		// OVN_GATEWAY_MODE environment variable. All EVPN tests must
 		// satisfy this condition; otherwise, they will be skipped.
 		_ = os.Setenv("OVN_GATEWAY_MODE", "local")
 	}
+	// OpenShift always supports network segmentation (UDN). The EgressIP
+	// E2Es check this env variable to decide whether to run UDN tests.
+	_ = os.Setenv("ENABLE_NETWORK_SEGMENTATION", "true")
```
arkadeepsen, line 127: I think this is a bug in CNO code. There was a PR opened for fixing it but it got closed: https://github.com/openshift/cluster-network-operator/pull/2739

Let's enable `ENABLE_NETWORK_SEGMENTATION` in the tests correctly as once the CNO bug is fixed this will not work.

openshift/test/infraprovider/openshift.go
```diff
@@ -73,30 +84,78 @@ func (o *OpenshiftInfraProvider) initClusterObjects(config *rest.Config) error {
...
+	return o.platformType == configv1.GCPPlatformType
+}
+
+// IsBastionBasedPlatform returns true for cloud platforms (AWS, GCP, Azure)
+// where external containers run on a bastion host with host networking.
+func (o *OpenshiftInfraProvider) IsBastionBasedPlatform() bool {
+	return o.platformType == configv1.AWSPlatformType ||
+		o.platformType == configv1.AzurePlatformType ||
+		o.platformType == configv1.GCPPlatformType
 }
```
arkadeepsen, line 159: Since here the bastion host part is not separately checked will it make sense to rename the function as `IsCloudPlatform`?

openshift/test/infraprovider/openshift.go
```diff
@@ -318,3 +377,34 @@ func (o *contextOpenshift) DeleteNetwork(network api.Network) error {
...
+	platformType := infra.Spec.PlatformSpec.Type
+	switch platformType {
+	case configv1.BareMetalPlatformType:
+		bm, err := initializeClusterInfra(infra)
+		if err != nil || bm == nil {
+			return nil, platformType, err
+		}
+		return bm, platformType, nil
+	case configv1.AWSPlatformType, configv1.AzurePlatformType, configv1.GCPPlatformType:
+		bi, err := initializeBastionInfra()
```
arkadeepsen, line 402: The naming of the functions `initializeClusterInfra` and `initializeBastionInfra` is a little confusing. Do you think these can be replaced with `initializeBaremetalInfra` and `intializeCloudInfra`?

openshift/test/infraprovider/baremetal.go
```diff
@@ -27,19 +21,27 @@ const (
...
-	primaryNetworkName         = "ostestbm_net"
+	bmPrimaryNetworkName       = "ostestbm_net"
 	frrContainerPrimaryNetIPv4 = "192.168.111.3"
 	frrContainerPrimaryNetIPv6 = "fd2e:6f44:5dd8:c956::3"
 	externalFRRContainerName   = "frr"
+
+	bmSecondaryNetworkName = "secondarynetwork"
+	// Secondary subnets must match the values in test/e2e/egressip.go
+	// (secondaryIPV4Subnet / secondaryIPV6Subnet) because EgressIP tests
+	// hardcode addresses from these ranges. The CI step that provisions
```
arkadeepsen, line 32: Instead of hardcoding can these be made configurable so that the EgressIPs are allocated from the provided CIDR ranges? Otherwise any u/s change of these EgressIPs can break our d/s CI.

openshift/test/infraprovider/baremetal.go
```diff
@@ -49,155 +51,227 @@ func initializeClusterInfra(config *rest.Config) (*baremetalInfra, error) {
...
+	bm := &baremetalInfra{base: h}
+
+	// Discover secondary network interface on hypervisor (optional).
+	secondaryNetInfo, err := findHypervisorNodeInterface(sshRunner, bmSecondaryIPv4Subnet, bmSecondaryIPv6Subnet)
+	if err == nil && secondaryNetInfo != nil {
+		secondaryNet := &network.ContainerEngineNetwork{NetName: bmSecondaryNetworkName}
+		secondaryNet.Configs = []network.ContainerEngineNetworkConfig{
+			{Subnet: bmSecondaryIPv4Subnet},
+			{Subnet: bmSecondaryIPv6Subnet},
+		}
```
arkadeepsen, line 100: nit:
```suggestion
		secondaryNet := &network.ContainerEngineNetwork{
			NetName: bmSecondaryNetworkName,
			Configs: []network.ContainerEngineNetworkConfig{
				{Subnet: bmSecondaryIPv4Subnet},
				{Subnet: bmSecondaryIPv6Subnet},
			}
		}
```

arkadeepsen (commented):
I'll do another round of review.

Can we add some unit tests for the code? A Makefile with the test target is already available. I am planning to add it to the unit CI job.

pperiyasamy:
@arkadeepsen Had to update the PR for the latest u/s and as well as for [PR](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6783) changes, will address your comments in the next PR update.

openshift/test/infraprovider/openshift.go
pperiyasamy: The `ENABLE_NETWORK_SEGMENTATION` is a test env which is not relevant to the CNO fix, The EgressIP E2Es expects this env to be "true" to run net-seg related tests, so ote is exporting this variable on the test container. we can get rid of env variables once existing E2Es starting to adapt with deploymentconfig's `IsConfigurationEnabled` API.

openshift/test/infraprovider/baremetal.go
pperiyasamy: The u/s EIP E2E is using below subnets for EIP secondary interface and hardcoded IP address (from this subnet for EIP).
```
	secondaryIPV4Subnet                        = "10.10.10.0/24"
	secondaryIPV6Subnet                        = "2001:db8:abcd:1234::/64"
```
so I'm just sticking to this subnet in ote as well by precreating secondary network on this subnet via https://github.com/openshift/release/pull/83781. I think we can just have this as an initial solution before generalizing subnets in u/s and d/s.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
pperiyasamy: this constant is used only here.

openshift/test/infraprovider/openshift.go
pperiyasamy: done.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
arkadeepsen: For now it's only being used here. If in future it's needed to be used somewhere else then it's be better to declare a constant for that. This also helps to avoid unintended typos.

BTW `featureLabelEVPN` is used only used once which is also a constant.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -86,22 +117,23 @@ func main() {
 	// No Parents: these tests run only in ovn-kubernetes/conformance/*, not the product-wide openshift/conformance/*.
 	// To inject a subset later, label those tests and add a suite with Parents=[openshift/conformance/parallel] + a matching qualifier.
 	ovnTestsExtension.AddSuite(extension.Suite{
-		Name:       "ovn-kubernetes/conformance/serial",
-		Qualifiers: []string{`labels.exists(l, l == "Serial")`},
+		Name:        "ovn-kubernetes/conformance/serial",
+		Parallelism: 1,
+		Qualifiers: []string{fmt.Sprintf(`labels.exists(l, l == "Serial") || labels.exists(l, l == "%s")`,
+			featureLabelEgressIP)},
```
arkadeepsen: Why is a separate condition for egress IP needed? The EIP tests are being added the `[Serial]` label here: https://github.com/openshift/ovn-kubernetes/pull/3227/changes#diff-a392bfc12a449dc2c3fe79ac9957b78d1b5721ccdc3c00c85cfb17c4ded66fc6R123-R125

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -86,22 +117,23 @@ func main() {
...
 	})
 
 	ovnTestsExtension.AddSuite(extension.Suite{
-		Name:       "ovn-kubernetes/conformance/parallel",
-		Qualifiers: []string{`!labels.exists(l, l == "Serial")`},
+		Name: "ovn-kubernetes/conformance/parallel",
+		Qualifiers: []string{fmt.Sprintf(`!labels.exists(l, l == "Serial") && !labels.exists(l, l == "%s")`,
+			featureLabelEgressIP)},
```
arkadeepsen: Same as above.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -64,7 +69,33 @@ func shouldIncludeTest(spec *extensiontests.ExtensionTestSpec) bool {
...
+	if !ocpInfra.HasPlatformInfra() && spec.Labels.Has(featureLabelEgressIP) {
+		return false
+	}
+	// secondary-host-eip tests: only include on platforms with a
+	// pre-configured secondary network (currently baremetal only)
+	if strings.Contains(spec.Name, "secondary-host-eip") && !ocpInfra.HasSecondaryHostEIPSupport() {
+		return false
+	}
+	// EgressIP host-networked pods tests require opening ports (30000 to 32767) on cluster nodes,
+	// which is not yet configured on bastion-based platforms.
```
arkadeepsen, line 85: What is neded for these tests to be enabled?

openshift/test/infraprovider/baremetal.go
arkadeepsen: For now we can go ahead with this solution, but it does seem fragile. Can we have a tracker for this so that it's not forgotten?

openshift/test/infraprovider/openshift.go
arkadeepsen: I didn't mean the test specific env var. The comment mentions `OpenShift always supports network segmentation (UDN)` which might not be correct as it's a bug in CNO. So, in future if this bug is fixed, but the test env var is set to true, it may end up failing if the `DisableMultiNetwork` on the `network.operator.openshift.io` CR is set to true.

openshift/test/infraprovider/base.go
```diff
@@ -0,0 +1,492 @@
+package infraprovider
```
arkadeepsen, line 1: I think there should be unit test coverage for this package, specifically for `base.go`, `baremetal.go` and `openshift.go` file.

openshift/test/infraprovider/README.md
```diff
@@ -0,0 +1,293 @@
...
+| Hypervisor                 |                      | Cluster Nodes          |
+| (SSH: root)                |                      |                        |
+|                            |  ostestbm_net        | - worker-0             |
+| Containers:                |  192.168.111.0/24    | - worker-1             |
+|                            |  (machine network)   | - worker-2             |
+| - EgressIP targets         |                      |                        |
+|   (podman)                 |  secondarynetwork    |                        |
+|                            |  10.10.10.0/24       | (secondary intf)       |
+| - EgressIP targets         |  2001:db8:abcd::/64  |                        |
+|   (podman)                 |                      |                        |
```
arkadeepsen: >- EgressIP targets
>  (podman)
> 
>- EgressIP targets
>  (podman)

nit: I think this is incorrectly repeated twice.

openshift/test/infraprovider/README.md
```diff
@@ -0,0 +1,293 @@
...
++----------------------------------------------+------+------+-------+------+
+```
+
+**Footnotes**:
+
+- **[1]** Skipped for UDN network types only (has `Feature:NetworkSegmentation`
+  label). Cluster Default variants run on GCP.
+- **[2]** Skipped on all cloud platforms. Host-networked pod tests require
+  opening ports 30000-32767 on cluster nodes, not configured on bastion-based
+  platforms.
```
arkadeepsen: It'll be good to mention if this can be enabled in the future or if this is not at all supported.

openshift/test/infraprovider/README.md
pperiyasamy: The EgressIP E2Es spins upto 3 external containers, so it tried to depict that, fixed the text with numbering.

openshift/test/infraprovider/README.md
pperiyasamy: done

openshift/test/infraprovider/base.go
pperiyasamy: added relevant unit tests now.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
pperiyasamy: it needs node port firewall rules for range 30000-32767 in the ocp nodes like it's done [here](https://github.com/openshift/release/pull/84557) for bastion host.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
pperiyasamy: previously it needed either ginkgo.Serial label in the EIP tests or this logic (`labels.exists(l, l == "%s")`, featureLabelEgressIP` ) to run EIP tests in the serial lane.
I don't remember it exactly when this problem happens, removed the logic now and let us see how tests run in serial lane.

openshift/test/infraprovider/baremetal.go
pperiyasamy: this is an underlay configuration for secondary network, I don't see a reason if this config would get changed in u/s anytime sooner, those tests are living there for years. even if that changes, we could adpat ote for those changes.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
pperiyasamy: now i see the problem, it happens at the run time with below error:

```
Suite run returned error: no tests to run
error: error running a test suite: no tests to run
```

This is because qualifiers looks for actual ginkgo Label. The `test/annotate/rules.go` just adds `[Serial]` string into test name which is not a label, so can't be used in Qualifier CEL expressions. Hence it needed to match on `featureLabelEgressIP` as well.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
arkadeepsen: Aren't node port ranges already allowed for the cloud platforms during deployment? How would node port work if those aren't allowed?

If it's indeed the case, then can this be tracked by a Jira story?

openshift/cmd/ovn-kubernetes-tests-ext/main.go
arkadeepsen: Should it be like how origin repo is doing it?
https://github.com/openshift/origin/blob/0735d46a135e8c6c4f4332b140230505f591887b/pkg/testsuites/standard_suites.go#L117-L141

PR #3237, opened 2026-06-08
NO-JIRA: DownStream Merge [06-08-2026]

arkadeepsen:
Fixed following merge conflicts:

```shell
CONFLICT (content): Merge conflict in go-controller/pkg/ovn/master_test.go
```

PR #3239, opened 2026-06-08
Fix primary UDN race condition in CNI server

jluhrsen:
## Summary

This PR fixes a critical race condition in primary UDN (User Defined Network) pod creation that was introduced in upstream commit 1826e9971.

## The Problem

All NetworkSegmentation e2e tests were failing with:
```
timed out waiting for OVS port binding (ovn-installed) for <MAC> [<IP>]
```

## Root Cause

Commit 1826e9971 ("Move primary UDN out of pod request") refactored primary UDN handling but inadvertently removed the annotation waiting logic. This created a race condition where the CNI tries to configure the OVS port before the OVN controller has created the logical port.

## The Fix

Add a single call to `GetPodWithAnnotations(isOvnReady)` to wait for the OVN controller to be ready before calling `cmdAdd()` on the primary UDN interface. This is a minimal 7-line fix that restores the original behavior.

## Testing

- Fixes 64 failing NetworkSegmentation tests in PR #3237
- Should be verified with: `/test e2e-aws-ovn`

## Related

- Fixes regression from upstream ovn-kubernetes/ovn-kubernetes commit 1826e9971
- Related to PR #3237 (downstream merge)

### DIFF
--- a/go-controller/pkg/cni/cniserver.go
+++ b/go-controller/pkg/cni/cniserver.go
@@ -333,6 +333,17 @@ func (s *Server) handleCNIRequest(r *http.Request) (result []byte, err error) {
 		}
 		if primaryPodRequest != nil {
 			klog.V(4).Infof("Pod %s/%s primaryUDN podRequest %v", primaryPodRequest.PodNamespace, primaryPodRequest.PodName, primaryPodRequest)
+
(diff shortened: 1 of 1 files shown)


jluhrsen:
Closing this PR. The fix should be added directly to PR #3237 instead of creating a separate PR.

PR #3240, opened 2026-06-08
[Depends on #3237] Fix primary UDN race condition

jluhrsen:
## Stack Dependencies
**Depends on**: #3237

## Summary
This PR adds a fix for the NetworkSegmentation test failures introduced in the downstream merge PR #3237.

## The Problem
PR #3237 merged upstream commit 1826e9971 which introduced a race condition in primary UDN handling. This causes all NetworkSegmentation tests to fail with:
```
timed out waiting for OVS port binding (ovn-installed) for <MAC> [<IP>]
```

## The Fix
7-line fix that adds annotation waiting before calling `cmdAdd()` for primary UDN interfaces. This restores the synchronization that was accidentally removed during the upstream refactoring.

## Merge Order
1. Merge #3237 first (downstream merge)
2. Then merge this PR (fix for the regression)

OR squash both when merging #3237.

## Testing
- Fixes all 64 failing NetworkSegmentation tests
- Should be verified with: `/test e2e-aws-ovn`


jluhrsen:
Closing - this fix addresses a symptom (CNI timeout) but not the root cause (network controllers failing to start due to watchResource timeouts).

PR #3257, opened 2026-06-17
OCPBUGS-74711: [release-4.20] EgressIP annotation corruption causes stale IPs on br-ex after failover

bpickard22:
## Summary

Backport of upstream [ovn-kubernetes/ovn-kubernetes#5951](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5951) to release-4.20.

Cherry-picked commits:
- `0e44890df` — EgressIP: Fix crash from mutating node informer object
- `b95fc8081` — Fixes gateway egress IP node update logic

## Problem

`addIPToAnnotation` and `deleteIPsFromAnnotation` in `gateway_egressip.go` read the node annotation from the informer cache and directly mutate the cached node object. With many EgressIPs (customer has 270), the rapid read-modify-write calls during `SyncEgressIP` overwrite each other due to stale informer reads. The annotation ends up tracking only ~7% of assigned IPs, preventing `SyncEgressIP` from cleaning up stale IPs on br-ex after failover. This causes duplicate IPs across nodes and service outages.

## Fix

1. Deep-copy the node object before mutation (stops corrupting the informer cache)
2. Replace informer reads with a local `annotationIPs` cache initialized at startup, eliminating the stale-read race

## Test plan

- [ ] Existing EgressIP unit tests pass
- [ ] CI green
- [ ] Verified fix is in 4.21+ and working

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-74711

### DIFF
--- a/go-controller/pkg/node/controllers/egressip/egressip.go
+++ b/go-controller/pkg/node/controllers/egressip/egressip.go
@@ -1142,8 +1142,12 @@ func (c *Controller) migrateFromAddrLabelToAnnotation() error {
 		if err != nil {
 			return err
 		}
-		node.Annotations[util.OVNNodeSecondaryHostEgressIPs] = string(patch)
-		return c.kube.UpdateNodeStatus(node)
+		nodeToUpdate := node.DeepCopy()
+		if nodeToUpdate.Annotations == nil {
+			nodeToUpdate.Annotations = map[string]string{}
+		}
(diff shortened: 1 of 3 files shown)


bpickard22:
backport tracking is a little wonky here, this fix from @trozet made it downstream via https://github.com/openshift/ovn-kubernetes/pull/3011 which is a no jira, so we will have to create a chain up to main to satisfy the bot, this is also why I am pointed to the upstream pr instead of something downstream as I think it provides better clarity

PR #3261, opened 2026-06-23
OCPBUGS-86018: Fix NodePort cleanup when allocateLoadBalancerNodePorts is disabled

ssonigra:
## Summary

When a LoadBalancer service has `allocateLoadBalancerNodePorts` set to false, OVN-Kubernetes was still creating NodePort load balancers and keeping NodePort listeners open on nodes.

## Root Cause

Two issues prevented proper cleanup:

1. **lb_config.go**: `buildServiceLBConfigs()` only checked if `NodePort != 0`, but Kubernetes keeps this value internally even after removal from the spec, so the check always passed.

2. **port_claim.go**: `UpdateService()` didn't detect when `allocateLoadBalancerNodePorts` changed, so it never triggered cleanup of existing NodePort listeners.

### Fix 1: lb_config.go (buildServiceLBConfigs)
Changed the logic to check service type and allocation policy instead of relying on the NodePort value:
- **NodePort services**: Always create NodePort LBs
- **LoadBalancer services**: Only create NodePort LBs if `allocateLoadBalancerNodePorts` is enabled (true or nil)

### Fix 2: port_claim.go (UpdateService)
Added detection for when `allocateLoadBalancerNodePorts` changes from true to false, triggering proper cleanup of NodePort listeners.

## Files Changed

- `go-controller/pkg/ovn/controller/services/lb_config.go` - Service LB config logic
- `go-controller/pkg/node/port_claim.go` - NodePort listener lifecycle management
- `go-controller/pkg/ovn/controller/services/lb_config_test.go` - Comprehensive test coverage

## Test Plan

- [x] LoadBalancer with `allocateLoadBalancerNodePorts=false` → no NodePort LB created
- [x] LoadBalancer with `allocateLoadBalancerNodePorts=true` → NodePort LB created  
- [x] NodePort service → NodePort LB always created (regardless of allocate field)
- [x] Transitioning from true→false closes existing NodePort listeners

## Verification Steps

Tested on OpenShift 4.18:
1. Create LoadBalancer service (NodePort auto-assigned, listener active on nodes)
2. Set `allocateLoadBalancerNodePorts: false` and remove `nodePort` from spec
3. **Result**: NodePort listener closes within seconds ✅

Before fix: Port remained open indefinitely (security issue)
After fix: Port closes immediately upon reconciliation

## Fixes

OCPBUGS-86018

---

## Summary by CodeRabbit

* **Bug Fixes**
  * NodePort load balancers now respect whether LoadBalancer services are configured to allocate NodePorts, preventing unnecessary NodePort exposure.
  * Services that change NodePort allocation settings now update port handling correctly, so stale port claims are avoided.
  * LoadBalancer services continue to get the expected cluster-wide load balancer behavior even when NodePort allocation is disabled.
* **Tests**
  * Expanded coverage for NodePort and LoadBalancer service combinations.

### DIFF
--- a/go-controller/pkg/node/port_claim.go
+++ b/go-controller/pkg/node/port_claim.go
@@ -160,7 +160,16 @@ func (p *portClaimWatcher) AddService(svc *corev1.Service) error {
 }
 
 func (p *portClaimWatcher) UpdateService(old, new *corev1.Service) error {
-	if reflect.DeepEqual(old.Spec.ExternalIPs, new.Spec.ExternalIPs) && reflect.DeepEqual(old.Spec.Ports, new.Spec.Ports) {
+	// Check if anything changed that would affect port claims:
+	// - ExternalIPs changed
+	// - Ports changed
+	// - AllocateLoadBalancerNodePorts changed (affects whether NodePorts should be claimed)
+	oldAllocate := util.LoadBalancerServiceHasNodePortAllocation(old)
+	newAllocate := util.LoadBalancerServiceHasNodePortAllocation(new)
+
+	if reflect.DeepEqual(old.Spec.ExternalIPs, new.Spec.ExternalIPs) &&
+	   reflect.DeepEqual(old.Spec.Ports, new.Spec.Ports) &&
+	   oldAllocate == newAllocate {
 		return nil
 	}
 	var errors, raw_errors []error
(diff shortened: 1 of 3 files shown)


ssonigra:
Closing this a upstream PR is raised

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6589

PR #3267, opened 2026-06-25
test/e2e: fix UDN services test for cloud platforms

jechen0648:
The **[Feature:NetworkSegmentation] Network Segmentation: services … should be reachable through their cluster IP, node port and load balancer** test was failing on dualstack AWS clusters due to three cloud-platform incompatibilities. This PR fixes all three.

Problem 1 — UDP LoadBalancer not supported on AWS

Cloud load balancers (e.g. AWS ELB) do not support UDP protocol. The test unconditionally created a UDP LoadBalancer service and called WaitForLoadBalancer, which timed out after 3 minutes on AWS.

Fix: Gate WaitForLoadBalancer and all checkConnectionToLoadBalancers calls behind isKindCluster (infraprovider.IsKind()), since UDP LoadBalancer is only available in KinD environments.

Problem 2 — External container (frr) not available on cloud clusters

The OpenShift infra provider's ExecExternalContainerCommand panics with "not implemented" when no baremetal cluster infrastructure is present (i.e. clusterInfra == nil). The test called checkConnectionToNodePortFromExternalContainer unconditionally, causing a panic.

Fix: Wrap the entire external container section — both LB and NodePort checks — inside the same isKindCluster guard.

Problem 3 — PodSecurity admission rejects pods in manually-created namespace

The test creates a secondary namespace (<ns>-default) using the raw API, without the pod-security.kubernetes.io/enforce=privileged labels that f.CreateNamespace sets automatically (via f.NamespacePodSecurityLevel = admissionapi.LevelPrivileged). On OpenShift AWS clusters, the default enforcement level is restricted, so pods without an explicit security context are rejected.

Fix: Add pod-security.kubernetes.io/enforce/warn/audit=privileged labels to the manually-created namespace at creation time, matching what the framework does for test namespaces.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Tests**
  * Adjusted network segmentation end-to-end checks to run LoadBalancer-related assertions only in supported KinD environments.
  * Kept NodePort connectivity validation active while skipping unsupported LoadBalancer checks on other cluster types.
  * Updated the default test namespace labels to align with the expected Pod Security settings for negative connectivity validation.

### DIFF
--- a/test/e2e/network_segmentation_services.go
+++ b/test/e2e/network_segmentation_services.go
@@ -30,6 +30,7 @@ import (
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
+	admissionapi "k8s.io/pod-security-admission/api"
 	utilnet "k8s.io/utils/net"
 )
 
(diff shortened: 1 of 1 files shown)


jechen0648:
UDN service test failed on all OCP platforms, it should be fixed in ovn-kubernetes-openshift/openshift/test/infraprovider/openshift.go  instread of test/e2e/network_segmentation_service.go

close this PR

PR #3269, opened 2026-06-26
[DNM]OCPBUGS-87249 - clustermanager: re-evaluate egress IPs when cloud egress IP config changes

jechen0648:
## 📑 Description

Problem
On dualstack AWS clusters, egressIPv6 was never assigned even though nodes correctly report IPv6 subnet and capacity in the cloud.network.openshift.io/egress-ipconfig annotation.

The underlying cause lives in two places. The primary cause is in cloud-network-config-controller (CNCC): a bug in getSubnet could cause it to initially write the annotation with no IPv6 subnet (or the wrong one) by unconditionally picking the first entry in Ipv6CidrBlockAssociationSet regardless of its state. This is fixed in a companion CNCC [PR](https://github.com/openshift/cloud-network-config-controller/pull/228).

However, even with the CNCC fix in place, there is a secondary failure mode in OVN-K's node UpdateResource handler. Consider this sequence:

A node comes up; CNCC has not yet annotated it, so initEgressIPAllocator fails. The node's egressIPConfig in the allocator cache has V6.Net == nil.
CNCC annotates the node (possibly with no IPv6 subnet initially, due to the CNCC bug).
OVN-K receives the node UPDATE, calls initEgressIPAllocator (cache updated), then hits the early-return guard:
if isOldReady == isNewReady && !isHostCIDRsAltered {
    return nil  // ← exits here; addEgressNode() is never called
}
Because only the CNCC annotation changed — not readiness and not host-cidrs — the handler returns early without calling addEgressNode().
Any EgressIPs whose IPv6 address could not be placed (no node had a matching V6.Net) remain permanently unassigned — addEgressNode() is the only code path that re-evaluates them.
Later, CNCC corrects the annotation (adding the IPv6 subnet). OVN-K again receives the UPDATE and again exits early, never reconsidering the unassigned IPv6 EgressIPs.
The net effect: even after both nodes report i.e. "ipv6":"2600:1f18:1816:ac01::/64" with capacity 14 in their annotation, the EgressIP status shows the IPv6 address as permanently unassigned.

Fix
pkg/util/node_annotations.go
Add CloudEgressIPConfigAnnotationChanged(), analogous to the existing NodeHostCIDRsAnnotationChanged()

pkg/clustermanager/egressip_event_handler.go
Add isCloudEgressIPConfigAltered to the early-return guard so that annotation changes from CNCC trigger addEgressNode(), giving previously-unassigned EgressIPs another chance to be reconciled with the updated node cache. A distinct log line is emitted when the trigger is an annotation change rather than a readiness transition.

Impact
Cloud platforms only (PlatformTypeIsEgressIPCloudProvider()): CloudEgressIPConfigAnnotationChanged only fires on nodes that CNCC manages, so bare-metal is unaffected.
Low frequency: CNCC typically writes the annotation once at node startup and rarely updates it, so addEgressNode() is not called more often than necessary.
No functional regression for IPv4: The change is additive — it only allows addEgressNode() to run in a case where it previously did not.

Related
Companion CNCC fix: [[openshift/cloud-network-config-controller#<PR>](https://github.com/openshift/cloud-network-config-controller) ](https://github.com/openshift/cloud-network-config-controller/pull/228) — fixes the root cause where getSubnet picks a stale disassociated IPv6 CIDR block association instead of the current associated one.

Testing
Pre-merge testing was performed with https://github.com/openshift/cloud-network-config-controller/pull/228 on dualstack AWS IPv4 primary and dualstack AWS IPv6 primary cluster
Regression testing with https://github.com/openshift/cloud-network-config-controller/pull/228 was performed on dualstack BM cluster

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of node updates so changes to cloud egress IP configuration are now recognized and reprocessed correctly.
  * Prevented updates from being skipped when only the cloud egress IP annotation changes.
  * Added clearer status logging during node re-evaluation, making it easier to understand why egress behavior is being recalculated.

### DIFF
--- a/go-controller/pkg/clustermanager/egressip_event_handler.go
+++ b/go-controller/pkg/clustermanager/egressip_event_handler.go
@@ -128,6 +128,7 @@ func (h *egressIPClusterControllerEventHandler) UpdateResource(oldObj, newObj in
 		isNewReady := h.eIPC.isEgressNodeReady(newNode)
 		isNewReachable := h.eIPC.isEgressNodeReachable(newNode)
 		isHostCIDRsAltered := util.NodeHostCIDRsAnnotationChanged(oldNode, newNode)
+		isCloudEgressIPConfigAltered := util.CloudEgressIPConfigAnnotationChanged(oldNode, newNode)
 		h.eIPC.setNodeEgressReady(newNode.Name, isNewReady)
 		if !oldHadEgressLabel && newHasEgressLabel {
 			klog.Infof("Node: %s has been labeled, adding it for egress assignment", newNode.Name)
@@ -142,7 +143,7 @@ func (h *egressIPClusterControllerEventHandler) UpdateResource(oldObj, newObj in
 			}
 			return nil
 		}
-		if isOldReady == isNewReady && !isHostCIDRsAltered {
+		if isOldReady == isNewReady && !isHostCIDRsAltered && !isCloudEgressIPConfigAltered {
 			return nil
 		}
 		if !isNewReady {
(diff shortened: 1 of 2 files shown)


jechen0648:
@coderabbitai review

jechen0648:
@coderabbitai help

jechen0648:
@coderabbitai full review

jechen0648:
u/s fix commits https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6648/commits have been synced to d/s, close this PR

PR #3318, opened 2026-07-23
OCPBUGS-99542: kubevirt: skip source LSP re-enable when source pod is not local

qinqon:
## 📑 Description

Manual backport of upstream ovn-kubernetes PR [#6229](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6229) to `release-4.20`.

Jira: [OCPBUGS-99542](https://issues.redhat.com/browse/OCPBUGS-99542)
4.21 backport: [OCPBUGS-99276](https://issues.redhat.com/browse/OCPBUGS-99276) (landed via branch sync openshift/ovn-kubernetes#3306)

During pod deletion on localnet topologies with interconnect, `enableSourceLSPFailedLiveMigration` attempts to re-enable the source pod's LSP when a migration is detected as failed. On localnet with IC the source pod's LSP only exists on the node where the source pod was scheduled, so when the target pod is deleted on a different node the function fails with "missing mac and ips", blocking target pod cleanup and leaving a zombie LSP re-created by the retry framework.

The fix adds the same locality guard used in `addLogicalPortToNetworkForNAD` to the delete path via a new `hasPodLogicalPort` helper: only touch the source LSP when it exists in the local zone (or the network is layer2 with IC).

Cherry-pick of 7c3330262d16 with conflict resolution for release-4.20:
- 4.20 uses `nadName` instead of `nadKey` and `isLayer2Interconnect()` instead of `isLayer2WithInterconnectTransport()`.
- Kept the 4.20-specific remote-pod cleanup block (`removeRemoteZonePodFromNamespaceAddressSet`) in `removePodForUserDefinedNetwork`, only switching its condition to the new `hasPodLogicalPort` helper.
- The added unit tests set the `kubevirt.io/domain` annotation and `virt-launcher` app label on the fake pods (as upstream later did in PR #6118) so this backport stays compatible with the [OCPBUGS-99543 ](https://github.com/openshift/ovn-kubernetes/pull/3319) backport regardless of merge order.

## ✅ Checks
- [x] My code requires tests

## How to verify it

Unit tests: `go test ./pkg/ovn/ -run TestClusterNode -args -ginkgo.focus="enableSourceLSPFailedLiveMigration"` (3 specs).

Manual: on a localnet secondary network with IC, live-migrate a VM, force the migration to fail, then delete the target virt-launcher pod from a node other than the source pod's node; verify no zombie LSP remains and connectivity is preserved.

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller_user_defined.go
+++ b/go-controller/pkg/ovn/base_network_controller_user_defined.go
@@ -357,10 +357,10 @@ func (bsnc *BaseUserDefinedNetworkController) addLogicalPortToNetworkForNAD(pod 
 		bsnc.logicalPortCache.remove(pod, nadName)
 	}
 
-	if shouldHandleLiveMigration &&
-		kubevirtLiveMigrationStatus.IsTargetDomainReady() &&
-		// At localnet there is no source pod remote LSP so it should be skipped
-		(bsnc.TopologyType() != types.LocalnetTopology || bsnc.isPodScheduledinLocalZone(kubevirtLiveMigrationStatus.SourcePod)) {
+	if shouldHandleLiveMigration && kubevirtLiveMigrationStatus.IsTargetDomainReady() {
+		// We have a source pod LSP at this zone if the source pod and:
+		// - layer2 IC There is one at all the zones since we have the remote LSP to implement east/west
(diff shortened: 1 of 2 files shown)


servolkov:
/verified by @servolkov

Verified OCPBUGS-99542 manually.

PR #3363, opened 2026-08-05
OCPBUGS-112470: Register test images using tests extension

arkadeepsen:
This PR registers the test images using tests extension so that these images are always available whenever the ovnk OTE tests are run.

### DIFF
--- a/openshift/Makefile
+++ b/openshift/Makefile
@@ -1,0 +1,6 @@
+EXCLUDE_DIRS ?= test/
+TEST_PKGS := $$(go list ./... | grep -v $(EXCLUDE_DIRS))
+
+.PHONY: test
+test:
+	go test -v $(TEST_PKGS)
--- a/openshift/cmd/ovn-kubernetes-tests-ext/images.go
+++ b/openshift/cmd/ovn-kubernetes-tests-ext/images.go
@@ -1,0 +1,89 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	imageutils "k8s.io/kubernetes/test/utils/image"
+
+	"github.com/openshift-eng/openshift-tests-extension/pkg/extension"
+)
+
+// requiredImage associates an e2e image pullspec with an index used to generate
+// the image tag so that it matches with the tag in quay.io/openshift/community-e2e-images.
+type requiredImage struct {
+	pullSpec string
+	index    int
+}
+
+var requiredImages []requiredImage
+
+func init() {
+	agnhostImage := requiredImage{
+		pullSpec: imageutils.GetE2EImage(imageutils.Agnhost),
+		index:    int(imageutils.Agnhost),
+	}
+	requiredImages = append(requiredImages, agnhostImage)
+}
+
+// registerTestImages advertises OVN-Kubernetes e2e images to the openshift-tests
+// extension so origin can list and mirror them (see "images" subcommand).
+func registerTestImages(ext *extension.Extension) error {
+	for _, ri := range requiredImages {
+		img, err := extensionImageFromPullSpec(ri.pullSpec)
+		if err != nil {
+			return fmt.Errorf("failed to register test image %q: %v", ri.pullSpec, err)
+		}
+		img.Index = ri.index
+		ext.RegisterImage(img)
+	}
+	return nil
+}
+
+// extensionImageFromPullSpec splits a pullspec into the registry/name/version
+// layout expected by k8s.io/kubernetes/test/utils/image.Config.GetE2EImage
+// (fmt.Sprintf("%s/%s:%s", registry, name, version)).
+func extensionImageFromPullSpec(pullSpec string) (extension.Image, error) {
+	registry, name, version, err := splitImagePullSpec(pullSpec)
+	if err != nil {
+		return extension.Image{}, err
+	}
+	return extension.Image{
+		Registry: registry,
+		Name:     name,
+		Version:  version,
+	}, nil
+}
+
+func splitImagePullSpec(pullSpec string) (registry, name, version string, err error) {
+	if pullSpec == "" {
+		return "", "", "", fmt.Errorf("empty image pullspec")
+	}
+	if strings.Contains(pullSpec, "@") {
+		return "", "", "", fmt.Errorf("digest image pullspecs are not supported: %q", pullSpec)
+	}
+
+	remainder := pullSpec
+	if tagIndex := strings.LastIndex(pullSpec, ":"); tagIndex > strings.LastIndex(pullSpec, "/") {
+		version = pullSpec[tagIndex+1:]
+		remainder = pullSpec[:tagIndex]
+	}
+	if version == "" {
+		version = "latest"
+	}
+
+	first, rest, ok := strings.Cut(remainder, "/")
+	if !ok {
+		// Short docker-library name, e.g. "nginx" → docker.io/library/nginx
+		return "docker.io", "library/" + remainder, version, nil
+	}
+
+	if strings.ContainsAny(first, ".:") || first == "localhost" {
+		// Explicit registry, e.g. "quay.io/foo/bar", "registry.k8s.io/...",
+		// or "localhost:5000/ovn/test"
+		return first, rest, version, nil
+	}
+
+	// Docker Hub user/org image, e.g. "cloudflare/goflow"
+	return "docker.io", remainder, version, nil
+}
--- a/openshift/cmd/ovn-kubernetes-tests-ext/main.go
+++ b/openshift/cmd/ovn-kubernetes-tests-ext/main.go
@@ -77,7 +77,10 @@ func main() {
 	// Create our registry of openshift-tests extensions
 	extensionRegistry := extension.NewRegistry()
 	ovnTestsExtension := extension.NewExtension("openshift", "payload", "ovn-kubernetes")
-	// TODO: register test images using tests extension
+	// register OVN-Kubernetes e2e images to the openshift-tests extension
+	if err := registerTestImages(ovnTestsExtension); err != nil {
+		panic(err)
+	}
 	// add ovn-kubernetes test suites into openshift suites
 	// by default, we treat all tests as parallel and only expose tests as Serial if the appropriate label is added - "Serial"
 	ovnTestsExtension.AddSuite(extension.Suite{
--- a/openshift/cmd/ovn-kubernetes-tests-ext/images_test.go
+++ b/openshift/cmd/ovn-kubernetes-tests-ext/images_test.go
@@ -1,0 +1,194 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/openshift-eng/openshift-tests-extension/pkg/extension"
+)
+
+func TestSplitImagePullSpec(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name             string
+		pullSpec         string
+		wantRegistry     string
+		wantName         string
+		wantVersion      string
+		wantErrSubstring string
+	}{
+		{
+			name:         "k8s promoter image",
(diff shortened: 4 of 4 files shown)


openshift/cmd/ovn-kubernetes-tests-ext/images.go
```diff
@@ -0,0 +1,81 @@
+package main
```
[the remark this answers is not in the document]
arkadeepsen: This is not meant for go-controller package.

openshift/cmd/ovn-kubernetes-tests-ext/main.go
```diff
@@ -77,7 +77,8 @@ func main() {
 	// Create our registry of openshift-tests extensions
 	extensionRegistry := extension.NewRegistry()
 	ovnTestsExtension := extension.NewExtension("openshift", "payload", "ovn-kubernetes")
-	// TODO: register test images using tests extension
+	// register OVN-Kubernetes e2e images to the openshift-tests extension
+	registerTestImages(ovnTestsExtension)
```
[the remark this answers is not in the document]
arkadeepsen: This package sets up the ovn-kubernetes e2e tests in openshift. There's no need for e2e for this code.

openshift/Makefile
```diff
@@ -0,0 +1,6 @@
+EXCLUDE_DIRS ?= test/
+TEST_PKGS := $$(go list ./... | grep -v $(EXCLUDE_DIRS))
+
+.PHONY: test
+test:
+	go test -v $(TEST_PKGS)
```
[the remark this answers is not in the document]
arkadeepsen: The makefile target usage is simple enough and doesn't need any separate documentation.

openshift/Makefile
```diff
@@ -0,0 +1,6 @@
+EXCLUDE_DIRS ?= test/
+TEST_PKGS := $$(go list ./... | grep -v $(EXCLUDE_DIRS))
```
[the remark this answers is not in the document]
arkadeepsen: Addressed in the latest push

arkadeepsen:
@coderabbitai review

openshift/cmd/ovn-kubernetes-tests-ext/images.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+var requiredImages []requiredImage
+
+func init() {
+	agnhostImage := requiredImage{
+		pullSpec: imageutils.GetE2EImage(imageutils.Agnhost),
+		index:    int(imageutils.Agnhost),
+	}
+	requiredImages = append(requiredImages, agnhostImage)
```
jcaamano, line 26: Is there any plans to integrate this with other images that upstream CI uses (iperf, netshoot, ....) already available through `images.Required()`?

openshift/cmd/ovn-kubernetes-tests-ext/images.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// registerTestImages advertises OVN-Kubernetes e2e images to the openshift-tests
+// extension so origin can list and mirror them (see "images" subcommand).
+func registerTestImages(ext *extension.Extension) {
+	for _, ri := range requiredImages {
+		img, err := extensionImageFromPullSpec(ri.pullSpec)
```
jcaamano, line 33: What is translating the pull spec that we get from `k8s.io/kubernetes/test/utils/image` to the mirrored image pull spec?

openshift/Makefile
```diff
@@ -0,0 +1,6 @@
...
+
+.PHONY: test
+test:
```
jcaamano, line 5: Do you plan to add this target to run in our CI?

openshift/cmd/ovn-kubernetes-tests-ext/images.go
arkadeepsen: The current d/s images.Required() returns the agnhost registered here:

https://github.com/openshift/ovn-kubernetes/blob/7b4de5ed3bd4381e3a17cdfddbe14be1432b9860/openshift/test/deploymentconfig/openshift.go#L63-L67

This is actually not the image used in the tests. The tests use the current k8s version's tagged agnhost:
https://github.com/openshift/ovn-kubernetes/blob/7b4de5ed3bd4381e3a17cdfddbe14be1432b9860/test/e2e/multihoming_utils.go#L298

Thus, directly using `images.Required()` didn't solve the problem. So I specifically used the current k8s version's agnhost. Also, no other images were returned by  `images.Required()`  and the metal CI jobs only complained about the agnhost image not being available. So, I presume, the tests which use the other images are not enabled via OTE?

Another thing to note here is that we need these images to be first available in `quay.io/openshift/community-e2e-images` from where these images will be mirrored in the metal CI jobs. So, if these images are not already available in the quay registry, they need to be registered first.

If they are already available, the same index should be used so that the correct image tag is generated. For the u/s k8s test images, the corresponding `ImageID` is the index which is used for registering them in the quay registry. That's what I have used in this PR as well for agnhost.

openshift/cmd/ovn-kubernetes-tests-ext/images.go
arkadeepsen: When registering the images, we need to provide each of the fields separately, `Registry`, `Name`, `Version` and `Index`. These registered images are then used by `openshift-tests` binary which generates the mirrored pull spec. The code resides in origin repo:

https://github.com/openshift/origin/blob/ed9fd91d108a55c891815f4f442cca8e3538de64/pkg/cmd/openshift-tests/images/images_command.go#L133-L139

There was an issue there which caused problem in generating the list of images to be mirrored when same image was registered by multiple OTE binaries. It was fixed as part of the same bug: https://github.com/openshift/origin/pull/31542

openshift/Makefile
arkadeepsen: Yes. I have already enabled the linter for the OTE code: https://github.com/openshift/release/blob/f56a1a9654b3995bf2c5d7f2460451508f200401/ci-operator/config/openshift/ovn-kubernetes/openshift-ovn-kubernetes-main.yaml#L376-L384

Will do the same for the test target as well.

openshift/cmd/ovn-kubernetes-tests-ext/images.go
jcaamano: > This is actually not the image used in the tests.

It is on tests that pull it through the `images.AgnHost()` method directly, isn't it?

> Another thing to note here is that we need these images to be first available in `quay.io/openshift/community-e2e-images` from where these images will be mirrored in the metal CI jobs. So, if these images are not already available in the quay registry, they need to be registered first.

Yup.

> If they are already available, the same index should be used so that the correct image tag is generated. For the u/s k8s test images, the corresponding `ImageID` is the index which is used for registering them in the quay registry. That's what I have used in this PR as well for agnhost.

index?

jcaamano: > index?

sorry, got it now

arkadeepsen: This is where I am using agnhost's `ImageID` as it's index: https://github.com/openshift/ovn-kubernetes/pull/3363/changes#diff-dfeeafda1798f6c00a4f9d5a03b37fec9aa520b0ed2f01db77ee9c82b2b7780bR16

This is where the same index is used: https://github.com/openshift/ovn-kubernetes/pull/3363/changes#diff-dfeeafda1798f6c00a4f9d5a03b37fec9aa520b0ed2f01db77ee9c82b2b7780bR37

This index along with the rest of the pullSpec (registry, name and version) is used to generate the image tag for mirroring. So, if the correct index is not used, then the image mirror will fail.

This should be the same index which is used to generate the tag for the `quay.io/openshift/community-e2e-images` registry as well.

openshift/cmd/ovn-kubernetes-tests-ext/images.go
jcaamano: Ok, what I was after is that after registering, `openshift-test images` will output it

arkadeepsen: Yes, it'll. That's is also what's being used by the metal job's script to get the list of images to mirror:
https://github.com/openshift/release/blob/4021c1460685d0a0c37cf2ddeeb9800c531c3199/ci-operator/step-registry/baremetalds/e2e/test/baremetalds-e2e-test-commands.sh#L114

openshift/cmd/ovn-kubernetes-tests-ext/images.go
jcaamano: so we should probably return `imageutils.GetE2EImage(imageutils.Agnhost)` in `GetAgnHostContainerImage` as well in this PR?

And then FUP to replace `GetAgnHostContainerImage` upstream/downstream with a more generic `GetE2EImage` that will just forward to `imageutils.GetE2EImage` when appropriate and otherwise our specific images.

arkadeepsen: > > This is actually not the image used in the tests.
> 
> It is on tests that pull it through the `images.AgnHost()` method directly, isn't it?

The tests use the upstream helper function which uses the u/s image directly and don't use `images.AgnHost()` for the image. Also the agnhost:2.40 is only needed for external container in metal jobs as mentioned in the comments.

arkadeepsen: > so we should probably return `imageutils.GetE2EImage(imageutils.Agnhost)` in `GetAgnHostContainerImage` as well in this PR?

We need the corresponding index (ImageID) to be returned as well, otherwise it'll not work correctly. Currently, only the image is being returned. This ImageID may change based on the k8s version.

jcaamano: I see it used in some pods, but why does it matter? Wouldn't we want to use the same AgnHost image everywhere as long we can pull it?

openshift/cmd/ovn-kubernetes-tests-ext/images.go
```diff
@@ -0,0 +1,88 @@
...
+	registry, name, version, err := splitImagePullSpec(pullSpec)
+	if err != nil {
+		return extension.Image{}, err
+	}
+	return extension.Image{
+		Registry: registry,
+		Name:     name,
+		Version:  version,
+	}, nil
+}
```
jcaamano, line 56: Have you checked `imageutils.GetConfig()`, it already gives you the split I think.

arkadeepsen: Yes. I was using it earlier here: https://github.com/arkadeepsen/openshift-ovn-kubernetes/commit/366973938#diff-dfeeafda1798f6c00a4f9d5a03b37fec9aa520b0ed2f01db77ee9c82b2b7780bR112-R121

I kept the parsing code as it'll work for images which are not part of upstream k8s test images.

openshift/cmd/ovn-kubernetes-tests-ext/images.go
```diff
@@ -0,0 +1,88 @@
...
+
+// registerTestImages advertises OVN-Kubernetes e2e images to the openshift-tests
+// extension so origin can list and mirror them (see "images" subcommand).
+func registerTestImages(ext *extension.Extension) {
+	for _, ri := range requiredImages {
+		img, err := extensionImageFromPullSpec(ri.pullSpec)
+		if err != nil {
+			panic(fmt.Sprintf("failed to register test image %q: %v", ri.pullSpec, err))
+		}
```
jcaamano, line 36: I would just return the error, and leave it up to the caller what to do with it, which will probably panic anyway

jcaamano (commented):
sorry for the staggered review

openshift/cmd/ovn-kubernetes-tests-ext/images.go
arkadeepsen: Done

PR #3383, opened 2026-08-11
NO-JIRA: DownStream Merge [08-11-2026]

arkadeepsen:
Manual d/s merge to resolve conflict in gateway_init_linux_test.go file. Following are the conflict resolution details:

- Resolve gateway_init_linux_test.go merge conflict for MCS nftables migration
  - Resolve conflict between downstream MCS iptables expectations and
    upstream nftables gateway-service expectations.
  - Keep upstream's expectedNFT path (nftablesRulesBase +
    nftablesRulesGatewayServices, plus local-gateway rules where needed).
  - Drop the conflicting OCP HACK iptables MatchState checks for MCS
    REJECT rules on filter/FORWARD and filter/OUTPUT.
  - Add nftablesRulesMCS and append it to expectedNFT in shared-gateway and
    local-gateway tests so MCS blocking is asserted via nftables instead.

### DIFF
--- a/go-controller/pkg/node/OCP_HACKS.go
+++ b/go-controller/pkg/node/OCP_HACKS.go
@@ -4,51 +4,115 @@
 package node
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/coreos/go-iptables/iptables"
 
+	"sigs.k8s.io/knftables"
+
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/config"
 	nodeipt "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/iptables"
+	nodenft "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/nftables"
 )
 
-// Block MCS Access. https://github.com/openshift/ovn-kubernetes/pull/170
-func generateBlockMCSRules(rules *[]nodeipt.Rule, protocol iptables.Protocol) {
-	var delRules []nodeipt.Rule
+const (
+	mcsBlockingChain        = "mcs-blocking"
+	mcsBlockingOutputChain  = "mcs-blocking-output"
+	mcsBlockingForwardChain = "mcs-blocking-forward"
+)
 
-	for _, chain := range []string{"FORWARD", "OUTPUT"} {
-		for _, port := range []string{"22623", "22624"} {
-			*rules = append(*rules, nodeipt.Rule{
-				Table:    "filter",
-				Chain:    chain,
-				Args:     []string{"-p", "tcp", "-m", "tcp", "--dport", port, "--syn", "-j", "REJECT"},
-				Protocol: protocol,
-			})
-			// Delete the old "--syn"-less rules on upgrade
-			delRules = append(delRules, nodeipt.Rule{
-				Table:    "filter",
-				Chain:    chain,
-				Args:     []string{"-p", "tcp", "-m", "tcp", "--dport", port, "-j", "REJECT"},
-				Protocol: protocol,
-			})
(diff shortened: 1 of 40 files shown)


jluhrsen:
@arkadeepsen everything almost passed for CI here except the perfscale job, but we have a [thread you started](https://redhat-internal.slack.com/archives/CU9HKBZKJ/p1786548221754239) and it's known to be failing.

go-controller/pkg/node/gateway_init_linux_test.go
```diff
@@ -101,6 +101,58 @@ add chain inet ovn-kubernetes ovn-kube-pod-subnet-masq
...
+add rule inet ovn-kubernetes services fib daddr type local dnat ip6 addr . port to meta l4proto . th dport map @nodeports-v6
+add rule inet ovn-kubernetes services-output jump services-itp
+add rule inet ovn-kubernetes services-output jump services
+add rule inet ovn-kubernetes services-prerouting jump services-etp
+add rule inet ovn-kubernetes services-prerouting jump services-etp-no-nodeport
+add rule inet ovn-kubernetes services-prerouting jump services
+`
+
+// OCP HACK: Block MCS Access. https://github.com/openshift/ovn-kubernetes/pull/170
+const nftablesRulesMCS = `
```
danwinship, line 145: We could avoid needing this (and thus avoid needing any changes at all to `gateway_init_linux_test.go` and thus avoid future merge conflicts) by putting the MCS filtering rules into a different nftables table (eg, `openshift-ovn-kubernetes`).

You could add `pkg/node/nftables/OCP_HACKS.go` with `GetOpenShiftNFTablesHelper()` to return a knftables interface for the `openshift-ovn-kubernetes` table. (It could know whether to make a real or fake interface by first calling `GetNFTablesHelper()` and seeing whether that function returns a `*knftables.Fake` or not). Then `pkg/node/OCP_HACKS.go:setupMCSBlockNFTRules()` could use that rather than `GetNFTablesHelper()`, and nothing else should need to change.

(This could also be done later, separately from the merge PR, if we need to get this in right away though.)

go-controller/pkg/node/OCP_HACKS.go
```diff
@@ -4,51 +4,115 @@
...
-			})
-			// Delete the old "--syn"-less rules on upgrade
-			delRules = append(delRules, nodeipt.Rule{
-				Table:    "filter",
-				Chain:    chain,
-				Args:     []string{"-p", "tcp", "-m", "tcp", "--dport", port, "-j", "REJECT"},
-				Protocol: protocol,
-			})
+// cleanupLegacyMCSBlockIptRules best-effort deletes leftover host iptables MCS REJECT
+// rules from before the nftables migration (both --syn and legacy non--syn variants).
```
danwinship, line 26: I suspect the "legacy non--syn variant" is old enough that we can't possibly be upgrading from a version of OCP with those rules at this point, so we don't really need to worry about that...

go-controller/pkg/node/OCP_HACKS.go
```diff
@@ -4,51 +4,115 @@
...
+		Name: mcsBlockingChain,
+	})
+	tx.Flush(&knftables.Chain{
+		Name: mcsBlockingChain,
+	})
+	tx.Add(&knftables.Rule{
+		Chain: mcsBlockingChain,
+		Rule: knftables.Concat(
+			"tcp dport { 22623, 22624 } tcp flags syn / fin,syn,rst,ack",
+			"reject",
```
danwinship, line 78: (This matches the per-pod MCS blocking rule from `pkg/cni/OCP_HACKS.go`, so it's presumably correct.)

danwinship (commented):
The changes look right to me

tssurya (approved):
/lgtm

PR #3422, opened 2026-08-26
OCPBUGS-105420: order bare-metal EgressIP failover to avoid dual ARP

Shraddhaa-28:
## Summary
- On bare-metal, EgressIP reassignment now removes invalid status first (like cloud), then assigns only after the previous node clears `k8s.ovn.org/bridge-egress-ips` / `k8s.ovn.org/secondary-host-egress-ips`.
- This prevents the new node from GARP'ing while the old node still answers ARP during label-removal failover (OCPBUGS-105420).
- Fail-open after 30s if annotation cleanup never happens so failover cannot stall forever.

## Test plan
- [x] Unit tests: status cleared while annotation remains; assignment after annotation cleanup
- [x] Unit tests: fail-open reassignment after cleanup timeout
- [x] Existing `should re-balance EgressIPs when their node is removed` still passes
- [ ] E2E / manual: remove `k8s.ovn.org/egress-assignable` from a node hosting many EIPs; confirm only the new node answers ARP

Made with [Cursor](https://cursor.com)

### DIFF
--- a/go-controller/pkg/clustermanager/egressip_controller.go
+++ b/go-controller/pkg/clustermanager/egressip_controller.go
@@ -44,6 +44,11 @@ import (
 
 const (
 	egressIPReachabilityCheckInterval = 5 * time.Second
+	// egressIPNodeCleanupTimeout is how long cluster-manager waits for an old
+	// egress node to remove an EgressIP from its node annotations (bridge /
+	// secondary-host) before re-assigning the IP. Avoids dual ARP answers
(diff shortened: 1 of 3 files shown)


ssonigra:
@Shraddhaa-28 was this fix tested? There is another bug from arnab which is fixing the same issue I think we will have to close this PR in favour of Arnab's fix

ssonigra:
@Shraddhaa-28  please close out this PR in favour of PR raised by Arnab who has already raised the pr to fix it.

PR #3424, opened 2026-08-26
[release-4.18] OCPBUGS-114016: Disable IGMP/MLD querier unless multicast is used

DeokarT:
OCPBUGS-114016

This is the leftover of the ARP/GARP 4096 work (OCPBUGS-85627). After 4.18.53 the unicast ARP/GARP warnings on `br-int` are gone. IGMP Membership Queries still hit the same OVS hard limit and get dropped.

On a dense cluster (the report had ~321 namespaces and a lot of pods) ovn-controller periodically injects IGMPv3 general queries. The ofproto log looks like:

```
over 4096 resubmit actions on bridge br-int while processing
ip,in_port=CONTROLLER,dl_dst=01:00:5e:00:00:01,nw_src=10.131.72.1,nw_dst=224.0.0.1,nw_proto=2,...
sending OFPFMFC_UNKNOWN error reply to OFPT_PACKET_OUT message
```

`in_port=CONTROLLER` is the important part: this is not a pod packet. OVN pinctrl is acting as the local IGMP/MLD querier. `nw_src` is the node logical-router port (`mcast_ip4_src`). Destination is `224.0.0.1` / `01:00:5e:00:00:01`. Those packet-outs are flooded to every local port on `br-int`. Nested resubmits go over 4096 and OVS replies `OFPFMFC_UNKNOWN` to the `OFPT_PACKET_OUT`.

OpenShift hits this even on clusters that never use multicast. ovn-kubernetes has a cluster-wide multicast-support flag; upstream that can be off, but OpenShift always turns it on. `createNodeLogicalSwitch` was therefore setting both `mcast_snoop=true` and `mcast_querier=true` on every node switch as soon as the gateway IP was known.

Snooping is cheap and we want it. The querier is the expensive part: it is what injects the queries. It was running even when no namespace had `k8s.ovn.org/multicast-enabled=true`. Most OpenShift clusters never opt into multicast, so every node was generating unused queries all the time.

This patch keeps snooping as-is when multicast support is enabled. It only sets `other_config:mcast_querier=true` while at least one namespace currently has the multicast annotation.

That is wired in three places:

- new node switches (`createNodeLogicalSwitch`)
- namespace multicast enable / disable / delete
- namespace sync, so existing switches get corrected without waiting for a node recreate

`syncNodeLogicalSwitchQueriers` only touches switches that already have `mcast_snoop=true`. If querier source addresses are missing we log and skip rather than turning the querier on half-configured.

If someone actually annotates a namespace for multicast, the querier comes back on. In that case the same flood path can still hit 4096. The datapath fix belongs in OVN (inject local queries with `MC_FLOOD_L2` instead of `MC_FLOOD`) and has to go through FDP, not this repo. This PR does not turn multicast support off; it only gates the querier.

How to test
Unit tests:

```
go test ./pkg/ovn -ginkgo.focus "IGMP querier"
```

That covers an existing node switch with the querier already on and no annotated namespace (should go `mcast_querier=false`), and annotate / unannotate a namespace (querier should flip true then false). Default network and UDN/network-segmentation IPv4 are both in the table.

On a 4.18 cluster with multicast support on and no `k8s.ovn.org/multicast-enabled` namespaces:

```
ovn-nbctl get logical_switch <node> other_config:mcast_snoop
ovn-nbctl get logical_switch <node> other_config:mcast_querier
```

Snoop should still be true. Querier should be false. The IGMP `in_port=CONTROLLER` 4096 warnings should stop.

Then annotate one namespace with `k8s.ovn.org/multicast-enabled=true`, confirm querier goes true and multicast still works, and remove the last annotation to confirm the querier goes false again.

### DIFF
--- a/docs/features/multicast.md
+++ b/docs/features/multicast.md
@@ -129,15 +129,22 @@ lists is the switch that is configured to act as an IGMP querier - on OVN-K's
 case, the node's logical switches.
 
 For these membership requests to be allowed in the network, each node's logical
-switch's must declare themselves "multicast queriers", by using the
-`other_config:mcast_querier=true` option.
+switch must declare itself a "multicast querier" with
+`other_config:mcast_querier=true`.
 
-The `other_config:mcast_snoop=true` is also used so the logical switches can
-passively snoop IGMP query / report / leave packets transferred between the
-multicast hosts and switches to compute group membership.
+Snooping and the querier are separate knobs. `mcast_snoop=true` is what lets
+the switch watch IGMP query / report / leave packets and build group
+membership. The querier is the thing that *injects* those queries. We only
+set `mcast_querier=true` while at least one namespace has
+`k8s.ovn.org/multicast-enabled=true`. OpenShift always enables multicast
+support in ovn-kubernetes, so if we also leave the querier on all the time,
+ovn-controller keeps sending unused IGMP/MLD queries. Those get flooded to
+every local port on `br-int` and can exceed the OVS 4096 resubmit limit.
 
-Without these two options, multicast traffic would be treated as broadcast
-traffic, which forwards packets to all ports on the network.
+Without snooping, multicast is treated as broadcast and forwarded to every
+port. The snippets below are from a cluster that is actually using multicast
+(`mcast_querier=true`). If multicast support is on but no namespace has the
+annotation, you should see `mcast_querier=false`.
 
 Please refer to the following snippet featuring the node's logical swithes
 of a cluster with one control plane node, and two workers, to see these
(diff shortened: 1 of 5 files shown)


DeokarT:
Prow is not rejecting the code. After linking [OCPBUGS-114675](https://issues.redhat.com/browse/OCPBUGS-114675) (target 4.19.z) as the dependent of [OCPBUGS-114016](https://issues.redhat.com/browse/OCPBUGS-114016), `/jira refresh` now finds that dependent.

The remaining `jira/invalid-bug` check is that OCPBUGS-114675 is **New**, not VERIFIED / RELEASE PENDING / CLOSED (ERRATA|CURRENT RELEASE|DONE|DONE-ERRATA). OpenShift will not treat a `release-4.18` bug as valid until that 4.19 parent is VERIFIED (latest-first clone chain) or a staff engineer / patch manager adds `jira/skip-dependent-bug-check`.

